PR #47, opened 2022-10-28
Add C++ dependencies and configuration files

### LINKED ISSUE #38: Add C++ dependencies and configuration files
### Changes
- Explore VS Code C++ integration
    - Linting and formatting (keep as tasks, or use extensions and format on save?)

### Resources
- https://code.visualstudio.com/docs/cpp/cpp-ide#:~:text=Code%20formatting%23,in%20right%2Dclick%20context%20menu.

hhenry01:
Update the workspace to utilize the C++ tools outlined [here](https://ubcsailbot.github.io/docs/reference/cpp/tools/) in conjunction with the [Network Systems repository](https://github.com/UBCSailbot/network_systems/tree/main).

Closes #38 

Tasks remaining:
- [x] Figure out CMake linting. ament_lint_cmake does not filter GoogleTest CMake files and will spit out a lot of errors.
- [ ] Get a unit test coverage tool working. I made an attempt with llvm-cov but could not figure it out. Some other tool like lcov could also work.
- [x] Integrate tools with CI.
- [x] Add separate build without lint option (for CI)
- [ ] Add ament_clang-tidy to required check once merged in
- [ ] Add Linux CAN interface dependency

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,27 +1,30 @@
 FROM ghcr.io/ubcsailbot/sailbot_workspace/dev:infra_updates
 
 ## Uncomment this section to install the packages used in the ROS 2 tutorials
-# ENV DEBIAN_FRONTEND=noninteractive
-# RUN apt-get update \
-#  && apt-get -y install --no-install-recommends \
-#  # Your package list here
-#  ros-eloquent-turtlesim \
-#  # Clean up
-#  && apt-get autoremove -y \
-#  && apt-get clean -y \
-#  && rm -rf /var/lib/apt/lists/*
-# ENV DEBIAN_FRONTEND=
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
+ && apt-get -y install --no-install-recommends \
+ # Your package list here
+ ros-eloquent-turtlesim \
+ # Clean up
+ && apt-get autoremove -y \
+ && apt-get clean -y \
+ && rm -rf /var/lib/apt/lists/*
+ENV DEBIAN_FRONTEND=
 
 ## Uncomment this section to install additional packages
-# ENV DEBIAN_FRONTEND=noninteractive
-# RUN apt-get update \
-#  && apt-get -y install --no-install-recommends \
-#  # Your package list here
-#  # Clean up
-#  && apt-get autoremove -y \
-#  && apt-get clean -y \
-#  && rm -rf /var/lib/apt/lists/*
-# ENV DEBIAN_FRONTEND=
+RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
+ENV DEBIAN_FRONTEND=noninteractive
+RUN apt-get update \
+ && apt-get -y install --no-install-recommends \
+ clang-10 clang-tidy-10 clangd-10 \
+ && update-alternatives --install /usr/bin/clangd clangd /usr/bin/clangd-10 100 \
+ # Your package list here
+ # Clean up
+ && apt-get autoremove -y \
+ && apt-get clean -y \
+ && rm -rf /var/lib/apt/lists/*
+ENV DEBIAN_FRONTEND=
 
 # Copy configuration files (e.g., .vimrc) from config/ to the container's home directory
 ARG USERNAME=ros
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,16 +1,16 @@
 FROM ghcr.io/ubcsailbot/sailbot_workspace/dev:cppworkspace
 
 ## Uncomment this section to install the packages used in the ROS 2 tutorials
-ENV DEBIAN_FRONTEND=noninteractive
-RUN apt-get update \
- && apt-get -y install --no-install-recommends \
- # Your package list here
- ros-eloquent-turtlesim \
- # Clean up
- && apt-get autoremove -y \
- && apt-get clean -y \
- && rm -rf /var/lib/apt/lists/*
-ENV DEBIAN_FRONTEND=
+# ENV DEBIAN_FRONTEND=noninteractive
+# RUN apt-get update \
+#  && apt-get -y install --no-install-recommends \
+#  # Your package list here
+#  ros-eloquent-turtlesim \
+#  # Clean up
+#  && apt-get autoremove -y \
+#  && apt-get clean -y \
+#  && rm -rf /var/lib/apt/lists/*
+# ENV DEBIAN_FRONTEND=
 
 ## Uncomment this section to install additional packages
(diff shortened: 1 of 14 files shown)


.devcontainer/config/sailbot_workspace.code-workspace
```diff
@@ -41,36 +41,13 @@
...
-			"statusBar.background": "#113262",
-			"statusBar.foreground": "#e7e7e7",
-			"statusBarItem.hoverBackground": "#19488d",
-			"statusBarItem.remoteBackground": "#113262",
-			"statusBarItem.remoteForeground": "#e7e7e7",
-			"titleBar.activeBackground": "#113262",
-			"titleBar.activeForeground": "#e7e7e7",
-			"titleBar.inactiveBackground": "#11326299",
-			"titleBar.inactiveForeground": "#e7e7e799",
-			"commandCenter.border": "#e7e7e799"
```
patrick-5546, line 73: You removed the border colours again, I think you have to manually uninstall peacock to stop this from happening.

.devcontainer/config/sailbot_workspace.code-workspace
```diff
@@ -114,7 +91,15 @@
 			"Wextra",
 			"Wpedantic",
 			"xmllint"
-		]
+		],
+		"editor.formatOnSave": false,
```
patrick-5546, line 95: If this is the default behavior, you can remove this

.github/actions/lint/run.sh
```diff
@@ -3,4 +3,10 @@ set -e
 
 source /opt/ros/${ROS2_DISTRO}/setup.bash
 ./setup.sh
-ament_${LINTER} src/
+if [[ "$LINTER" == "clang-tidy" ]]
+then
+    ./build.sh OFF
```
patrick-5546, line 8: Why do you build only for clang?

.vscode/c_cpp_properties.json
```diff
@@ -7,11 +7,11 @@
                 "/opt/ros/eloquent/include/**"
             ],
             "defines": [],
-            "compilerPath": "/usr/bin/gcc",
+            "compilerPath": "/usr/bin/clang",
```
patrick-5546, line 10: Path doesn't exit, I think you have to use `clang-10`

patrick-5546 (changes requested):
Wow great work on this PR, I know almost nothing about c++ infrastructure so it's cool to see what it can do

Some general comments I had:
- Do you not have to include/exclude directories for any linter anymore?
- Do these changes warrant a docs page on how to use them?
- You can add the network_systems folder to the workspace right after sailbot_workspace
- network_systems builds now in 10.5s, but it has some warnings
    image
- the clang-tidy task doesn't print anything to stdout, look at the output of other linting tasks for reference.
    image
    image
- Do we need linting tasks for linters that have vscode extensions? Extensions and tasks with problem matchers will both highlight errors inline in the editor, but extension errors will disappear once fixed while problem matcher errors will only disappear after the task is run again.

patrick-5546:
I added a to do in the PR description

.devcontainer/Dockerfile
```diff
@@ -13,6 +13,7 @@ FROM ghcr.io/ubcsailbot/sailbot_workspace/dev:infra_updates
 # ENV DEBIAN_FRONTEND=
 
 ## Uncomment this section to install additional packages
+# RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key|sudo apt-key add -
```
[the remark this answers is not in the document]
hhenry01: Yes

.devcontainer/config/sailbot_workspace.code-workspace
hhenry01: Weird, I don't have peacock installed at all.

.github/actions/lint/run.sh
hhenry01: clang-tidy relies on a file generated by CMake to run properly.

hhenry01:
> Wow great work on this PR, I know almost nothing about c++ infrastructure so it's cool to see what it can do
> 
> Some general comments I had:
> 
>     * Do you not have to include/exclude directories for any linter anymore?
> 
>     * Do these changes warrant a docs page on how to use them?
> 
>     * You can add the network_systems folder to the workspace right after sailbot_workspace
> 
>     * network_systems builds now in 10.5s, but it has some warnings
>       image
> 
>     * the clang-tidy task doesn't print anything to stdout, look at the output of other linting tasks for reference.
>       image
>       image
> 
>     * Do we need linting tasks for linters that have vscode extensions? Extensions and tasks with problem matchers will both highlight errors inline in the editor, but extension errors will disappear once fixed while problem matcher errors will only disappear after the task is run again.

1. For the time being, since I removed googletest from the Network Systems repo, there isn't any third-party code causing style violations.
2. I'll probably expand the tools page to explain how I got googletest, protobuf, etc to work with CMake to make the build system less esoteric.
3. k
4. This is caused by auto-generated protobuf files and is emitted by the clang++ itself instead of clang-tidy. Since I configured clang-tidy to emit warnings for unused parameters, I think it should be ok to disable that specific warning for clang++.
5. I'm not sure what you mean by this. If there's no problems it returns 0 and if it catches a problem it shows it and returns 1.
image
6. Not sure. Since the CI doesn't use the tasks themselves they're definitely not essential. They're just a convenient way to check for linting before you push, especially when sometimes squiggly error lines are wrong. The clang-tidy task automatically resolves itself without rerunning at least.

patrick-5546:
> 2. I'll probably expand the tools page to explain how I got googletest, protobuf, etc to work with CMake to make the build system less esoteric.

I was more referring to a docs page specific to sailbot_workspace talking about the c++ infrastructure, specifically how to use it

> 4. This is caused by auto-generated protobuf files and is emitted by the clang++ itself instead of clang-tidy. Since I configured clang-tidy to emit warnings for unused parameters, I think it should be ok to disable that specific warning for clang++.

Isn't a good idea in general to check for unused parameters? Another option would be to ignore the lib directory
 
> 5. I'm not sure what you mean by this. If there's no problems it returns 0 and if it catches a problem it shows it and returns 1. I think it's just weird cause it's printing out all the warnings generated by the ROS and system source code even though it's ignoring it all.

Makes sense, I just wanted to double check that it is behaving as expected

> 6. Not sure. Since the CI doesn't use the tasks themselves they're definitely not essential. They're just a convenient way to check for linting before you push, especially when sometimes squiggly error lines are wrong. The clang-tidy task automatically resolves itself without rerunning at least.

I'm thinking about making a new infra issue to clean up the tasks. For example, we don't need to have a flake8 task if we are using the flake8 vs code extension. The extension would only check the files that you have open, but I think this is fine because main should pass CI when you branched off it.

patrick-5546 (approved):
One more question that I have: what will have to change if we move the dev container ubuntu distro from 18.04 to 22.04? Let's discuss this in the upcoming infra meeting

PR #53, opened 2022-11-05
Ignore distrowatch links

patrick-5546:
Was failing markdownlinkcheck CI


patrick-5546:
@hhenry01 this is how you would add a link to the ignore list for markdownlinkcheck

PR #60, opened 2022-11-19
Update Scripts to Migrate CI to Subteam Repos

hhenry01:
### Description
- Resolves https://github.com/UBCSailbot/tasks/issues/22


patrick-5546 (changes requested):
Build dev image and markdown linters should remain in sailbot_workspace.

hhenry01:
> Build dev image ~and markdown linters~ should remain in sailbot_workspace.
> 
> Edit: What do you think about removing all markdown linters and their respective configuration files from this repo and subteam repos, and putting all documentation in the docs site?

I'm fine with that. Though I think there would be a few cases where, if a README is simple enough, it should also be in the repo so that there's less switching between apps.

patrick-5546:
> I'm fine with that. Though I think there would be a few cases where, if a README is simple enough, it should also be in the repo so that there's less switching between apps.

Maybe essential information, like run instructions and arguments, can be in the README?

hhenry01:
> > I'm fine with that. Though I think there would be a few cases where, if a README is simple enough, it should also be in the repo so that there's less switching between apps.
> 
> Maybe essential information, like run instructions and arguments, can be in the README?

Yeah that was what I was thinking. Kind of like: https://github.com/UBCSailbot/network_systems/blob/main/projects/example/README.md

patrick-5546:
> Yeah that was what I was thinking. Kind of like: https://github.com/UBCSailbot/network_systems/blob/main/projects/example/README.md

Let's discuss this in the software-wide meeting today, I added a discussion topic.

patrick-5546:
Instead of deleting ROS lint and test I made them run weekly on Friday evenings (regression testing)

patrick-5546 (approved):

Issue #31, opened 2022-09-28
Support ARM computers


patrick-5546:
Push ARM versions of the docker images so that this workspace can run natively on ARM computers like the M1 Macbooks.

### Resources
- https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

patrick-5546:
Not a priority until we have someone on the team that uses an ARM computer

patrick-5546:
@hhenry01 can we use docker containers on the main computer?

hhenry01:
Yes, but there's not really necessary. Ideally we would compile the code on our own computers and copy the binaries over.

patrick-5546:
Resolve blockers:
- [ ] Installing OMPL with Python bindings: see issue 958 in ompl repo

Resources for building multi-arch docker images
- https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/
- https://billglover.me/notes/build-multi-arch-docker-images/

PR #61, opened 2022-11-23
Overhaul dev container

### LINKED ISSUE #33: Explore alternative ways to install OMPL
### Purpose
- Likely needed to #31 with minimal hassle

### Resources
- https://ompl.kavrakilab.org/download.html
- https://ompl.kavrakilab.org/installation.html
- https://ompl.kavrakilab.org/python.html

patrick-5546:
### Description

- Resolves #33 
- Breaks ability to run Raye codebase
    - I don’t foresee a need to be able to run Raye and new project code together; the architectures of the 2 projects are different
    - Would probably be a pain to figure out ROS 1 bridge
    - Would simplify devcontainer and drastically reduce its size and build time, potentially leading to performance gains

### Changelog
- Base image
    - Ubuntu 18.04 -> 22.04
    - ROS eloquent -> humble
    - OMPL 1.5.0 based off script -> master branch based off dockerfile
    - Builds in 1 hr -> 10 min
        - Still uses a lot of RAM though, 16GB+ strongly recommended
- Dev image
    - Network systems dependencies
        - clang: 10 -> 14 (default for 22.04)
        - googletest: 1.12 -> 1.11 (apt repository)
        - boost: 1.80 -> 1.74 (apt repository)
- Dev container image
    - Size: 6.41GB -> 3.77GB

### What's left
- [x] Update dev container to use the new dev image
- [x] Setup tools deprecation warning when building Python packages: add to #66 
- [x] See if ROS autocomplete is working in zsh: will tackle in another issue #65
- [x] Add ARM versions of containers: will tackle in another issue #31 
- [x] Compare with the develop branch
- [x] Compare with the humble branch of athackst/vscode_ros2_workspace
- [x] Source ROS overlay in `~/.bashrc`
- [x] Merge in #60 
- [x] Remove ament prefix from clang-tidy CI
- [x] Uncomment network_systems, comment out example repos
- [x] Merge with main
- [x] Run regression test
- [x] Create new branch named raye for what's in main before this gets merged in
    - [x] Protect branch
    - [x] Add it to workflows

### After merging
- [ ] Revert network_systems branch to main once v2-updates is merged in
- [ ] Update documentation: UBCSailbot/docs#111


.devcontainer/dev/dev.Dockerfile
```diff
@@ -0,0 +1,109 @@
...
+    clangd \
+    googletest \
+    libboost-all-dev \
+    libprotobuf-dev \
+    llvm \
+    protobuf-compiler && \
+  apt-get autoremove -y && \
+  apt-get clean -y && \
+  rm -rf /var/lib/apt/lists/{apt,dpkg,cache,log} /tmp/* /var/tmp/*
+ENV DEBIAN_FRONTEND=
```
patrick-5546, line 88: @hhenry01 How does this look to you? Will the apt version of boost and googletest suffice, they're much newer in 22.04 than 18.04. See the changelog section of this PR's description for more details.

Also, which packages are necessary for functionality? We should migrate them to the base image.

patrick-5546: Getting a cmake error, is the best course of action to downgrade cmake or upgrade boost?
image

patrick-5546: Also is it worth installing the cmake tools extension?
image

hhenry01: > Also is it worth installing the cmake tools extension? image

Nope. It serves two functions:
1. Make starting a new CMake based project easier - already done
2. Make managing different builds and running them easier - we already have scripts/tasks to do this.

hhenry01: > Getting a cmake error when building, is the best course of action to downgrade cmake or upgrade boost? image

Darn, I was hoping apt would have v1.78 or higher. If it simplifies things then we can use Boost v1.74

patrick-5546: When possible I think we should use apt repositories. If you want the features offered by newer version we can use them. To save space in the image we can build it in one container then copy over the binaries to the main container. See `.devcontainer/base/base.Dockerfile` for some examples

patrick-5546: Also is it necessary to install `libbost-all-dev` in the dev image? Here is what is installed for ompl in the base image
image

patrick-5546:
Documentation resources
- [ ] Logging into container registry: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry#authenticating-to-the-container-registry

.devcontainer/dev/dev.Dockerfile
hhenry01: I'm ok with the version of boost on apt. Does it take much longer to install `libboost-all-dev`? Finding out later down the line that a specific boost library is not installed and having to install it in the image would be quite inconvenient.

patrick-5546: > I'm ok with the version of boost on apt. 

- I created a PR for this: https://github.com/UBCSailbot/network_systems/pull/8

> Does it take much longer to install `libboost-all-dev`? Finding out later down the line that a specific boost library is not installed and having to install it in the image would be quite inconvenient.

- Should be fine

Also, which packages are necessary for functionality? We should migrate them from dev to base image.

```
    clang \
    clang-tidy \
    clangd \
    googletest \
    libboost-all-dev \
    libprotobuf-dev \
    llvm \
    protobuf-compiler && \
```

patrick-5546: Still not able to build network_systems in this branch. I updated `LLVM_VERSION` to 14 in network_systems CMakeLists.txt, but now I get this:

```
Starting >>> network_systems
--- stderr: network_systems                             
Building Network Systems with build type 'RelWithDebInfo' and flags: '-O2 -g -DNDEBUG -std=c++14 -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -pthread'
gmake[2]: *** No rule to make target '/usr/local/lib/libgtest.a', needed by 'projects/example/test_example'.  Stop.
gmake[2]: *** Waiting for unfinished jobs....
error: too many errors emitted, stopping now [clang-diagnostic-error]
/opt/ros/humble/include/rclcpp/rclcpp/client.hpp:550:10: error: no member named 'optional' in namespace 'std' [clang-diagnostic-error]
    std::optional<CallbackInfoVariant>
    ~~~~~^
/opt/ros/humble/include/rclcpp/rclcpp/client.hpp:550:19: error: use of undeclared identifier 'CallbackInfoVariant' [clang-diagnostic-error]
    std::optional<CallbackInfoVariant>
                  ^
...
```

hhenry01: > Still not able to build network_systems in this branch. I updated `LLVM_VERSION` to 14 in network_systems CMakeLists.txt, but now I get this:
> 
> ```
> Starting >>> network_systems
> --- stderr: network_systems                             
> Building Network Systems with build type 'RelWithDebInfo' and flags: '-O2 -g -DNDEBUG -std=c++14 -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -pthread'
> gmake[2]: *** No rule to make target '/usr/local/lib/libgtest.a', needed by 'projects/example/test_example'.  Stop.
> gmake[2]: *** Waiting for unfinished jobs....
> error: too many errors emitted, stopping now [clang-diagnostic-error]
> /opt/ros/humble/include/rclcpp/rclcpp/client.hpp:550:10: error: no member named 'optional' in namespace 'std' [clang-diagnostic-error]
>     std::optional<CallbackInfoVariant>
>     ~~~~~^
> /opt/ros/humble/include/rclcpp/rclcpp/client.hpp:550:19: error: use of undeclared identifier 'CallbackInfoVariant' [clang-diagnostic-error]
>     std::optional<CallbackInfoVariant>
>                   ^
> ...
> ```

Working on a commit to fix this. One cause of the error is because ROS humble requires C++17 or greater but we were building for C++14. I'll let you know when I resolve these errors.

hhenry01: I pushed commits to this branch and the v2-updates branch of network_systems to fix these issues. PR for network_systems is here: https://github.com/UBCSailbot/network_systems/pull/10

Make sure you updates src/new_project.repos so that it targets the v2-updates branch of network_systems if you want it to work.

patrick-5546: It built but print something to stderr:
```
--- stderr: network_systems                                
Building Network Systems with build type 'RelWithDebInfo' and flags: '-O2 -g -DNDEBUG -std=c++20 -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -pthread'
```

Is this an error? If not, is there a way not to print this?

hhenry01: > It built but print something to stderr:
> 
> ```
> --- stderr: network_systems                                
> Building Network Systems with build type 'RelWithDebInfo' and flags: '-O2 -g -DNDEBUG -std=c++20 -Wall -Wextra -Wpedantic -Werror -Wno-unused-parameter -pthread'
> ```
> 
> Is this an error? If not, is there a way not to print this?

Not an error. I set it up that way because printing to stdout doesn't work for some reason.

PR #94, opened 2023-02-23
Add script to run container via docker cli

### LINKED ISSUE #80: Automate building and running base image on RPi with Docker CLI
### Purpose

Our RPi needs to operate without a GUI, and just needs to run our software. Dedicated scripting to build and run will make the process a lot faster.

### Changes

New script(s) to build the container, run it, and also execute a list of programs when given the paths/commands (a default to run all software should also be defined).

### Resources

- https://docs.docker.com/engine/reference/commandline/cli/

hhenry01:
### Description

- Resolves #80 
- Created deployment_scripts/ directory as I anticipate there will be more deployment specific scripting in the future, like running a set of programs
- deployment_scripts/start_container.sh can either build and run the latest version of the workspace that is cloned, or it can run an arbitrary version built in the past
- Docker run arguments defined under the "runArgs" field in devcontainer.json are automatically parsed and used when starting the container, but there is a caveat with respect to certain devcontainer.json fields corresponding to specific arguments. I left a comment at the top of devcontainer.json warning about this.
- When running the container it is run as root. It should be easy to switch the user to ros, but I don't really see the point.

### Verification

- I ran the script locally on my computer, setup.sh, and build.sh. All succeeded without issue.
- Running the script, setup.sh, and quickbuild.sh on the raspberry pi succeeded


patrick-5546 (commented):
Quick question: does this script only work for the RPi, or can it be used on any computer (for example if you don't want to use VS Code)?

hhenry01:
> Quick question: does this script only work for the RPi, or can it be used on any computer (for example if you don't want to use VS Code)?

Works on any computer. A few bells and whistles prevent it from being 1:1 like switching the user to ros and copying config files into the container, but all essential functions are present.

patrick-5546:
Would like to see this running on the Rpi, specifically how its performance compares to your laptop. I will turn on EDC PC tomorrow.

hhenry01:
Works on rpi. Added a quick_build script for convenience. Building with it on my laptop takes ~14-16 seconds after a purging while it takes 2 min on the rpi.

patrick-5546:
Can you login to the ubcsailbotsoftware GitHub account on the rpi so that we can push quick fixes. It's probably easiest to set this up outside the container and login using a PAT

patrick-5546 (commented):
General notes:
- I think we can move `deployment_scripts/` to `.devcontainer/` so that everything related to docker is there
- Would be good to add a README in `deployment_scripts/` describing how and when to use each script (example commands, etc.)
- Does this build the dev container image? Ideally this should build the base image (same tag as dev), it is 1 GB smaller

hhenry01:
@patrick-5546  all comments have been addressed.

patrick-5546 (approved):
Made a couple changes to fit the deployment workflow I documented in the README, if it looks good to you feel free to merge

PR #114, opened 2023-04-30
Add Boat Simulator and Enhance Build Tasks

DFriend01:
### Description

Related to [this issue](https://github.com/UBCSailbot/boat_simulator/issues/2). Adds the boat simulator and modifies the build script.

The following was modified:
- Added the boat simulator to the project repositories
- Modified the build script to allow for building specific packages rather than all of them at once
- Modified the VS Code tasks:
  - `Build All` is the old `Build` task
  - `Build` task prompts for specific packages to be built

### Verification

- Sailboat Simulator:
  - Rebuilt the container to see if the boat simulator repo is cloned
- Build Script:
  - The build script ran without any arguments operates as it originally did -- it builds all packages
  - `CTRL + SHIFT + B` brings up both `Build All` and `Build` tasks as build options


patrick-5546 (changes requested):
Tested out the new build task: it works perfectly for one package, but not multiple packages.

I think this is because packages are separated by spaces, which in bash delimits arguments. My suggestion would just be to keep it simple: delete the packages task input and use the package.

I don't think it's necessary to be able to specify multiple packages to build, would be easier just build everything using the regular build task.

DFriend01:
Wait to merge until this issue is resolved:

https://github.com/UBCSailbot/boat_simulator/issues/6

build.sh
```diff
@@ -1,11 +1,24 @@
 #!/bin/bash
 set -e
 
+# Packages to build
+packages=""
```
patrick-5546, line 5: Should be singular: `package`

.devcontainer/config/sailbot_workspace.code-workspace
```diff
@@ -210,6 +213,17 @@
 				},
 				"problemMatcher": "$gcc"
 			},
+			{
+				"label": "Build Package",
+				"detail": "Build specific packages in the workspace",
```
patrick-5546, line 218: Should be singular: `Build a specific package...`

patrick-5546 (commented):
LGTM, will approve once comments resolved and CI passes

patrick-5546 (approved):
CI will pass once you merge in your boat_simulator PR and rerun failed jobs

patrick-5546 (commented):
One more idea that I had regarding the build tasks: incorporating `quick_build.sh`. There are multiple ways that this can be done, I'll let you decide what to do.

What do you think about combining quick_build.sh and build.sh by adding an argument to build.sh? Also if we keep quick_build.sh I think we should rename it build_quick.sh so that it is next to build.sh alphabetically. The deployment README/script would have to be updated accordingly.

.devcontainer/config/sailbot_workspace.code-workspace
```diff
@@ -247,6 +250,17 @@
 				},
 				"problemMatcher": "$gcc"
 			},
+			{
+				"label": "Build Package",
```
patrick-5546, line 254: Also I'm thinking that it might be better to have a dedicated task for building each package. This way you don't have to type the package name every time.

patrick-5546: Used `pickString` to interactively select package from a list

PR #115, opened 2023-05-15
Add docs site

### LINKED ISSUE #101: Add docs site
### Purpose

Able to develop docs site inside sailbot_workspace.

### Changes

- [x] Clone docs site
    - Easiest would be for this to be defined in [`new_project.repos`](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/new_project.repos), but I'm not sure if the workspace will still build with non-ROS packages in the `src/`
    - If it can't build when in `src/`, clone in `~/`
- [x] Add path to docs repository to `folders` in [`sailbot_workspace.code-workspace`](https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/config/sailbot_workspace.code-workspace)
- [ ] Add VS Code config in [`.vscode/` of docs](https://github.com/UBCSailbot/docs/tree/main/.vscode) except extensions to [`sailbot_workspace.code-workspace`](https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/config/sailbot_workspace.code-workspace)
    - Extensions would go in [`devcontainer.json`](https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/devcontainer.json)
- [x] `pip install -Ur <path to docs requirements.txt>` in [`setup.sh`](https://github.com/UBCSailbot/sailbot_workspace/blob/main/setup.sh) inside VCS if statement after VCS command
- [x] Add [apt dependencies](https://github.com/UBCSailbot/docs/blob/6b5d0aeb8ab24e8ecd8be5bca4038a787efd6e0b/.github/workflows/publish.yml#L30) to [dev image dockerfile](https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/base-dev/base-dev.Dockerfile)
- [ ] Docs: create page for docs repo with snippet of Docs README
    - Update README if needed

DFriend01:
### Description

- Resolves #101

Refer to the checklist in #101. 

I wasn't able to programmatically add the vscode settings files from the docs repo to the codespace JSON file, so I'll probably need some help there. The not so great alternative is to copy and paste.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,10 +1,29 @@
 #!/bin/bash
 set -e
 
-if [[ $DISABLE_VCS != "true" ]]
-then
+# Display warning message
+function warn() {
+    message=$1
+    echo -e "\e[1;33m${message}\e[0m"
+}
+
+# Dependency installation for websites
+function installWebsiteDependencies() {
+    # Install dependencies for docs website
+    echo "Installing docs website Python dependencies..."
+    pip3 install -Ur websites/docs/docs/requirements.txt --upgrade-strategy eager
+}
+
+# Import all project repositories
+if [[ $DISABLE_VCS != "true" ]]; then
+    echo "Importing project repositories..."
     vcs import < src/new_project.repos src
+    vcs import < websites/web.repos websites
+    installWebsiteDependencies
+else
+    warn "VCS disabled. Skipping project repository imports..."
 fi
+
 sudo apt-get update
 rosdep update
 rosdep install --from-paths src --ignore-src -y
--- a/setup.sh
+++ b/setup.sh
@@ -7,19 +7,10 @@ function warn() {
     echo -e "\e[1;33m${message}\e[0m"
 }
 
-# Dependency installation for websites
-function installWebsiteDependencies() {
-    # Install dependencies for docs website
-    echo "Installing docs website Python dependencies..."
-    pip3 install -Ur websites/docs/docs/requirements.txt --upgrade-strategy eager
-}
-
 # Import all project repositories
 if [[ $DISABLE_VCS != "true" ]]; then
     echo "Importing project repositories..."
     vcs import < src/new_project.repos src
-    vcs import < websites/web.repos websites
-    installWebsiteDependencies
 else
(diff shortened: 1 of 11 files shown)


setup.sh
```diff
@@ -1,10 +1,29 @@
...
+function warn() {
+    message=$1
+    echo -e "\e[1;33m${message}\e[0m"
+}
+
+# Dependency installation for websites
+function installWebsiteDependencies() {
+    # Install dependencies for docs website
+    echo "Installing docs website Python dependencies..."
+    pip3 install -Ur websites/docs/docs/requirements.txt --upgrade-strategy eager
```
patrick-5546, line 14: I don't think an eager upgrade strategy is needed. Upgrade strategies are for dependencies of the upgraded packages, not the upgraded packages themselves. What do you think?

From the output of `pip install --help`
```
 --upgrade-strategy <upgrade_strategy>
                              Determines how dependency upgrading should be handled [default: only-if-needed].
                              "eager" - dependencies are upgraded regardless of whether the currently installed version satisfies the requirements of the upgraded package(s).
                              "only-if-needed" -  are upgraded only when they do not satisfy the requirements of the upgraded package(s).
```

patrick-5546:
I changed the dev container to use docker compose instead of just a docker file. This allows it to run multiple containers. My plan is for each website (or other app that is purely for development purposes) to be in its own container (isolated environment). Will be talking more about this in the sailbot workspace overview meeting tonight.

patrick-5546:
One downside of the current implementation is that the repositories need to be cloned before the dev container can be built. Still thinking of the best way to automate this, but will probably not do so in this PR.

```
pip install vcs-tool
vcs import < src/new_project.repos src
```

patrick-5546:
VS Code settings should be resolved now
- cspell: uses existing file
- extensions.json: manually added them to `devcontainer.json`
- launch.json: not needed anymore
- settings.json: manually added to `sailbot_workspace.code-workspace`

patrick-5546 (approved):
Great work, thanks for the help

PR #107, opened 2023-03-22
Add potential solutions for the website

patrick-5546:
### Description

Adds the website repository and other potential solutions for the website, along with their infrastructure.

Notes:
- Changed dev container to read from a docker-compose file rather than a `Dockerfile`
- Website dependencies are installed in a new image, `local-base`
- Can't add a data source in Grafana if you are running in a GitHub Codespace

### To Do
- [x] Document `local-base` in sailbot_workspace
    - [ ] Document `local-base` in docs (add to table)
    - [ ] Add usage instructions to website README
- [x] Troubleshoot mongodb grafana data source: https://github.com/meln5674/grafana-mongodb-community-plugin
- [x] Add volume for grafana
- [x] Figure out what to do with `runArgs` (removed), as it is parsed by the deployment script
    - [ ] Update deployment script accordingly: need to add all arguments, not just CAN ones?
- [x] Switch website to start from mongodb example
- [x] Move virtual iridum dependencies to local base
- [x] Change virtual iridum repo url to end in `.git`
- [x] Rebuild docker images after merging
- [x] Change ros msgs repo url to end in `.git`
- [x] Make website a separate container? This would require deployment script to use docker compose: 119

**Open action items were copied to #119**

### Resources
- https://github.com/microsoft/vscode-dev-containers/tree/main/containers/javascript-node-mongo

PR #116, opened 2023-05-22
Use docker compose

patrick-5546:
### Description

Make the dev container use docker compose. Previously, it was built on `.devcontainer/Dockerfile`. Now it is built on `.devcontainer/docker-compose.yml`, which builds the dockerfile. Using docker compose allows the dev container to run multiple images.

### Verification

- [x] Rebuild dev container locally

### Resource
- https://github.com/devcontainers/templates/tree/main/src/javascript-node-mongo


patrick-5546:
@hhenry01 FYI see the first commit. We previously discussed that the dev container doesn't need to see CAN, as the mocks won't be using VCAN.

Overall plan is to refactor https://github.com/UBCSailbot/sailbot_workspace/pull/107 into multiple smaller PRs, the first of which is this one.

PR #129, opened 2023-05-27
Add more workflow input arguments and fix linting workflow

DFriend01:
### Description

Two things to note:

- I split up the `ament_lint` job into 4 separate jobs. You cannot reference matrix values in a `job.if` statement for some reason, so this was the best I could do to apply the input variables to the linters.
- I turned off linting in sailbot workspace. Not all repositories use all linters, so you will constantly get failing CI. I know you probably wanted to run linting on all repos every time, but unless we find a more elegant solution to this, disabling linting is the only clear thing to me at the moment.

### Verification

- I have been testing with a PR in the boat simulator here: https://github.com/UBCSailbot/boat_simulator/pull/8

### DIFF
--- a/.github/actions/ament-lint/run.sh
+++ b/.github/actions/ament-lint/run.sh
@@ -1,9 +1,72 @@
 #!/bin/bash
 set -e
 
+# Display error message
+function error {
+    message=$1
+    echo -e "\e[1;31m${message}\e[0m"
+}
+
+# Display warning message
+function warn {
+    message=$1
+    echo -e "\e[1;33m${message}\e[0m"
+}
+
+function do_lint_cmake {
+    VALID_SRC_DIRS_ARG=$1
+    LINTED_FILES=`find ${VALID_SRC_DIRS_ARG} -type f \( -name "CMakeLists.txt" -o -name "*.cmake" -o -name "*.cmake.in" \)`
+
+    if [[ ! -z ${LINTED_FILES} ]]; then
+        ament_lint_cmake $VALID_SRC_DIRS_ARG
+    else
+        warn "WARNING: No files found for lint_cmake. Skipping lint_cmake..."
+    fi
+}
+
+function do_flake8 {
+    VALID_SRC_DIRS_ARG=$1
+    LINTED_FILES=`find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"`
+
+    if [[ ! -z ${LINTED_FILES} ]]; then
+        ament_flake8 $VALID_SRC_DIRS_ARG
+    else
+        warn "WARNING: No files found for flake8. Skipping flake8..."
+    fi
+}
+
+function do_pep257 {
+    VALID_SRC_DIRS_ARG=$1
+    LINTED_FILES=`find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"`
+
+    if [[ ! -z ${LINTED_FILES} ]]; then
+        ament_pep257 $VALID_SRC_DIRS_ARG
+    else
+        warn "WARNING: No files found for pep257. Skipping pep257..."
+    fi
+}
+
+function do_xmllint {
+    VALID_SRC_DIRS_ARG=$1
+    LINTED_FILES=`find ${VALID_SRC_DIRS_ARG} -type f -name "*.xml"`
+
+    if [[ ! -z ${LINTED_FILES} ]]; then
+        ament_xmllint $VALID_SRC_DIRS_ARG
+    else
+        warn "WARNING: No files found for xmllint. Skipping xmllint..."
+    fi
+}
+
 source /opt/ros/${ROS_DISTRO}/setup.bash
 ./setup.sh
 cd src
+
 # exclude virtual_iridium because it is a legacy library
 VALID_SRC_DIRS=$(ls | grep -v -e virtual_iridium -e new_project.repos)
-ament_${LINTER} $VALID_SRC_DIRS
+case ${LINTER} in
+    lint_cmake) do_lint_cmake $VALID_SRC_DIRS ;;
+    flake8) do_flake8 $VALID_SRC_DIRS ;;
+    pep257) do_pep257 $VALID_SRC_DIRS ;;
+    xmllint) do_xmllint $VALID_SRC_DIRS ;;
+    *) error "ERROR: Invalid linter ${LINTER} specified in ament-lint action"; exit 1 ;;
+esac
--- a/.github/actions/ament-lint/run.sh
+++ b/.github/actions/ament-lint/run.sh
@@ -41,6 +41,6 @@ source /opt/ros/${ROS_DISTRO}/setup.bash
 ./setup.sh
 cd src
 
-# Exclude virtual_iridium because it is a legacy library and docs
-VALID_SRC_DIRS=$(ls | grep -v -e virtual_iridium -e docs -e new_project.repos)
+# Exclude repos and files we don't want to lint
+VALID_SRC_DIRS=$(ls | grep -v -e virtual_iridium -e docs -e raye-local-pathfinding -e new_project.repos)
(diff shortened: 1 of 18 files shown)


DFriend01:
@patrick-5546 CI checks fail because adding new parameters and referencing them break the current CI. I think it's still referencing the CI in main.

patrick-5546:
> @patrick-5546 CI checks fail because adding new parameters and referencing them break the current CI. I think it's still referencing the CI in main where the new parameters don't exist.

What you can do is change this PR to reference your sailbot workspace PR's branch, then create a separate PR here after the sailbot workspace PR is merged in to switch it back to the main branch

DFriend01:
The previously mentioned issues have been resolved. The `ament-lint` action has been refactored to search for files before linting to avoid running the linters with no files, which causes errors if no files are specified.

The `ament-lint` workflow has been reverted to its previous implementation. The refs were updated to this branch, so we need to make a follow-up PR to change it back to main.

patrick-5546 (commented):
~Discussed over call that we should explore controlling what ament_lint jobs run at the GitHub workflow level. One possible implementation is that the caller workflow passes the linter matrix as an input to the reusable workflow, but inputs do not have a matrix type. My recommendations for how to explore this are on Slack~ See review below

.github/workflows/test_definitions.yml
```diff
@@ -15,18 +15,36 @@ on:
...
+        default: true
+        type: boolean
+      markdownlint:
+        required: false
+        default: true
+        type: boolean
+      markdown-link-check:
+        required: false
+        default: true
+        type: boolean
```
patrick-5546, line 39: Reduce inputs to `ros-ci` and `markdown-ci`

.github/actions/ament-lint/run.sh
```diff
@@ -1,9 +1,46 @@
...
+function warn {
+    message=$1
+    echo -e "\e[1;33m${message}\e[0m"
+}
+
+function get_search_command {
+    case ${LINTER} in
+        lint_cmake) CMD='find ${VALID_SRC_DIRS_ARG} -type f \( -name "CMakeLists.txt" -o -name "*.cmake" -o -name "*.cmake.in" \)' ;;
+        flake8) CMD='find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"' ;;
+        pep257) CMD='find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"' ;;
```
patrick-5546, line 20: Can delete this line, not using pep257 anymore

.github/actions/ament-lint/run.sh
```diff
@@ -1,9 +1,46 @@
...
+    echo -e "\e[1;31m${message}\e[0m"
+}
+
+# Display warning message
+function warn {
+    message=$1
+    echo -e "\e[1;33m${message}\e[0m"
+}
+
+function get_search_command {
```
patrick-5546, line 16: Add short description for function

run_clang-tidy.sh
```diff
@@ -1,3 +1,5 @@
 files=$(find ./src/network_systems/ -type f \( -iname \*.cpp -o -iname \*.h \))
 
-clang-tidy -p ./build/ --format-style=file $files
+if [[ ! -z ${files} ]]; then
+    clang-tidy -p ./build/ --format-style=file $files
+fi
```
patrick-5546, line 5: Not needed anymore, just set clang-tidy input to false if not needed

.github/actions/ament-lint/run.sh
```diff
@@ -1,9 +1,46 @@
...
+        flake8) CMD='find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"' ;;
+        pep257) CMD='find ${VALID_SRC_DIRS_ARG} -type f -name "*.py"' ;;
+        xmllint) CMD='find ${VALID_SRC_DIRS_ARG} -type f -name "*.xml"' ;;
+        *) error "ERROR: Invalid linter ${LINTER} specified in ament-lint action"; exit 1 ;;
+    esac
+
+    echo $CMD
+}
+
+function lint {
```
patrick-5546, line 28: Short description

.github/workflows/test_definitions.yml
```diff
@@ -6,53 +6,56 @@ on:
...
-
-      - name: Test
-        uses: ./.github/actions/test/
-        env:
-          DISABLE_VCS: ${{ inputs.disable-vcs }}
+    - name: Checkout workspace
+      uses: actions/checkout@v3
+      with:
+          repository: UBCSailbot/sailbot_workspace
+          ref: user/dfriend01/add_lang_ci_args
```
patrick-5546, line 38: Is checking out to this branch necessary?

.github/workflows/test_definitions.yml
```diff
@@ -62,65 +65,69 @@ jobs:
       matrix:
           # ament_lint_common, except for copyright, cppcheck, cpplint, uncrustify, and pep257
           linter: [lint_cmake, flake8, xmllint]
+    if: ${{ inputs.ros-ci }}
```
patrick-5546, line 68: Can if condition be after `run-on` like other jobs?

.github/workflows/test_definitions.yml
```diff
@@ -6,53 +6,56 @@ on:
...
-        required: true
-        type: string
       repository:
         required: true
         type: string
       rebuild-docs:
         required: true
         type: boolean
+      markdown-ci:
+        required: false
```
patrick-5546, line 22: My perspective is to make everything required so that all arguments can be found in tests.yml and all tests.yml should look the same. What do you think?

patrick-5546: But I also understand it's a pain to update every test.yml file every time an input is added/removed. We could mitigate this by pointing the tests.yml's only to dev branches, not main

.github/workflows/test_definitions.yml
patrick-5546: I guess it is to test the changes to the scripts you made

.github/workflows/test_definitions.yml
DFriend01: I like the idea of a separate repo and just letting dependabot handle it.

patrick-5546: Me too. In that case I think it's fine to leave the PR as is. Can you write an issue for this?

.github/workflows/test_definitions.yml
```diff
@@ -6,53 +6,56 @@ on:
...
+      uses: actions/checkout@v3
+      with:
+          repository: UBCSailbot/custom_interfaces
+          ref: user/colinli02/custom_messages
+          path: src/custom_interfaces
+
+    - name: Test
+      uses: ./.github/actions/test/
+      env:
+        DISABLE_VCS: ${{ inputs.repository != github.event.repository.name }}
```
patrick-5546, line 58: Did you check that this works when it's true and when it's false?

.github/workflows/test_definitions.yml
patrick-5546: Actually we don't even need to put the workflow in its own repo, we can reference a release instead of a branch. Dependabot supports this: https://github.blog/changelog/2023-03-13-dependabot-updates-support-reusable-workflows-for-github-actions/

.github/workflows/test_definitions.yml
DFriend01: It has been confirmed that bash can accept booleans and can still interpret them as strings in the script.

.github/workflows/test_definitions.yml
DFriend01: We are planning to use versioning instead as noted by https://github.com/UBCSailbot/sailbot_workspace/pull/129#discussion_r1225587949

patrick-5546 (approved):
Great work on this PR

PR #137, opened 2023-06-11
Fix rebuild docs

patrick-5546:
Need to update the workflow file rebuild-docs is running after changes to the CI in the docs site from https://github.com/UBCSailbot/docs/pull/144


patrick-5546:
Pending https://github.com/UBCSailbot/docs/pull/146

PR #155, opened 2023-07-25
Specify interpreter path for python extensions

DFriend01:
### Description

Add path to python interpreter in workspace config file for black formatted. Without it, black would try to make assumptions on which interpreter to use and may get it wrong.


patrick-5546 (changes requested):
Great catch, didn't know this could be a problem. We should also set the interpreter manually for the rest of the Python extensions: isort, flake8, and mypy (in #156)

patrick-5546 (approved):

PR #171, opened 2023-10-02
Add local pathfinding python dependencies

patrick-5546:
Install pyproj and shapely


patrick-5546:
Pending refactoring of images to decrease build times

PR #173, opened 2023-10-03
Fix base/dev image long build times

patrick-5546:
### Description

- Since #161, base/dev image build times went from 17 min to 2 hrs+
- The cause of this was building `mongo-cxx-driver` from source
- Since this is needed for everything, moved to the pre-base image
- Effect: base/dev image build times are back to 17 min, and reduced their size by 100mb


patrick-5546:
@hhenry01 FYI

PR #178, opened 2023-10-14
ROS Parameters: Add boat simulator ROS parameters

chrischang5:
### Description

- Resolves https://github.com/UBCSailbot/boat_simulator/issues/37
- Implementation of reading these parameters in the `boat_simulator` is done in https://github.com/UBCSailbot/boat_simulator/pull/43

### Verification

- [ ] See https://github.com/UBCSailbot/boat_simulator/pull/43 for verification steps of reading in parameters in the `boat_simulator`

### Resources

- [ROS2 YAML parameter structure](https://roboticsbackend.com/ros2-yaml-params/#Why_using_YAML_for_ROS2_params)
- [ROS2 Parameter docs](https://docs.ros.org/en/humble/Concepts/Basic/About-Parameters.html)

### DIFF
--- a/src/global_launch/config/globals.yaml
+++ b/src/global_launch/config/globals.yaml
@@ -2,3 +2,33 @@
    ros__parameters:
       # Publishers' period (seconds): how often the publishers publish
       pub_period_sec: 0.5
+
+# boat_simulator parameters
+low_level_control_node:
+   ros__parameters:
+      rudder:
+         disable_actuation: false
+         fixed_angle_deg: 0.0 # Only used if rudder actuation is disabled
+         actuation_request_period_sec: 10.0
+         actuation_execution_period_sec: 0.5
+         pid: # Only used if rudder actuation is enabled
+            kp: 0.0
+            ki: 0.0
+            kd: 0.0
+            buffer_size: 50
+      wingsail:
+         disable_actuation: false
+         fixed_angle_deg: 0.0 # Only used if actuation is disabled
+         actuation_request_period_sec: 10.0
+         actuation_execution_period_sec: 0.5
+         actuation_speed_deg_per_sec: 0.1 # Only used if actuation is enabled
+physics_engine_node:
+   ros__parameters:
+      wind_sensor:
+         generator_type: gaussian # One of "gaussian", "constant"
+         gaussian_params:
+            mean: [1.0, 0.0] # Mean for x and y components
+            std_dev: [0.1, 0.1] # Standard deviation for x and y components
+            corr_xy: 0.0
+         constant_params:
+            value: [1.0, 0.0] # Constant value for x and y components
--- a/src/global_launch/config/globals.yaml
+++ b/src/global_launch/config/globals.yaml
@@ -6,6 +6,7 @@
 # boat_simulator parameters
 low_level_control_node:
    ros__parameters:
+      logging_throttle_period_sec: 2.0
       rudder:
          disable_actuation: false
          fixed_angle_deg: 0.0 # Only used if rudder actuation is disabled
@@ -24,6 +25,7 @@ low_level_control_node:
          actuation_speed_deg_per_sec: 0.1 # Only used if actuation is enabled
 physics_engine_node:
    ros__parameters:
+      logging_throttle_period_sec: 2.0
       wind_sensor:
          generator_type: gaussian # One of "gaussian", "constant"
          gaussian_params:
--- a/src/global_launch/config/globals.yaml
+++ b/src/global_launch/config/globals.yaml
@@ -10,8 +10,7 @@ low_level_control_node:
       rudder:
          disable_actuation: false
          fixed_angle_deg: 0.0 # Only used if rudder actuation is disabled: [0.0, 360.0]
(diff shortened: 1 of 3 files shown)


patrick-5546 (approved):
LGTM infrastructure wise, I'll trust @DFriend01 to verify that it works

src/global_launch/config/globals.yaml
DFriend01, line 1: When using interval notation, let's try to be more mathematically specific (i.e. use `(` for non-inclusive boundaries and `[` for inclusive boundaries). Remember for infinity, use `(` (i.e. for `DOUBLE_MAX`, `DOUBLE_MIN`)

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
       # Publishers' period (seconds): how often the publishers publish
       pub_period_sec: 0.5
+
+# boat_simulator parameters
+low_level_control_node:
+   ros__parameters:
+      logging_throttle_period_sec: 2.0 # [0.0, MAX_DOUBLE]
+      rudder:
+         disable_actuation: false
+         fixed_angle_deg: 0.0 # Only used if rudder actuation is disabled: [0.0, 360.0]
```
DFriend01, line 12: Range should be [-45, 45]

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
+         actuation_execution_period_sec: 0.5 # How often the rudder action server routine's main loop executes: [0.0, MAX_DOubLE]
```
DFriend01, line 13: Should not include zero

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
+         pid: # Only used if rudder actuation is enabled
+            kp: 0.0 # [0.0, MAX_DOUBLE]
+            ki: 0.0 # [0.0, MAX_DOUBLE]
+            kd: 0.0 # [0.0, MAX_DOUBLE]
+            buffer_size: 50 # [0, MAX_INT]
```
DFriend01, line 18: Should not include zero (start from 1)

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
+      wingsail:
+         disable_actuation: false
+         fixed_angle_deg: 0.0 # Only used if actuation is disabled: [0.0, 360.0]
```
DFriend01, line 21: Let's make the range symmetric => [-180, 180)

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
+         actuation_execution_period_sec: 0.5 # How often the sail action server routine's main loop executes: [0.0, MAX_DOUBLE]
+         actuation_speed_deg_per_sec: 0.1 # Only used if actuation is enable: [0.0, MAX_DOUBLE]
```
DFriend01, line 23: Should not include zero

src/global_launch/config/globals.yaml
```diff
@@ -2,3 +2,37 @@
...
+physics_engine_node:
+   ros__parameters:
+      rudder:
+         actuation_request_period_sec: 10.0 # How often the rudder action client requests a rudder actuation: [0.0, MAX_DOUBLE]
+      wingsail:
+         actuation_request_period_sec: 10.0 # How often the sail action server requests a wingsail actuation: [0.0, MAX_DOUBLE]
```
DFriend01, line 29: Should not include zero

DFriend01 (changes requested):
Just some comments on the parameter ranges.

src/global_launch/config/globals.yaml
chrischang5: Will update value to reflect this then.

DFriend01:
@chrischang5 I realize you are working on this right now, but can you create a `README.md` file inside of the `config` directory instead, of defining the configuration parameters?

For each parameter, can you add:
- An explanation of what the parameter is
- Its type
- The range of acceptable values

The config file is starting to look a little messy, so perhaps this would be a better approach to documenting these values. What do you think?

chrischang5:
> @chrischang5 I realize you are working on this right now, but can you create a `README.md` file inside of the `config` directory instead of the configuration file itself?
> 
> For each parameter, can you add:
> 
> * An explanation of what the parameter is
> * Its type
> * The range of acceptable values
> 
> The config file is starting to look a little messy, so perhaps this would be a better approach to documenting these values. What do you think? I trust that you can devise an acceptable format for the README.

README seems like a better solution. One drawback is that it's a separate place we need to maintain alongside the config file, but I think its' better than a messy config file.

chrischang5:
We should spend more time in the design phase next time to avoid changing implementations all the time. Might have made this process smoother.

DFriend01:
I updated the configuration parameter descriptions. @patrick-5546 did you want to take a quick look?

patrick-5546:
Looks great! A couple of suggestions
- Add the README to the docs site
- Alphabetize the parameters

DFriend01 (approved):

PR #185, opened 2023-11-03
Fix test script

hhenry01:
### Description

- Resolves #_issue_

### Verification

- [ ] 

### Resources

-


patrick-5546:
- Can verify it works for repos other than NET in this PR: https://github.com/UBCSailbot/local_pathfinding/pull/47

hhenry01:
Properly fixing it requires a change in network_systems. But I can't push network_systems fix while sailbot_workspace is broken, so I'm reverting the commit first.

patrick-5546 (approved):

PR #192, opened 2023-11-04
Revert "Fix sailbot workspace ref in CI"

patrick-5546:
Reverts UBCSailbot/sailbot_workspace#191, did not work


patrick-5546:
- Fixed in #193

PR #200, opened 2023-11-08
Add timeout to each CI test

patrick-5546:
### Description
Add a timeout for each job, because if something runs for much longer than usual there's probably something wrong.

@hhenry01 do these timeouts look reasonable to you?

### Resources
- https://stackoverflow.com/a/59076067

### DIFF
--- a/.github/workflows/build-images.yml
+++ b/.github/workflows/build-images.yml
@@ -16,6 +16,7 @@ concurrency:
 jobs:
   docker:
     runs-on: ubuntu-latest
+    timeout-minutes: 30
 
     steps:
       - name: Set up QEMU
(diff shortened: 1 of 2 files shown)


hhenry01:
Is that enough time for build-images? Prebase takes like 3 hours, is it not included in the timeout?

patrick-5546:
> Is that enough time for build-images? Prebase takes like 3 hours, is it not included in the timeout?

Build images only builds base and dev. For pre-base you have to build locally and push yourself.

hhenry01:
> > Is that enough time for build-images? Prebase takes like 3 hours, is it not included in the timeout?
> 
> Build images only builds base and dev. For pre-base you have to build locally and push yourself.

In that case LGTM. I just need to make sure clang-tidy doesn't balloon out of control.

hhenry01 (approved):

patrick-5546:
It currently takes ~20 minutes to build base and dev in CI: https://github.com/UBCSailbot/sailbot_workspace/actions/workflows/build-images.yml

Although we technically have unlimited minutes for GitHub actions, that may change in the future and the time saved > convenience of CI in this case IMO

PR #206, opened 2023-11-12
Add Controller to Global Launch File

DFriend01:
### Description

The controller was added to the list of launchable ROS packages in the global launch file. This PR is closely related to the following PR for the controller: https://github.com/UBCSailbot/controller/pull/10

### Verification

- The entire system, including the controller, is started. The desired debug statements are observed from the controller when the `log_level` argument is set to `debug`.


patrick-5546 (approved):
LGTM

We should add some CI to run the launch files to check that everything runs. Maybe it will be required for then integration tests that @hhenry01 is working on

PR #210, opened 2023-11-13
Add support for jupyter notebooks

patrick-5546:
@DFriend01 fyi


patrick-5546:
I noticed that you have to restart the kernel after installing pip packages to import them, so there's still a small manual step required.

PR #217, opened 2023-12-03
Support multiple global launch configs

### LINKED ISSUE #215: Allow Global Launch to Accept Multiple Configs
### Purpose

Currently, global_launch only accepts a `.yaml` that sets every ros parameter. A more modular approach would be to allow multiple config files that can override defaults, allowing for more flexibility in testing.

### Changes

- Make it standard to define default ros parameters in ros node code
- Config `.yaml` files only exist to override default parameters

### Resources

-

hhenry01:
### Description

- Resolves #215 
Allows us to specify multiple `.yaml` files when doing `ros2 launch`. What this does is allow us to specify a base configuration and allow for overriding of specific nodes. For example, if we do `ros2 launch network_systems main_launch.py config:=config1.yaml,config2.yaml`, then:
1. globals.yaml is loaded and processed
2. config1.yaml is loaded and overrides any duplicate parameters with globals.yaml
3. config2.yaml is loaded and override previous duplicate parameters
Non-duplicate parameters are unchanged.

To take advantage of this overriding structure, main_launch.py in subteam repos should be updated such that:
- inside `get_global_launch_arguments()` assign a global variable `global_launch_config = module.GLOBAL_LAUNCH_CONFIG`
- In each `get_x_description()`, assign `ros_parameters = [global_launch_config, *LaunchConfiguration("config").perform(context).split(","), <whatever else>]`

### Verification

- Tested with this network_systems PR: https://github.com/UBCSailbot/network_systems/pull/80

### Resources

-

### DIFF
--- a/src/global_launch/main_launch.py
+++ b/src/global_launch/main_launch.py
@@ -22,10 +22,11 @@ DEVELOPMENT_ROS_PACKAGES = ["controller", "boat_simulator", "local_pathfinding",
 ROS_PACKAGES_DIR = os.path.join(
     os.getenv("ROS_WORKSPACE", default="/workspaces/sailbot_workspace"), "src"
 )
+GLOBAL_LAUNCH_CONFIG = os.path.join(ROS_PACKAGES_DIR, "global_launch", "config", "globals.yaml")
 GLOBAL_LAUNCH_ARGUMENTS = [
     DeclareLaunchArgument(
         name="config",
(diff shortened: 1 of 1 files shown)


patrick-5546 (approved):
Works for me, @DFriend01 what do you think? Should we make the argument a path or a filename in `global_launch/config`?

PR #223, opened 2023-12-10
Added a write parameter for the Mock Global Path node

SPDonaghy:
### Description

Linked to Issue [49](https://github.com/UBCSailbot/local_pathfinding/issues/49) in Local Pathfinding.

The purpose was to add a parameter to determine whether the generate_path function in node_mock_global_path.py should write a generated path to a new csv file.

The added parameter is:
`write`

### Verification
1. Run the global or local pathfinding launch file with development mode (default)
    ```
    ros2 launch local_pathfinding main_launch.py
    ```
2. In a different terminal run some ros2 param commands
    ```
    ros2 param get /mgp_main global_path_filepath
    ros2 param set /mgp_main global_path_filepath /workspaces/sailbot_workspace/src/local_pathfinding/global_paths/path_2.csv
    ```

### Resources

- [Link](https://github.com/UBCSailbot/local_pathfinding/pull/61) to the PR for this issue.

### DIFF
--- a/src/global_launch/config/README.md
+++ b/src/global_launch/config/README.md
@@ -43,7 +43,7 @@ ROS parameters specific to the nodes in the local_pathfinding package.
 **`interval_spacing`**
 
 - _Description_: The upper bound on spacing between each point in the global path in km.
-- _Datatype_: `float`
+- _Datatype_: `double`
 - _Range_: `(0.0, MAX_DOUBLE)`
 
 **`write`**
--- a/src/global_launch/config/README.md
+++ b/src/global_launch/config/README.md
@@ -52,6 +52,12 @@ ROS parameters specific to the nodes in the local_pathfinding package.
 - _Datatype_: `boolean`
 - _Acceptable Values_: `true`, `false`
 
+**`gps_threshold`**
+
+- _Description_: A new path will be generated if the GPS position changed by more than this amount in km.
+- _Datatype_: `double`
+- _Acceptable Values_: `(0.0, MAX_DOUBLE)`
+
 ## Boat Simulator Parameters
 
 ROS parameters specific to the nodes in the boat simulator.
--- a/src/global_launch/config/README.md
+++ b/src/global_launch/config/README.md
@@ -58,6 +58,12 @@ ROS parameters specific to the nodes in the local_pathfinding package.
 - _Datatype_: `double`
 - _Acceptable Values_: `(0.0, MAX_DOUBLE)`
 
+**`force`**
+
+- _Description_: Force the mock global path callback to update the global path when set to true.
+- _Datatype_: `boolean`
+- _Acceptable Values_: `true`, `false`
+
 ## Boat Simulator Parameters
 
 ROS parameters specific to the nodes in the boat simulator.
--- a/src/global_launch/config/README.md
+++ b/src/global_launch/config/README.md
@@ -54,9 +54,9 @@ ROS parameters specific to the nodes in the local_pathfinding package.
 
 **`gps_threshold`**
 
-- _Description_: A new path will be generated if the GPS position changed by more than this amount in km.
+- _Description_: A new path will be generated if the GPS position changed by more thangps_threshold*interval_spacing.
 - _Datatype_: `double`
-- _Acceptable Values_: `(0.0, MAX_DOUBLE)`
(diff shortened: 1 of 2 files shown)


patrick-5546 (approved):

patrick-5546:
I rebased this branch so that it only contains new commits that aren't already in other PR's that were squashed and merged into main

SPDonaghy:
How much do you think the GPS position should change for us to want to generate a new global path? I am just wanting to set a default value for a gps_threshold param in the yaml file

patrick-5546:
> How much do you think the GPS position should change for us to want to generate a new global path?

@jamenkaye any ideas? Some ideas that come to mind:

- If the current position is greater than `x * interval_spacing` away from the closet global waypoint
- Time based, every `x` minutes

And just thought of this, is there a way to force update the path even if it stays the same? I guess saving the file would do this?

SPDonaghy:
> And just thought of this, is there a way to force update the path even if it stays the same? I guess saving the file would do this?

Could we implement a service to run global_path_callback() on request, and request it from CLI with rosservice?

Saving the csv file would also do it though.

patrick-5546:
> Could we implement a service to run global_path_callback() on request, and request it from the command line with rosservice?

Ooh interesting idea, I think that could work. As long as it's not too complicated, since saving the file already works. Also would want to run some version of the callback without all the checks right

SPDonaghy:
> * If the current position is greater than `x * interval_spacing` away from the closet global waypoint

Should we skip adding another parameter, and just use interval_spacing as the threshold distance?

patrick-5546:
> Should we skip adding another parameter, and just use interval_spacing as the threshold distance?

I'm just worried if we use interval spacing directly then we could run into some trouble if right when we reach one global waypoint we go back a bit, which would cause the path to regenerate

SPDonaghy:
okay, sounds good, I will keep it in there

patrick-5546:
> If the current position is greater than `x * interval_spacing` away from the closet global waypoint

This idea would make `x` the parameter and have it be some float > 1 so that it is a function of `interval_spacing`. What do you think about that?

SPDonaghy:
> This idea would make `x` the parameter and have it be some float > 1 so that it is a function of `interval_spacing`. What do you think about that?

Sounds good! I made it 2 for now

jamenkaye:
Sorry I'm a little late to the party! I wanted to focus on exams.

I like the `x` parameter and `inteval_spacing` idea! I think something around 1-1.5 should be good. If $x \geq 1$, this lets us sail at no more than 60° upwind (which is pretty easy) and use only 1 tack to get to the next waypoint, even if the global path is heading straight upwind.

image

(Since the angles are all 60°, its an equilateral triangle so the spacing between waypoints is the same as the distance from the boat at the rightmost corner to the waypoint)

I'm a little unclear on the context for this feature though. Are we still planning to manually send the global path in deployment, or is this an autonomous global pathfinding that just plots a straight line? If its the latter, that would seem a little silly considering we already have a fancy global pathfinding that uses weather data too.

SPDonaghy:
> or is this an autonomous global pathfinding that just plots a straight line? If its the latter, that would seem a little silly considering we already have a fancy global pathfinding that uses weather data too.

Aside from the mgp node, what I've made so far is a custom path builder, and an interpolation function that fills in any intervals (which are more than interval_spacing apart) with a straight, evenly spaced sub path. While I think everything I've built so far is just inteded for mocks, I think there could be a use case, for what I made, in deployment, if we want to specify more complicated paths than just start->end, for example.

jamenkaye:
Ah makes sense, thanks for explaining! 👍

src/global_launch/config/README.md
```diff
@@ -43,9 +43,27 @@ ROS parameters specific to the nodes in the local_pathfinding package.
...
+- _Datatype_: `double`
 - _Range_: `(0.0, MAX_DOUBLE)`
 
+**`write`**
+
+- _Description_: Whether or not to write a generated global path to a new csv file.
+- _Datatype_: `boolean`
+- _Acceptable Values_: `true`, `false`
+
+**`gps_threshold`**
```
patrick-5546, line 55: Description and acceptable values need to be updated, as well as a log message in node_mock_global_path.py

PR #224, opened 2023-12-14
Integration Test Framework

### LINKED ISSUE #142: Design workspace-wide ROS infrastructure
### Purpose

At some point we will have to run packages from other subteams to achieve our goals of being able to fully simulate sailbot in software.

### Changes

- Integration/system tests
- Launch files

### Resources

- https://docs.ros.org/en/humble/The-ROS2-Project/Contributing/Developer-Guide.html#linters-and-static-analysis
- https://docs.ros.org/en/rolling/Tutorials/Intermediate/Launch/Using-ROS2-Launch-For-Large-Projects.html#top-level-organization
- https://github.com/ros2/launch/tree/master/launch_testing
- https://github.com/ros2/launch_ros/tree/rolling/launch_testing_ros

hhenry01:
### Description

- Resolves #142 

#### Config Files
This is where we define our tests. 

`required_packages` is basically a list of stuff we want to have running for the test, and any extra config options/files we want to give it. Non-ROS executables like virtual_iridium and the website are to be put here too, they'll just be handled explicitly.

`inputs` is a list of inputs. I'm thinking the `type` can be something like ROS or HTTP, the `name` is the topic name or URL in the case of HTTP, and `body` is the actual input data. The `body` will need to be processed differently depending on the `type`, but we should be able to generically set key value pairs based on the `type`. 

`expected_outputs` is formatted the exact same way as the `inputs`, just on the other end.

#### Test Flow
1. Run the python script with a config file.
2. Start all specified packages.
3. Start monitoring output topics.
4. Send all inputs.
5. Over the span of 3(?) seconds, check that the expected outputs appear.

### Verification

- [ ] 

### Resources

-


src/integration/run.py
patrick-5546, line 1: What do you think about integrating this into `src/global_launch/main_launch.py`

src/integration/test.yaml
patrick-5546, line 1: What do you think about using ROS parameters and moving this to `src/global_launch/config/`?

patrick-5546 (commented):
I think it would be nice if our integrations tests just ran our global launch file. We can have test-only nodes and parameters if needed

src/integration/run.py
hhenry01: I don't see a problem with importing a `run()` function that just does what this file does into main_launch.py

src/integration/test.yaml
hhenry01: That could work. I'll try it out. I just didn't want to deal with ROS boilerplate for the POC haha.

hhenry01:
> I think it would be nice if our integrations tests just ran our global launch file. We can have test-only nodes and parameters if needed

Right now I have it so the integration test file runs the `ros2 launch ...` command to start up our nodes. I'll see how launching nodes from a node will look and hopefully it won't be super janky.

src/integration/test.yaml
```diff
@@ -0,0 +1,16 @@
+
+required_packages:
```
hhenry01, line 2: New complication. Is it even possible for ROS to parse a yaml with an arbitrary sized list of parameters?

hhenry01: Would it be too scuffed if the ros_parameter is just the name of a custom yaml config file for the test?

patrick-5546: An alternative to an arbitrary sized list is a flag for each package

hhenry01: I implemented dynamic typing based on the config file fields so there's no way using ROS parameters for the test config file is gonna work.

hhenry01:
I've got everything I consider as essential features implemented. One nice-to-have would be to allow ranges for expected values, seeing as we can't be exact with some of our expectations. This would be a fair bit of extra work though so I have a DONT_CARE key for such scenarios instead.

I still need to work on the documentation. My naming conventions are also a mess because I was loosely following the style used in boat_simulator and controller but didn't read any style guidelines :P

For getting it to work with ros2 launch I'll need to look into how Python ROS packages work.

patrick-5546:
@DFriend01 as the ros launch expert can you help out with this?

DFriend01:
It's not a requirement for a launch file to be registered with a package. However, I am fairly sure that it is required to have a node to be registered with a package (i.e. the integration testing node in this case).

We probably need to create a ROS package for integration tests if we want to use the ROS CLI when using this integration test framework (`ros2 run`, `ros2 launch`, etc.)

hhenry01:
> It's not a requirement for a launch file to be registered with a package. However, I am fairly sure that it is required to have a node to be registered with a package (i.e. the integration testing node in this case).
> 
> We probably need to create a ROS package for integration tests if we want to use the ROS CLI when using this integration test framework (`ros2 run`, `ros2 launch`, etc.)

Because of this, I'll move the integration test files back into a separate folder from global_launch. Having it in the same folder makes global_launch too messy imo, and makes it harder to understand global_launch itself.

hhenry01:
I think it would be best to merge the PR in now as is and implement HTTP and website support in a follow up PR. It'd probably be worthwhile to present this in a team meeting as well to get opinions from website and global pathfinding. I'm also not sure how to best integrate this with the CI, but the PR is big enough as it is.

Have fun reviewing the 1100 line diff.

patrick-5546:
Yea can you present this in the first software-wide meeting back on jan 13? Do you want me to review before then?

hhenry01:
> Yea can you present this in the first software-wide meeting back on jan 13? Do you want me to review before then?

If you have the time that would be beneficial so that it's more accessible for people to view.

hhenry01:
Added a diagram outlining the structure. It renders as:
image

.gitignore
```diff
@@ -5,6 +5,10 @@
 
 # vcstool packages
 /src/*
+!/src/integration_tests
+/src/integration_tests/integration_tests/ros_dtypes.py
+/src/integration_tests/config
+!/src/integration_tests/generate_configs.sh
```
patrick-5546, line 11: I don't see this file. Is it because it is the wrong filepath, it's in the config directory right

src/integration_tests/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+For example:
+
+```shell
+ros2 run integration_tests run --ros-args -p testplan:=testplans/example.yaml
+```
+
+## Invoking with ros2 launch
+
+Since the custom testplan `.yaml` files are not valid ROS parameter config files, using them with `ros2 launch` requires
```
patrick-5546, line 54: Could this be done dynamically? Do what this script does when launching the package but don't write into file and pass parameters directly

src/integration_tests/integration_tests/integration_test_node.py
```diff
@@ -0,0 +1,662 @@
...
+import rclpy.node
+import yaml
+from rclpy.impl.rcutils_logger import RcutilsLogger
+from rclpy.node import Node
+
+from integration_tests.gen_ros_dtypes import gen_ros_dtypes
+
+# Generate and import datatypes
+gen_ros_dtypes()
+ROS_DTYPES_MOD = importlib.import_module("integration_tests.ros_dtypes")
```
patrick-5546, line 23: Why do you need to use importlib for this?

.gitignore
hhenry01: Fixed.

src/integration_tests/README.md
hhenry01: I'm a bit confused about what you're getting at with this. 

If you're talking about the `generate_configs.sh` script, I tried looking into running it during a build command, but colcon doesn't support that. You can do it with CMake, but that's a bit overkill for a Python package.

If you're trying to say "why don't  we pass the custom testplan file and just 'know' that it's the custom file?", then yeah that's doable, but risks confusing the file formats. I put them in a `testplans/` folder to explicitly show that it's a different format.

src/integration_tests/integration_tests/integration_test_node.py
hhenry01: * Need to import `gen_ros_dtypes()` and run it.
* Need to import `ros_dtypes`.
* The formatter and linter *really* don't like this if we're using normal `import` statements

patrick-5546: Could try importing in the function it is used? Or commenting `# noqa: EXX` to ignore the warning

hhenry01: Fixed. The formatter made it really ugly the first time when I didn't have a comment above it.

src/integration_tests/integration_tests/integration_test_node.py
```diff
@@ -0,0 +1,618 @@
...
+            for e_ros_output in self.__test_inst.ros_expected_outputs():
+                topic = e_ros_output.name
+                self.__monitor.register(topic, e_ros_output.msg)
+                sub = self.create_subscription(
+                    msg_type=e_ros_output.msg_type,
+                    topic=topic,
+                    # Fun fact, setting this callback without functools.partial() requires
+                    # writing something atrocious like:
+                    # callback=((lambda topic: (lambda msg: self.__sub_ros_cb(msg, topic)))(topic))
+                    # Python variable scoping can be very funky
```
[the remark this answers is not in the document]
hhenry01: The naming does nothing. I just have the `__` because boat_simulator and controller seem to prefix "private" member functions with `__`.

src/integration_tests/README.md
hhenry01: Just for a common and consistent launch interface.

src/integration_tests/integration_tests/gen_ros_dtypes.py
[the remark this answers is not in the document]
hhenry01: Generating it this way means that any changes to custom_interfaces is automatically applied to the integration test tool. I don't see how importlib can do that.

src/integration_tests/config/generate_configs.sh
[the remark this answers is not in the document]
hhenry01: Done.

src/integration_tests/testplans/example.yaml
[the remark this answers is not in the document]
hhenry01: Not at the moment.

src/integration_tests/integration_tests/integration_test_node.py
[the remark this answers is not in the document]
hhenry01: In my personal opinion, splitting files for the sake of splitting files isn't worth it. There's a non-zero chance someone might want to move files around in the future and having separate files can make that more annoying.

src/integration_tests/README.md
patrick-5546: I think we can remove, adds unneeded complexity

src/integration_tests/integration_tests/integration_test_node.py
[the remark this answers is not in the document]
patrick-5546: I still think it would be better **if** it reduces the amount of code that http implementor has to read, but it's marginal so feel free to resolve this comment however you like

patrick-5546 (approved):
It wasn't importlib that was needed to simplify getting ros dtypes, but getattr

src/integration_tests/integration_tests/integration_test_node.py
```diff
@@ -0,0 +1,687 @@
...
+
+        # custom_interfaces
+        case _:
+            try:
+                return (
+                    getattr(custom_interfaces.msg, dtype)(),
+                    getattr(custom_interfaces.msg, dtype),
+                )
+            except AttributeError:
+                raise TypeError(f"INVALID TYPE: {dtype}")
```
patrick-5546, line 285: Is there a way to propagate this error? In stdout all I see is a generic message "Failed to setup tests! Integration test code is probably buggy!"

hhenry01: My exception handling was sloppy. I cleaned it up and it works now.

src/integration_tests/README.md
hhenry01: Damn, there was another reason. The global launch code is what sets the log directory. I'll have to set it inside the node.

PR #230, opened 2024-01-23
Update pre-base image

patrick-5546:
### Description

Some upstream changes have broken pre-base.Dockerfile, specifically being able to build OMPL. In this PR, I'm updating OMPL to the latest commit and manually adding the changes made to [ompl.Dockerfile](https://github.com/ompl/ompl/blob/main/scripts/docker/ompl.Dockerfile) since I last looked at it.

### Verification

- [x] Rebuild all images with these changes

### Resources

- Commit history of OMPL: https://github.com/ompl/ompl/commits/main/
- Commit history of ompl.Dockerfile: https://github.com/ompl/ompl/commits/main/scripts/docker/ompl.Dockerfile
- Building pre-base: https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/pre-base/README.md#how-to-build
- Building base, local-base, and dev: https://github.com/UBCSailbot/sailbot_workspace/blob/main/.devcontainer/base-dev/README.md#how-to-build


patrick-5546:
@SPDonaghy @stevenxu27 can you try to build the pre-base and base-dev images following the instruction in the READMEs? @SPDonaghy you don't have to build the base-dev one because you did it before. Reference: https://ubcsailbot.github.io/docs/latest/current/sailbot_workspace/docker_images/

If you have any questions comment them on this PR. Once you're successful please approve, thanks

SPDonaghy:
@patrick-5546 I am just looking at the README for the pre-base image and I don't think my computer has the resources to build the image. 

Just so I understand better, why does it need to be run on a personal computer?

patrick-5546:
> I am just looking at the README for the pre-base image and I don't think my computer has the resources to build the image.

Maybe overnight can you try building it to see if it works? I'm just curious, I dynamically limited the number of processes building OMPL uses so hopefully it doesn't crash. But just in case save and close other all other programs before building lol.

> Just so I understand better, why does it need to be run on a personal computer?

Before it was because the VM that GitHub Actions ran on was even worse than most personal computers. However that seems to have changed, it seems like the runners now have 4 processors and 16 GB of RAM. Let me try to run this in CI, I'll keep you updated.

One more thing I should have mentioned, the script should be run in the Ubuntu terminal (not in the dev container)

SPDonaghy:
Got it, okay I will try to run it overnight and give an update tomorrow.

patrick-5546:
> Let me try to run this in CI

It took 2 hours and 45 minutes to build from scratch, which isn't too bad I guess: https://github.com/UBCSailbot/sailbot_workspace/actions/runs/7689182429

However, as maintainers-to-be of sailbot workspace, I still think you should at least try going through the [updated] debugging steps: https://github.com/UBCSailbot/sailbot_workspace/tree/update-pre-base/.devcontainer/base-dev#debugging-locally

.devcontainer/base-dev/base-dev.Dockerfile
```diff
@@ -1,4 +1,177 @@
-FROM ghcr.io/ubcsailbot/sailbot_workspace/pre-base:ros_humble-ompl_4c86b2f-mongo_367-v2 as base
```
SPDonaghy, line 1: I see that you combined pre-base and base-dev images. So in the future, when members add a dependency to the base-dev image, the workflow to build the image will install ompl and ros each time right?

SPDonaghy (commented):
Just testing out the review process.

.devcontainer/base-dev/base-dev.Dockerfile
patrick-5546: Technically yes, but in reality docker caches each command in the dockerfile (creates a layer for each command). So ideally Docker will start executing commands at the first line that you change.

stevenxu27:
I ran into some type of error last night (I think something went wrong with my laptop haha) but I'm rebuilding right now.

patrick-5546:
> I think something went wrong with my laptop

Hopefully it wasn't the build crashing your laptop by using too much memory

stevenxu27:
Hey Patrick! I ran the Create New Driver and Build Command from [https://github.com/UBCSailbot/sailbot_workspace/tree/update-pre-base/.devcontainer/base-dev#debugging-locally](url) but I was a bit confused about how to proceed further. Was that all I needed to move on to building the base-dev images?

patrick-5546:
> Was that all I needed to move on to building the base-dev images?

If the build command completed successfully then you have already built all the images and you're good to go. No need to continue on and push the images, and you can approve this PR.

In the work party this week I plan to explain to y'all what each of the commands are doing, please bring any other questions you have as well. Then we can discuss what you can work on next

stevenxu27:
Sounds good I think it did build correctly. I was checking with Sean about that and it seems we have the same thing. Looking forward to learning more this weekend!

stevenxu27 (approved):

SPDonaghy:
I also had it build successfully last night. Took 2 hrs 7 mins

PR #235, opened 2024-01-27
Integrate website into tests

patrick-5546:
### Description
To integrate the website into our tests, we will be enforcing these conventions:
- By default and for integration, all repositories connect to the `sailbot_db` DB
    - No one writes to this
- Unit tests for remote transceiver and website connect to their own DB
   - Website: `TestDB`
   - Remote transceiver: `test`

image

This PR enables the website container to run by default in the dev container and in our CI.

### To Do
- [x] @hhenry01: logic to clear and populate `sailbot_db` before running unit tests
- [x] @jahn18: website reads from `sailbot_db` by default
- [x] Revert local pathfinding branch
- @jahn18: switch the website to start from a more standard image (part of deployment, doesn't have to be done in this PR)
    https://github.com/UBCSailbot/sailbot_workspace/blob/8d15221729131d962c2a75cd253e9af6a9f4cf73/.devcontainer/website/website.Dockerfile#L4 

### Verification

- [x] @SPDonaghy: https://github.com/UBCSailbot/local_pathfinding/pull/70 CI passes receiving position from website API instead of mock GET server


SPDonaghy:
> ### Verification
> * [ ]  @SPDonaghy: [Global Path HTTP Module local_pathfinding#70](https://github.com/UBCSailbot/local_pathfinding/pull/70) CI passes receiving position from website API instead of mock GET server

@patrick-5546 @jahn18 Is the idea that running the test task should also populate some data at `http://localhost:3005/api/gps` ?

I have a successful connection to the url when I run the test task, but there is no data.

patrick-5546:
> @patrick-5546 @jahn18 Is the idea that running the test task should also populate some data at http://localhost:3005/api/gps ?

Yes once the to do's in the PR description is complete. That's why I created this PR as a draft (not ready for review)

patrick-5546:
Decided to change the common database name from `db` to `sailbot_db` to avoid confusion

patrick-5546:
@jahn18 can you switch the website to read from `sailbot_db` by default? I think this would require updating the `.env.local` file right? Would this cause any problems with running your own tests?

PR #280, opened 2024-03-09
Local pathfinding PR 66

patrick-5546:
### Description

- Original PR (make sure diff matches this) https://github.com/UBCSailbot/local_pathfinding/pull/66
- Resolves local pathfinding issue #33 

- Added two functions in the parent class: (i) Creates random sample states; (ii) Compares every pair in that sample and picks the maximum motionCost()
- Inherited these functions in each Objective class to find the max_motionCost for each objective function
- Modified the motionCost() function of each Objective class to normalise its cost using max_motionCost

### Verification

- Verified using a simple setup example problem
- Haven't included unit tests

### Resources

-


akash-venkateshwaran:
Hi Patrick, I have completed the normalization. Done the testing of speed obj as well
There an issue in logic for normalization where I got 1.0 for every pair of random samples. I have fixed that bug. 
Also there is a problem with SpeedObjectiveMethod.SAILBOT_TIME where no matter how many samples I use I get diff max_motion_cost every time I run it. I don't think its a huge issue. Please check it out

patrick-5546 (commented):
Couple comments
- Local pathfinding tests now take a really long time to run (went from 6s to >1min). Did you observe the same on your machine? Any idea what may be the cause?
- Why do you think `SAILBOT_TIME` doesn't converge and why isn't it an issue?
- How much latency does setting the num_samples to 2k add for `SAILBOT_TIME`?
    - It took a lot of time for me to run this individually 10 times (4s vs. 426s)

FireBoyAJ24:
We will redo normalizing objectives as there are issues with the implementation, with adding pyompl bindings, and the code itself.

Issue #310, opened 2023-10-21
Adding Realism to Simulated Sensor Data 


chrischang5:
### Purpose

Tasks the progress of development of error generation classes or functionality. The high-level details are outlined on [confluence](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1809678337/Physics+Engine+Fluid+Simulation+Design+Specification#Artificial-Noise-Generation).

Noise (error) and delay will be added to any data that is output from sensors and published to ROS2 topics. The data that will be impacted by noise and delay is listed [here on confluence](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1785790595/Boat+Simulator+Design+Specification#ROS-Message-Interfaces).

This is representative of real sensors just having occasional spikes in readings or delays in readings when they simply don't update for a short period of time.

#### Noise Generation
- [x]  https://github.com/UBCSailbot/boat_simulator/issues/53: Design `Sensor` classes: `WindSensor` and `GPS`, add support for noise generation, and update the `Output Interface` section on [Confluence](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1809678337/Physics+Engine+Fluid+Simulation+Design+Specification#Artificial-Noise-Generation)

#### Delay Generation
- [ ] UBCSailbot/sailbot_workspace#308

Issue #308, opened 2024-01-20
Emulate Delays in Sensor Readings


DFriend01:
### Purpose

Adding artificial delays adds to sensor realism in the simulation.

### Description

The sensor classes should be updated to support artificial delays. In particular, we should be able to observe the following:

Suppose we have a publish period of 0.5 seconds, and a delay period of 0.8 seconds. On the first publish, 0.5 seconds have passed, but the sensor reading should not be updated because of "delays". On the second publish, 1.0 seconds have passed, and you should now update the sensor reading since we have exceeded the delay period. The delay period should be updated according to some policy (i.e. fixed, random, etc.)

PR #311, opened 2024-03-10
Sensors: Add delay model

DFriend01:
### Description

- Resolves UBCSailbot/sailbot_workspace#310 and UBCSailbot/sailbot_workspace#308

Add a delay modeling option, which just delays registering values in the sensor by a single update cycle. "Registering" refers to storing a value in the sensor.

    Delay model will delay sensor value updates by one update cycle:
        - Sensor has initial data x0 at t = 0
        - Sensor provided new data x_1 at t = 1
        - Sensor provided new data x_2 at t = 2. At t = 2, x1 is registered into the sensor.
        - Sensor provided new data x_i at t = i. At t = i, x_{i-1} is registered into the sensor.

### Verification

- Run tests in `src/boat_simulator/tests/unit/common/test_gps_sensor.py` and `src/boat_simulator/tests/unit/common/test_wind_sensor.py`

### Resources

- N/A


chrischang5:
Comments leftover from https://github.com/UBCSailbot/boat_simulator/pull/75

https://github.com/UBCSailbot/boat_simulator/pull/75#discussion_r1518541636
> The standard deviation will be at different units/scales for each of these properties. We should have three instead of one. We should also make it clear that the standard deviation is only used for noise generation in its variable name

Resolved via commit.

https://github.com/UBCSailbot/boat_simulator/pull/75#discussion_r1518541839
> What happens when it's zero? Similar question for the wind sensor

Behaves like the constant generator.

PR #320, opened 2024-03-13
Implement controllers for rudder and sail actuation mechanisms

### LINKED ISSUE #306: Implement controllers for rudder and sail actuation mechanisms
### Purpose

There are high-level controllers that give us high-level information on what we want to do. For example, the wingsail controller computes the trim tab angle that we want to have in order to achieve maximum thrust. However, what is the mechanism that actually turns the trim tab from position $a$ to $b$? That is the goal of the low-level control node.

In particular, this issue focuses on implementing the backbone logic of the low-level control node -- the logic that actually turns the rudder and trim tab. In reality, this is handled by ELEC in the real system, but we need a way to simulate this in our boat simulator.

#### Controller Behavior

The first thing we need to consider is what behavior we should be observing as we step through time. Recall that the simulator operates iteratively in timesteps, so the controller should be designed with this in mind. For example, suppose we want to turn the rudder from $5\degree$ to $20\degree$. A naive implementation would immediately set the rudder angle to $20\degree$ after one timestep. The timestep we use is 0.5 seconds, so that would correspond to turning the rudder 15 degrees in 0.5 seconds (that is fast!). What we really want is a gradual adjustment over time -- something like the following:

$$5\degree \to 7.5\degree \to 10\degree \to \dots \to 20\degree$$

#### Software Design
The software design should utilize OOP principles that allow the implementation to generalize to both the rudder and sail actuators. The idea is that if somebody wanted to implement their own controller, there should be little to no changes in the public interface that is exposed. A few things to consider are:
- The actuation behavior (linear? fast in the beginning but slows down towards the end? PID?)
- Input information. Do you give the setpoint directly, or do you give some other information that allows the controller to compute the setpoint internally?
- The controllers operate **iteratively**, meaning that it may take more than one timestep to reach the setpoint

#### Concrete Implementations
There are two concrete implementations that we want out of this issue:
1. The rudder actuator
2. The sail actuator

For the rudder controller, we decided to go with [Raye's implementation](https://github.com/UBCSailbot/raye-boat-controller/blob/master/python/tack_controller.py) since there were issues with the PID controller, but in the future we will likely want to use PID instead to reflect the real system. The **new rudder angle** is given as follows (note that it's in **radians**):

$$\phi_{\text{target}} = \dfrac{k_p}{1 + c_p \left|e[t]\right|} \cdot e[t]$$

$$e[t] = \text{atan2}\left(\sin\left(\text{DesiredHeading}[t] - \text{CurrentHeading}[t]\right), \cos\left(\text{DesiredHeading}[t] - \text{CurrentHeading}[t]\right)\right)$$

For your reference, Raye used $k_p = 0.7$ and $c_p = 0.34$, but feel free to adjust when necessary. You can assume that the desired heading and current heading are supplied when beginning a new actuation.

For the sail controller, you can assume that the target trim tab angle is supplied directly when beginning a new actuation (no need to compute it as the rudder controller does -- the high-level sail controller does that work for you).

Your implementation should be inside the directory `boat_simulator/nodes/low_level_control`.

stevenxu27:
### Description

- Resolves #306

- Rudder and Sail Actuation Controller Implementations. Working on unit tests still but needed to reopen PR from migration.

### Verification

- The tests in `tests/unit/nodes/low_level_control` under the `TestRudderController` test the functions of the controller implementation for the rudder and sail simulation controller. 

### Resources

- https://stackoverflow.com/a/2007279 heading error calculation
- https://github.com/UBCSailbot/raye-boat-controller/blob/master/python/tack_controller.py Raye controller mechanism


patrick-5546 (commented):
Hey Steven can you delete `.vscode/settings.json` and `compile_commands.json`? They're not used anymore

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+        return rudder_change_deg
+
+    def compute_setpoint(self) -> Scalar:
+        """Calculates the rudder setpoint angle. Only called at the very beginning once
+        to determine the setpoint angle.
+
+        Returns:
+            The rudder setpoint angle in degrees
+        """
+        return self.current_control_ang + self.compute_error_angle()
```
DFriend01, line 145: Apologies but I misinterpreted the paper, but the setpoint the just the value returned by `self.compute_error_angle`. So really, the value returned is $\phi_{\text{setpoint}}$ and not $\Delta\phi$. I realized this after seeing the following in Raye's codebase (and rereading section 5.1 of the paper):

https://github.com/UBCSailbot/raye-boat-controller/blob/3a9a363eb9d769c23a73e9a07666e1cdf28a077a/python/boat_controller_node.py#L103-L105

Thankfully this fix is simple with a few variables and function renaming with some slight changes to documentation. Sorry about this, I am going to update the issue accordingly.

If you plot out the function in desmos as well, and convert to degrees (with $k_p = c_p = 0.3$), you'll see the function saturate at about $\pm 55 \degree$, which makes more sense to support that this is the setpoint rather than the change in the angle.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+
+    def reset_desired_heading(self, target_angle: Scalar) -> None:
+        """Resets a new desired heading angle. Must recompute all errors and
+        feedback angles again for new iteration loop.
+
+        Args:
+            `target_angle` (Scalar): New desired heading
+
+        """
+        self.desired_heading = target_angle
```
DFriend01, line 155: I might put a reset function in the superclass since both actuators use the notion of "resetting" in a new control loop. I recommend that you have the reset function change the angle setpoint, but have the necessary calculations done internally (like computing the target setpoint from the current and desired headings, etc.)

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+        Returns:
+            False if target angle was not reached, else True if desired heading has been reached
+        """
+
+        if self.running_error > 0:
+            change = self.rudder_speed * self.time_step
+        elif self.running_error < 0:
+            change = -self.rudder_speed * self.time_step
+        else:
+            return True
```
DFriend01, line 170: I would be careful trying to compare this way because you assume that you can get exactly zero. But you might get numerical issues with floating point numbers, so try and compare within a tolerance rather than exact equality.

You can use `math.copysign` to get the sign of a number rather than using this `if-else` statement as well.

First see if you are close to zero, otherwise do the actuation.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+        """
+        self.desired_heading = target_angle
+
+    def update_state(self) -> bool:
+        """Updates the rudder angle iteratively towards the target rudder setpoint based
+        on the rudder_speed and time step parameters
+
+        Returns:
+            False if target angle was not reached, else True if desired heading has been reached
+        """
```
DFriend01, line 163: I skimmed this function in both classes and they look identical. Consider just implementing this in the superclass to avoid DRY code. You might have to rename some class variables though.

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
DFriend01, line 1: I recommend three more types of tests:

1. A test where you do a full actuation until it's done.
2. A test where you do one full actuation, reset the controller to a new target, and then do another full actuation to that new target.
3. Test the sail actuator as well

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,149 @@
...
+
+import numpy as np
+import pytest
+
+from boat_simulator.common.types import Scalar
+from boat_simulator.common.utils import bound_to_180
+from boat_simulator.nodes.low_level_control.controller import RudderController
+
+
+class TestRudderController:
```
DFriend01, line 13: You can use pytest fixtures to define a rudder controller before each test so you don't have to redefine a new controller in every single test.

DFriend01 (changes requested):
Great work! I made an error in the issue with respect to computing the rudder setpoint, so take a look at my comment related to that.

stevenxu27:
> Hey Steven can you delete `.vscode/settings.json` and `compile_commands.json`? They're not used anymore

For sure I'll get that deleted!

.vscode/settings.json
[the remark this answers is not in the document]
stevenxu27: Will get it deleted!

compile_commands.json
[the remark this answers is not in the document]
stevenxu27: Same thing!

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+from boat_simulator.common.types import Scalar
+from boat_simulator.common.utils import bound_to_180
+
+
+class ActuatorController(ABC):
+    """Abstract class for rudder and sail actuation mechanisms.
+
+    Attributes:
+        `current_control_ang` (Scalar): Current control mechanism angle in degrees.
+        `time_step` (Scalar): Time per iteration given in degrees/second.
```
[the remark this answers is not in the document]
stevenxu27: Yep so basically a 0.5 time_step with a rudder_speed of 2 would mean every 0.5 seconds, the rudder rotates 2 degrees

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+    def compute_error(
+        self, desired: Scalar, current: Scalar
+    ) -> Scalar:  # angle passed in as radians
+        """Computes the error between desired and current heading
+        implementation taken from: https://stackoverflow.com/a/2007279
+        Angles are bound with the convention (-180, 180]
+
+        Args:
+            `desired` (Scalar): New desired heading in degrees
+            `current` (Scalar): New current heading in degrees
```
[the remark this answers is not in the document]
stevenxu27: I will need to change this convention. I believe I'm bounding (-180, 180], so the zero is currently at the positive x. I'll alter my parameters to take in this convention

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+        Calculations done using Raye's implementation
+            https://github.com/UBCSailbot/raye-boat-controller/blob/master/python/tack_controller.py
+
+        Returns:
+            Scalar: Difference between current and target rudder angle in degrees.
+        """
+        error = self.compute_error(self.current_heading, self.desired_heading)
+
+        if abs(error) > (pi / 4):  # limited between (-45, 45)
+            raise ValueError("Heading_error must be between -pi/4 and pi/4")
```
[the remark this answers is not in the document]
stevenxu27: Sounds good I'm thinking about just embedding constants into my class itself. I wasn't sure if users might want to pass in custom angle ranges in the future, but I can definitely take out that hard code error check.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+            `current_heading` (Scalar): Current boat heading direction in degrees.
+            `desired_heading` (Scalar): Target boating heading direction in degrees.
+            `current_control_ang` (Scalar): Current control mechanism angle in degrees.
+            `time_step` (Scalar): Time per iteration given in degrees/second.
+            `kp` (Scalar): Proportional constant when calculating error.
+            `cp` (Scalar): Timeseries of error values computed over time.
+            `max_angle_range` (tuple): Tuple of minimum to maximum control angle range in degrees
+            `rudder_speed` (Scalar): Speed of controller change in degrees per second
+
+        """
```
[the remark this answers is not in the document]
stevenxu27: Sounds good I think I have that change already. I actually wasn't too sure; what is cp exactly?

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @abstractmethod
+    def update_state(self) -> bool:
+        """Updates the controller position based on target heading angles
+
+        Returns:
+            False if target angle was not reached, else True if desired heading has been reached
+        """
+
+        pass
```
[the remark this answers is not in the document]
stevenxu27: Yes, I will clarify that better in my docstring

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+
+    @abstractmethod
+    def compute_error_angle(self) -> Scalar:
+        """Computes the corresponding feedback based on error and desired heading"""
+        pass
```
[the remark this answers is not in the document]
stevenxu27: Will change!

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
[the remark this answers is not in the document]
stevenxu27: For this can I just specify when the user inputs an angle to use this convention? This way when they pass in the degree I can still bound it accordingly, or am I simplifying it too much?

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Returns:
+            Scalar: Difference between current and target rudder angle in degrees.
+        """
+        error = self.compute_error(self.current_heading, self.desired_heading)
+
+        if abs(error) > (pi / 4):  # limited between (-45, 45)
+            raise ValueError("Heading_error must be between -pi/4 and pi/4")
+
+        rudder_change = (self.kp * error) / (1 + (self.cp * error))
```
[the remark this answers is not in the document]
stevenxu27: Yes!

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
stevenxu27: Sounds good I will just take the outpoint of one of my function as the new setpoint! Will take a look at the implementation to confirm.

stevenxu27: This function would no longer exist for the sail actuator since we pass in a target angle right?

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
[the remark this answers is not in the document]
DFriend01: The desired heading and current heading will both be sensor readings that you'll get directly from the ROS network, so they are already in that convention. It's a matter of convenience for you to adopt this convention in your actuator to avoid switching conventions.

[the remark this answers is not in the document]
DFriend01: So yes, you want to tell the use what convention to use. Euler angles are unfortunately arbitrary sometimes, so we are forced to specify angle conventions all the time to avoid confusion.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
stevenxu27: I've created a reset function for each and added it into the superclass.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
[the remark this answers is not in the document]
stevenxu27: Sounds good. If my understanding is correct, there wasn't anything in my code that really depended on the convention; it would just be important for the user to understand what convention is being used when understanding how to utilize my controller right? I made changes to specify the angle convention in the doc string.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,258 @@
...
+        """Resets a new desired heading angle, therefore recalculating the corresponding
+        setpoint and running error.
+
+        Args:
+            `new_desired_heading` (Scalar): New desired heading in degrees
+
+        """
+
+        self.desired_heading = new_desired_heading
+        self.compute_setpoint()
```
DFriend01, line 180: What about current heading? We need both desired heading and current heading to determine the new rudder setpoint.  This value is given externally since the boat simulator should determine this value and pass it on to you.

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,185 @@
...
+    # Test compute_setpoint method of RudderController
+    setpoint = rudder_controller.compute_setpoint()
+    heading_error = rudder_controller.compute_error()  # heading_error in radians
+
+    rudder_setpoint = (rudder_controller.kp * heading_error) / (
+        1 + (rudder_controller.cp * abs(heading_error))
+    )
+
+    if abs(rudder_setpoint) > RUDDER_MAX_ANGLE_RANGE[1]:
+        raise ValueError("rudder_change must be between -pi/4 and pi/4")
```
DFriend01, line 60: Exception?

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,185 @@
...
+            rudder_controller.update_state()
+            if rudder_controller.running_error == 0:
+                progress = True
+                break
+            if counter > 1000:
+                break
+    assert progress
+
+
+def test_update_state_reset(rudder_controller):
```
DFriend01, line 101: You should also test what the rudder angle should be while it actuates in addition to the running error

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,185 @@
...
+    np.equal(calculated, expected)
+
+
+# testing for one iteration only
+def test_update_state(rudder_controller):
+    rudder_controller.compute_setpoint()
+    assert rudder_controller.update_state() == (abs(rudder_controller.running_error) == 0)
+
+
+def test_update_state_continuous(rudder_controller):
```
DFriend01, line 83: Check the rudder angle too

DFriend01 (changes requested):
One thing to keep in mind for these classes is that they will be used over and over again across control iterations. This means:

1. We compute a set point
2. Actuate to that set point
3. Upon a new actuation, reset and recompute the set point
4. Repeat the above process

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+
+    def __init__(
+        self,
+        current_control_ang: Scalar,
+        time_step: Scalar,
+        control_speed: Scalar,
+        running_error: Scalar,
+        max_angle_range: Tuple[Scalar, Scalar],
+    ):
```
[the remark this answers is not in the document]
stevenxu27: I needed the running error to be a part of the parent class to be passed into the update_state function, but it definitely didn't need to be specified by the user. I've set it to default to 0.0 when declaring the initializing the object in the constructor.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,258 @@
...
+
+        Returns:
+            False if target control angle was not reached, else
+            True if target control angle has been reached (running error = 0)
+        """
+        error_tracker = False
+        if np.isclose(self.running_error, 0, 0.01):
+            return True
+        else:
+            change = copysign(1, self.running_error) * self.control_speed * self.time_step
```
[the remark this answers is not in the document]
stevenxu27: Definitely easier!

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,258 @@
...
+            next_control = max(self.current_control_ang + change, self.max_angle_range[0])
+            self.running_error -= change
+        else:
+            next_control = max(
+                self.current_control_ang + self.running_error, self.max_angle_range[0]
+            )
+            self.running_error = 0
+            error_tracker = True
+
+        self.current_control_ang = min(next_control, self.max_angle_range[1])
```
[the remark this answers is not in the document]
stevenxu27: Basically, the first if statement checks if the magnitude of the running error is greater than the magnitude of change. This determines whether a full "change" iteration can be completed within the next time step, or if the difference is less than what is required for one iteration. For example, if the running_error was 10, and our control speed per time step was 2, we know that we can change by a full iteration of 2 degrees. In contrast, if the running _error were 1.5, we would know to not actuate by 2 degrees to avoid overshooting the target control angle. 

In the loop itself, next_control stores the larger value between the current control angle + change or the index 1 of the RUDDER_MAX_ANGLE constant, which in this case is -45 degrees, bounding it above -45 degrees. I'm currently calculating running_error by target - current, meaning if the error is negative, the target is larger than the current and the change is negative. If the error is positive, that means the target angle is less than the current control, meaning our change should be positive. If the running error is negative and its change is negative, subtracting the change from the running error will result in running_error converging towards zero. In the else loop, we run the same next_control check, except we add whatever is remaining in the running error, which will be less than the change. Our running_error should be set to 0, and the error tracker tracks that we reached our target control angle, returning True. The last check bounds the angle value to be less than 45 degrees.

Hopefully that made some sense haha.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        max_angle_range: Tuple[Scalar, Scalar],
+    ):
+        """Initializes the class attributes. Note that this class cannot be directly instantiated.
+
+        Args:
+            `current_control_ang` (Scalar): Current control mechanism angle in degrees.
+            `time_step` (Scalar): Time per iteration given in seconds.
+            `control_speed` (Scalar): Speed of control angle change in degrees / second.
+            'running_error' (Scalar): Error between current and target control angle. Default to 0
+        """
```
DFriend01, line 37: You probably want to remove running error and replace with max angle range

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+
+        Returns:
+            False if target control angle was not reached, else
+            True if target control angle has been reached (running error = 0)
+        """
+        error_tracker = False
+        if np.isclose(self.running_error, 0, 0.01):
+            return True
+        else:
+            change = copysign(self.control_speed * self.time_step, self.running_error)
```
DFriend01, line 60: A few suggestions:

1. Rename `error_tracker` to something more descriptive to its purpose, like `is_target_reached`
2. Have the `if` statement be the first thing you do, however, don't include an `else` branch and just compute the change outside the `if` statement. If the `if` statement evaluates to true, you just return anyways, so there is no need for an else branch. Most other languages would complain that you defined `change` in the else branch but use it outside because it could "technically be not defined at all".

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+    def __init__(
+        self,
+        current_heading: Scalar,
+        desired_heading: Scalar,
+        current_control_ang: Scalar,
+        time_step: Scalar,
+        kp: Scalar,
+        cp: Scalar,
+        control_speed: Scalar,
+        max_angle_range=None,
```
DFriend01, line 91: Set the default to `RUDDER_MAX_ANGLE_RANGE` rather than `None`.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+
+        super().__init__(
+            current_control_ang,
+            time_step,
+            control_speed,
+            max_angle_range,
+        )
+
+        self.max_angle_range = RUDDER_MAX_ANGLE_RANGE
```
DFriend01, line 116: If you address the comment above, this line is no longer necessary because the super class constructor already does the assignment for you!

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        self.running_error = 0.0
+        self.current_heading = bound_to_180(current_heading)  # bound (-180, 180] in degrees
+        self.desired_heading = bound_to_180(desired_heading)  # bound (-180, 180] in degrees
+        self.kp = kp
+        self.cp = cp
+        self.setpoint = 0.0  # current setpoint angle in degrees
+        self.running_error = 0.0
```
DFriend01, line 123: Running error is defined twice. Also, it's already defined in the super class and set to zero, so this isn't necessary unless you plan on changing it to something other than zero in this subclass.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def compute_error(self) -> Scalar:  # angle passed in as radians
```
DFriend01, line 125: There are no arguments, so the comment can be removed I believe

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+            Scalar: The error between the given headings in radians, 0 radians at positive
+            y axis and increasing clockwise
+        """
+        desired_rad, current_rad = np.deg2rad(self.desired_heading), np.deg2rad(
+            self.current_heading
+        )
+
+        error = atan2(sin(desired_rad - current_rad), cos(desired_rad - current_rad))
+
+        return error  # in radians
```
DFriend01, line 140: Rather than putting this comment here, you could just change the variable name to `error_rad`. Recall that you also specify that the return value is in radians.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+    def compute_setpoint(self) -> Scalar:
+        """Computes the corresponding control error angle between current control angle and
+        target control angle. Uses Raye's implementation from:
+            https://github.com/UBCSailbot/raye-boat-controller/blob/master/python/tack_controller.py
+
+        Returns:
+            Scalar: Corresponding error angle between the
+            current and target control angle in degrees"""
+
+        heading_error = self.compute_error()  # heading_error in radians
```
DFriend01, line 151: Could change variable to `heading_error_rad` rather than having this comment here

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        self.control_speed = control_speed
+        self.max_angle_range = max_angle_range
+        self.running_error = 0.0
+
+    def update_state(self) -> bool:
+        """Updates the controller position iteratively based on control_speed * time_step, until
+        the control target angle is reached.
+
+        Args:
+            `max_angle_range` (tuple): Max angle range individual to each different controller
```
DFriend01, line 49: There are no arguments

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        rudder_setpoint = min(
+            max(rudder_setpoint, np.deg2rad(self.max_angle_range[0])),
+            np.deg2rad(self.max_angle_range[1]),
+        )
+
+        rudder_setpoint_deg = np.rad2deg(rudder_setpoint)  # in degrees
+        self.running_error = rudder_setpoint_deg - self.current_control_ang  # in degrees
+        self.setpoint = rudder_setpoint_deg  # in degrees
+
+        return rudder_setpoint_deg  # in degrees
```
DFriend01, line 164: Make use of descriptive variable naming to remove comments

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def reset_setpoint(self, new_desired_heading: Scalar, new_current_heading: Scalar) -> None:
+        """Resets a new desired heading angle, therefore recalculating the corresponding
+        setpoint and running error.
+
+        Args:
+            `new_desired_heading` (Scalar): New desired heading in degrees
+
```
DFriend01, line 172: Missing the second argument

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+    def __init__(
+        self,
+        target_angle: Scalar,
+        current_control_ang: Scalar,
+        time_step: Scalar,
+        kp: Scalar,
+        cp: Scalar,
+        control_speed: Scalar,
+        max_angle_range=SAIL_MAX_ANGLE_RANGE,
+    ):
```
DFriend01, line 204: `kp` and `cp` are not used in this class, only the rudder actuator

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        """Initializes the class attributes.
+
+        Args:
+            `target_angle` (Scalar): Target angle for sail controller in degrees.
+             0 degrees (North) at the positive y axis and increasing clockwise.
```
DFriend01, line 209: In particular, the target angle is for the **trim tab** on the sail, not the sail itself.

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        super().__init__(
+            current_control_ang,
+            time_step,
+            control_speed,
+            max_angle_range,
+        )
+
+        self.max_angle_range = SAIL_MAX_ANGLE_RANGE
+        self.running_error = 0.0
```
DFriend01, line 226: Does running error need to be defined here?

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+
+        """
+        super().__init__(
+            current_control_ang,
+            time_step,
+            control_speed,
+            max_angle_range,
+        )
+
+        self.max_angle_range = SAIL_MAX_ANGLE_RANGE
```
DFriend01, line 225: Angle range is already defined in the super class

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        self.running_error = 0.0
+        self.target_angle = target_angle
+        self.kp = kp
+        self.cp = cp
```
DFriend01, line 229: Not used

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        self.running_error = self.target_angle - self.current_control_ang
+        return self.running_error
+
+    def reset_setpoint(self, new_target: Scalar) -> None:
+        """Resets a new desired sail actuator angle and updates the running_error
+
+        Args:
+            `target_angle` (Scalar): New desired sail controller angle
+
+        """
```
DFriend01, line 248: In particular, this is for the trim tab angle

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,259 @@
...
+        self.compute_error()
+
+    def reset_target_angle(self, changed_target_angle) -> None:
+        """Changes target_angle desired to a new angle. Used for testing purposes
+
+        Args:
+            `changed_target_angle` (Scalar): New desired heading in degrees
+
+        """
+        self.target_angle = changed_target_angle
```
DFriend01, line 259: What is the difference between this function and the one above? What is this needed for?

DFriend01 (changes requested):

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
DFriend01, line 1: - Call reset in the constructor
- Only indicate `update_state` and `reset` to be public and used by the user. Everything else is the internal workings of the implementation

DFriend01 (changes requested):

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -0,0 +1,238 @@
...
+            np.deg2rad(self.max_angle_range[1]),
+        )
+
+        rudder_setpoint_deg = np.rad2deg(rudder_setpoint_rad)
+        self.running_error = rudder_setpoint_deg - self.current_control_ang
+        self.setpoint = rudder_setpoint_deg
+
+        return rudder_setpoint_deg
+
+    def reset_setpoint(self, new_desired_heading: Scalar, new_current_heading: Scalar) -> None:
```
DFriend01, line 162: By convention, public functions should go before private ones

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,212 @@
...
+    expected = rudder_controller.reset_setpoint(new_desired_heading, new_current_heading)
+    rudder_controller._change_desired_heading(new_desired_heading)
+    calculated = rudder_controller._compute_setpoint()
+    np.equal(calculated, expected)
+
+
+# testing for one iteration only
+def test_update_state(rudder_controller):
+    current_control = rudder_controller.current_control_ang
+    rudder_controller._compute_setpoint()
```
DFriend01, line 85: Since you reset on initialization, do you need this line anymore?

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,212 @@
...
+    assert rudder_controller.update_state() == (abs(rudder_controller.running_error) == 0)
```
DFriend01, line 86: Wouldn't `assert rudder_controller.update_state()` be good enough here? I'm also a little confused on why running error needs to be zero here. Couldn't it be non-zero since we now call reset upon initialization? A single iteration shouldn't bring the running error to zero in general, right?

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,212 @@
...
+                    rudder_controller.control_speed * rudder_controller.time_step,
+                    rudder_controller.running_error,
+                )
+            ),
+            rudder_controller.current_control_ang,
+        )
+
+
+def test_update_state_continuous(rudder_controller):
+    rudder_controller._compute_setpoint()
```
DFriend01, line 101: Needed?

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,212 @@
...
+        while not np.isclose(rudder_controller.running_error, 0, 0.1):
+            counter += 1
+            rudder_controller.update_state()
+            if rudder_controller.running_error == 0:
+                progress = True
+                break
+            if counter > 1000:
+                break
+    assert progress
+    assert np.equal(rudder_controller.current_control_ang, rudder_controller.setpoint)
```
DFriend01, line 116: Check if they are close rather than for exact equality. In general, try not to compare for exact equality in floats cause you might get numerical issues.

DFriend01 (changes requested):
Just a few comments mostly related to your tests. Almost there!

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
stevenxu27: Basically, update.state() will return false if we aren't at the target yet, so assert rudder_controller.update.state() will usually be false since running_error != 0. My check there is basically checking if update_state is false, that means the magnitude of running_error is greater than 0. If update_state is true, then running_error = 0. That was my way of checking the two possible cases of a single iteration, either moving towards our target angle or reaching our target angle.

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
```diff
@@ -0,0 +1,212 @@
...
+    rudder_controller._compute_setpoint()
+    progress = False
+    counter = 0
+    if np.isclose(rudder_controller.running_error, 0, 0.1):
+        progress = True
+    else:
+        while not np.isclose(rudder_controller.running_error, 0, 0.1):
+            counter += 1
+            rudder_controller.update_state()
+            if rudder_controller.running_error == 0:
```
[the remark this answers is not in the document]
stevenxu27: That's really cool! I didn't realise that's what properties meant.

src/boat_simulator/tests/unit/nodes/low_level_control/test_controller.py
DFriend01: Oh I see, thanks for explaining

DFriend01 (approved):

PR #331, opened 2024-03-19
Update links to deprecated sub-repositories

patrick-5546:
### Description

There were a couple places that still referenced the old repositories. This PR either updates the url or deletes them

### DIFF
--- a/.devcontainer/docs/README.md
+++ b/.devcontainer/docs/README.md
@@ -1,6 +1,6 @@
 # Docs Image
 
-Used for running [our docs site](https://github.com/UBCSailbot/docs).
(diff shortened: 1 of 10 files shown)


jamenkaye (commented):
Looks good to me! I double checked the new links work. I'm guessing the links to the repository specific test folders were taken out because the tests were moved, since they're redundant after the monorepo migration right?

patrick-5546:
> 'm guessing the links to the repository specific test folders were taken out because the tests were moved, since they're redundant after the monorepo migration right?

They linked to the "Tests" GitHub actions for each repository. Since we don't use those repositories anymore there's no point in keeping their badges in their READMEs. The "Tests" badge for sailbot workspace can be found in the main README, and runs the tests for all packages

PR #332, opened 2024-03-19
Remove legacy-peer-deps header when installing node packages for website

jahn18:
### Description
A weird side-effect occurs in the `package-lock.json` file when we run `npm install` using the header `--legacy-peer-deps`. The header is no longer needed, so I'm removing it from the Dockerfile for the website.  

Also, I'm setting next.js to a fixed version. It's not good to install the "latest" version on every build - we want control over exactly which packages update and how that update happens.

### Verification

- Rebuilt the containers, the website still functions fine.


patrick-5546 (approved):
Rebuilt the container and verified all tests passed and package-lock.json was not modified. Thanks!

Issue #269, opened 2023-11-24
Transmit Global Path Flow


hhenry01:
### Purpose

Get global path waypoints from the remote server to the Iridium satellite network.

### Description

Make sure to save POSTed global path to the db. 
1. Add a `/global-path` target in the remote transceiver HTTP server.
2. Work with Global Pathfinding to define and parse the POST format.
3. Write POSTed path to the database.
4. Test by POSTing and querying the database.

### Resources

- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1814954153/Remote+Transciever#Transmit-Global-Path-Sequence
  - *diagram does not show writing global path to database. 

How to parse json:
```
// You need to add two includes to the top of the file
#include <boost/property_tree/json_parser.hpp>
#include <boost/property_tree/ptree.hpp>

// after checking for application/json
std::string json_str = beast::buffers_to_string(req_.body().data());    
std::stringstream           ss(json_str);
boost::property_tree::ptree json_tree;
boost::property_tree::read_json(ss, json_tree);
std::string timestamp = json_tree.get<std::string>("timestamp");
std::cout << "timestamp: " << timestamp << std::endl;
for (const auto & waypoint : json_tree.get_child("waypoints")) {
    float lat = waypoint.second.get<float>("latitude");
    float lon = waypoint.second.get<float>("longitude");
    std::cout << "waypoint latlon: " << lat << " " << lon << std::endl;
}
```

PR #344, opened 2024-03-23
Transmit Global Path Flow # 269

vaibhavambastha:
Resolved transmit global path flow #269, getting global path waypoints from the remote server to the Iridium satellite network. Tested by querying the database and verifying the appropriate global path/timestamps. - Vaibhav/Hassan


samdai01:
still need to finish testing and sort out merge from main

PR #365, opened 2024-04-25
Changed Port for Posting Global Paths in Tests

### LINKED ISSUE #363: Unit tests start 2 http servers on the same port
### Description

A recent NET PR causes the test task to fail: both NET and PATH unit tests start http servers on port 8081

#### Expected Behavior

Test task should pass

#### Actual Behavior

On my machine the local pathfinding unit tests run after net, and I get this error:
image

### Steps to Reproduce

1. Run the test task

### Relevant Code

- NET: https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/network_systems/scripts/http_echo_server.py
- PATH: https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/test/post_server.py
    - Temporary substitute for NET global path endpoint

### Potential Solutions

@hhenry01 any ideas? Is the NET global path endpoint working yet? Or should one of the servers run on a different port

SPDonaghy:
### Description

- Resolves #363 

Moved the test post_server from port 8081 to 8085.
Also refactored the test files a bit so that we only need to change one line to change the test port again if needed.

### Verification

- Ran the test task, but the local path tests might be running before NET on my machine

### DIFF
--- a/src/local_pathfinding/local_pathfinding/global_path.py
+++ b/src/local_pathfinding/local_pathfinding/global_path.py
@@ -140,11 +140,12 @@ def get_path(file_path: str) -> Path:
     return path
 
 
-def post_path(path: Path) -> bool:
+def post_path(path: Path, url: str = PATH_URL) -> bool:
     """Sends the global path to NET via POST request.
 
     Args:
         path (Path): The global path.
+        url (str): URL to post path to. Should always use default value except for tests.
 
     Returns:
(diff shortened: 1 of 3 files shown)


patrick-5546 (approved):
Thanks. Verified by running test task locally in main branch (failed) then in this branch (passed)

PR #388, opened 2024-05-15
Make new leads codeowners

patrick-5546:
Merge in after all other PR's by graduating members are merged. Then I will make a release


patrick-5546:
@UBCSailbot/soft-admin FYI

PR #390, opened 2024-05-25
Add website/tests/node_modules and settings.json to .gitignore

SPDonaghy:
### Description

Added src/website/tests/node_modules and settings.json to the .gitignore file to ensure these are not tracked by git.

### Verification

CI passes

### Resources

Slack Reference:
https://ubcsailbot.slack.com/archives/C042XBRFBKJ/p1716044689558669

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,7 @@
 
 # microsoft c/c++ intellisense (disabled in favor of clangd)
 /.vscode/c_cpp_properties.json
+/.vscode/settings.json
 
 # python
 __pycache__/
@@ -32,3 +33,6 @@ venv/
 .cache/
 # Build
 site/
+
+# web
+/src/website/tests/node_modules/*
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,7 @@
 .cache/
 
 # microsoft c/c++ intellisense (disabled in favor of clangd)
-/.vscode/c_cpp_properties.json
-/.vscode/settings.json
+/.vscode/*
 
 # python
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -12,6 +12,7 @@
 
 # microsoft c/c++ intellisense (disabled in favor of clangd)
 /.vscode/c_cpp_properties.json
+/.vscode/settings.json
```
patrick-5546, line 15: Is settings.json still being autogenerated? What is its contents? Haven't had this happen to me before

SPDonaghy: Yeah I just did a fresh rebuild of everything and it showed up for me. I can check the contents when I get home, I didn't look too closely at what was in there

patrick-5546: it doesn't for me, maybe due to a difference in our personal settings

SPDonaghy: That's probably what it is, do you think settings.json should be tracked?

patrick-5546: maybe just ignore the entire directory since we don't use it. All settings should be in the code-workspace file

SPDonaghy: Yeah thats a good point, I'll do that

.gitignore
```diff
@@ -32,3 +33,6 @@ venv/
 .cache/
 # Build
 site/
+
+# web
+/src/website/tests/node_modules/*
```
JordanChen123, line 38: Looks good to me!

JordanChen123 (approved):

PR #405, opened 2024-06-13
Updated CI/CD Tests

### LINKED ISSUE #387: [CD] Setup docker on remote VM
### Purpose

Modify our CD workflow to ensure the remote server we're using has `docker` installed. 

### Description

We're using `docker-compose` to start up all our containers in the remote VM. 

See the PR below, you'll want to an additional step in `.github/workflows/deployment.yml`   

### Resources

- https://github.com/UBCSailbot/sailbot_workspace/pull/366

JordanChen123:
### Description

- Resolves #387 
- Added checks to the workflow, including Lint Checks and Build Test

### Verification

- [ ] 

### Resources

-

### DIFF
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -38,3 +38,4 @@ jobs:
             touch src/website/.env.local # Adds a local environment file to VM to override env vars in .env.production  
             echo NEXT_PUBLIC_SERVER_HOST=http://${{ secrets.SSH_HOST }} >> src/website/.env.local
             docker compose -f .devcontainer/docker-compose.yml -f .devcontainer/website/docker-compose.website.prod.yml up --force-recreate -d --build --remove-orphans
+            
--- a/.github/workflows/deployment.yml
+++ b/.github/workflows/deployment.yml
@@ -16,6 +16,8 @@ jobs:
     steps:
       - name: Checkout workspace
         uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.GH_TOKEN }}
       - name: Install dependencies and run lint checks
         run: |
           cd sailbot_workspace/src/website
@@ -38,4 +40,3 @@ jobs:
             touch src/website/.env.local # Adds a local environment file to VM to override env vars in .env.production  
             echo NEXT_PUBLIC_SERVER_HOST=http://${{ secrets.SSH_HOST }} >> src/website/.env.local
             docker compose -f .devcontainer/docker-compose.yml -f .devcontainer/website/docker-compose.website.prod.yml up --force-recreate -d --build --remove-orphans
-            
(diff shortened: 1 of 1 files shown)


.github/workflows/deployment.yml
```diff
@@ -1,28 +1,49 @@
...
-  # Uncomment once we are ready to deploy 
-  #push:
-  #  branches:
-  #    - main
   workflow_dispatch:
+    # Uncomment once we are ready to deploy 
+    #push:
+    #  branches:
+    #    - main
+  pull_request:
```
SPDonaghy, line 5: What would be the motivation for running this on every pull request rather than on pushes to main? 
Wouldn't this just be testing the configuration of the main branch and not the changes being introduced?
If we want to keep it as being triggered by pull requests can we only trigger it when files in src/website and .devcontainer are modified?

.github/workflows/deployment.yml
```diff
@@ -1,28 +1,49 @@
...
+          host: ${{ secrets.SSH_HOST }}
+          username: ${{ secrets.SSH_USERNAME }}
+          key: ${{ secrets.SSH_KEY }}
+          script: |
+            [ ! -d "sailbot_workspace" ] && git clone https://github.com/UBCSailbot/sailbot_workspace.git
+            cd sailbot_workspace
+            git pull
+            touch src/website/.env.local # Adds a local environment file to VM to override env vars in .env.production  
+            echo NEXT_PUBLIC_SERVER_HOST=http://${{ secrets.SSH_HOST }} >> src/website/.env.local
+            docker compose -f .devcontainer/docker-compose.yml -f .devcontainer/website/docker-compose.website.prod.yml up --force-recreate -d --build --remove-orphans
```
SPDonaghy, line 45: I don't see a verification that docker is installed on the VM prior to deploying to it. Which is fine, just wanted to clarify as you mentioned it resolved that issue

.github/workflows/deployment.yml
JordanChen123: From my understanding, running this on every PR would ensure that changes in the PR would not break the deployment before they are merged into the main branch, which would also provide another layer of testing to catch issues early. What do you think?

Also, good point. If we do want to keep it as being triggered by PR's, then it should only trigger when files in src/website and .devcontainer are modified. I will fix that

SPDonaghy: Yeah sounds like a great idea to test out how changes deploy before actually deploying to the real site.  

~~The main branch will be checked out on the VM, and not the development branch~~
I'm sorry that's totally wrong ^

Would the VM that this workflow deploys to be like a pre-production environment and not the actual web host?

If so, then I think keeping it as triggered by PRs makes sense, and maybe we can keep the push trigger as well and deploy to either the pre-production VM or the actual website host VM depending on the trigger event.

What do you think?

JordanChen123: Your idea sounds great, since the VM that the workflow deploys will not be the actual web host.

PR #413, opened 2024-06-29
Updating CAN Frames and Adding Tests

lross03:
### Description

- Updates some old CAN Frames and adds new ones. Tests implemented for each.


samdai01:
commits and changes messed up, replacing with new PR

PR #415, opened 2024-07-04
Boat state update with new wind and current vectors

### LINKED ISSUE #401: Calculate physics engine data
### Purpose

This purpose of this issue is to update the physics engine node to combine the fluid generation class with the boat state class in a timer callback to calculate real values for the physics engine.

### Description

Create a method in the `PhysicsEngineNode` class that will:
1. Generate new fluid vectors for wind and water.
2. Update the boat's state using these fluid vectors.

Then call this method in `__publish()` before any of the other methods are called.

Take a look at the [FluidGenerator](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/physics_engine/fluid_generation.py) class for simulating wind and current velocities and the [BoatState](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py) class for tracking and updating the boat's state based on the fluid vectors.

### Ressources

- [UBC Sailbot Python Conventions and Style Guide](https://ubcsailbot.github.io/sailbot_workspace/main/reference/python/conventions/)

alberto-escobar:
### Description

- Resolves #401 

- Generate new fluid vectors for wind and water by calling `next` method on respective fluid generator objects
- Pass new fluid vectors into `__boat_state` through `step` method

### Verification

- No testing done, this will be tested through integration testing


src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
```diff
@@ -278,6 +278,7 @@ def __init_timer_callbacks(self):
 
     # PUBLISHER CALLBACKS
     def __publish(self):
+        self.__update_boat_state()
         """Synchronously publishes data to all publishers at once."""
```
eomielan, line 282: Could you please move the method call so that it appears just below the docstring

eomielan (changes requested):
Looks good! Just one small comment.

src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
alberto-escobar: I have changed this in the new commit.

eomielan (approved):

PR #417, opened 2024-07-13
Remade "Updating CAN Frames and Adding Tests" PR

lross03:
### Description

- [Description Updates some old CAN Frames and adds new ones. Tests implemented for each.](https://github.com/UBCSailbot/sailbot_workspace/pull/413)


src/network_systems/projects/can_transceiver/inc/can_frame_parser.h
```diff
@@ -29,33 +30,36 @@ namespace msg    = custom_interfaces::msg;
...
-    SAIL_WSM_DATA_FRAME_1  = 0x63,
-    SAIL_WIND_DATA_FRAME_1 = 0x65,
-    PATH_GPS_DATA_FRAME_1  = 0x80,
-    PATH_WIND_DATA_FRAME   = 0x84,
-    GENERIC_SENSOR_START   = 0x100,
-    GENERIC_SENSOR_END     = 0x1FF
+    PWR_MODE             = 0x00,
+    MAIN_HEADING         = 0x01,
+    MAIN_TR_TAB          = 0x02,
+    RESERVED             = 0x29,
```
samdai01, line 36: what is the purpose of this reserved `0x29` can frame ID?

samdai01 (changes requested):
Please review the CAN Frames document, and ensure that all of the CAN Frame representations correctly match them. Each bit needs to match in this case or else the frame will be parsed improperly. 

Also, as a tie in above, ELEC will not be sending any floats over to us, so ensure that there are no float representations at all in CAN Transceiver.

src/network_systems/projects/can_transceiver/inc/can_frame_parser.h
```diff
@@ -140,13 +144,11 @@ class BaseFrame
...
 public:
-    // Valid CanIds that a Battery object can have
-    static constexpr std::array<CanId, 2> BATTERY_IDS       = {CanId::BMS_P_DATA_FRAME_1, CanId::BMS_P_DATA_FRAME_2};
-    static constexpr uint8_t              CAN_BYTE_DLEN_    = 8;
-    static constexpr uint8_t              BYTE_OFF_VOLT     = 0;
-    static constexpr uint8_t              BYTE_OFF_CURR     = 2;
-    static constexpr uint8_t              BYTE_OFF_MAX_VOLT = 4;
-    static constexpr uint8_t              BYTE_OFF_MIN_VOLT = 6;
+    // Valid CanIds that a Battery object can have. There used to be two, but now there is only one.
+    static constexpr std::array<CanId, 1> BATTERY_IDS    = {CanId::BMS_DATA_FRAME};
```
samdai01, line 148: Since we only have one type of CAN frame for the battery, do we still need to use the `BATTERY_IDS` array of one element? there should be no need to verify the CAN ID of the message now. 

Please remove the `BATTERY_IDS` array, update `publishBattery` in `can_transciever_ros_intf.cpp` to not go through the for loop,  and update the publisher to publish `HelperBattery.msg`. Keep the original `Batteries.msg` for now.

samdai01 (changes requested):
Moved the BATTERY_IDS comment here for context.

src/network_systems/projects/can_transceiver/inc/can_frame_parser.h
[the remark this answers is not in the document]
lross03: Reserved is used in testing to test a known invalid frame ID.

samdai01 (approved):

PR #418, opened 2024-07-20
added voyage descriptions from file

fyang151:
### Description

- Resolves #_issue_

### Verification

- [ ] 

### Resources

image


src/website/views/components/Polaris/PolarisContainer.tsx
```diff
@@ -22,88 +22,92 @@ function stringToParagraphs(text: string) {
   ));
 }
```
JordanChen123, line 23: Could you remove this function, as it is not needed anymore

src/website/views/components/Polaris/PolarisContainer.tsx
```diff
@@ -22,88 +22,92 @@ function stringToParagraphs(text: string) {
...
+      action: downloadBatteriesData,
+    },
+    {
+      title: 'Wind Sensors',
+      data: ['Wind Sensors', '12 hours', 'JSON', 'Download'],
+      action: downloadWindSensorsData,
+    },
+    {
+      title: 'Generic Sensors',
+      data: ['Generic Sensors', '12 hours', 'JSON', 'Download'],
```
JordanChen123, line 70: Please remove 'Download' from the data array for Wind Sensors and Generic Sensors.

src/website/views/components/Polaris/PolarisContainer.tsx
```diff
@@ -22,88 +22,92 @@ function stringToParagraphs(text: string) {
...
-        title: 'Generic Sensors',
-        data: ['Generic Sensors', '12 hours', 'JSON', 'Download'],
-        action: downloadGenericSensorsData,
-      },
-    ];
+  useEffect(() => {
+    fetch('/info/PolarisVoyageDescription.txt')
+      .then((res) => res.text())
+      .then((data) => setDescription(data));
+  }, []);
```
JordanChen123, line 33: Good job! The code is a lot cleaner now with the large text moved to a separate file.

I noticed that the text isn't formatted into paragraphs anymore, as we cannot use \n in the text file anymore. 

One way we could address this is by adding HTML paragraph tags. As an example, in the text file, you could have: 
`<p>Raye is an 18 foot...<p>`
`<p>Raye’s hull was designed off...<p>`
`<p>Our electrical system is...<p>`
`<p>Raye’s software can be separated...<p>`

The HTML `<p>` tags are to define paragraphs, which allows the content to be rendered correctly into paragraphs when used in an html context. 

Now In `PolarisContainer.tsx`, you can render the HTML content stored in `description` to allow the `<p>` tags in the text file to be rendered correctly as html paragraphs. 

A React property you could use is `dangerouslySetInnerHTML`. It can allow you to set HTML content directly by rendering the HTML content from the text file. Using this property can be risky in some scenarios where there can be user-generated input, but since we control the content of the text file, it can work without being dangerous.

src/website/views/components/Polaris/PolarisContainer.tsx
JordanChen123, line 1: Since it is not needed anymore, please remove the `stringToParagraphs` function.

JordanChen123 (changes requested):

PR #421, opened 2024-07-28
Fix view and edit source links in docs site

patrick-5546:
### Description
- The edit URI incorrect so it didn't open the right link; this PR should fix that

image


SPDonaghy (approved):
Looking good 😎

PR #429, opened 2024-08-16
Updated physics engine to use wind sensor

### LINKED ISSUE #426: Update physics engine node to use WindSensor
### Purpose
The purpose of this issue is to update the [physics engine node](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py) to use the `SimWindSensor` class.

### Description
1. Create a private instance of the [SimWindSensor class](https://github.com/UBCSailbot/sailbot_workspace/blob/3bf448c71e83f4cc4cbba4d6527c837414a02b9c/src/boat_simulator/boat_simulator/common/sensors.py#L83).
2. Create a new private method: `__update_sim_wind_sensor()` and update the wind attribute of the `SimWindSensor` instance you created in step 1 using the wind generator (`self.__wind_generator`).
3. Call this method right before `self.__update_boat_state()` in the `__publish()` function.
4. Update the `self.__update_boat_state()` method to use the wind sensor data rather than the wind generator so that we introduce noise.

### Resources

- [FluidGenerator class](https://github.com/UBCSailbot/sailbot_workspace/blob/3bf448c71e83f4cc4cbba4d6527c837414a02b9c/src/boat_simulator/boat_simulator/nodes/physics_engine/fluid_generation.py#L11)

arynrosh:
- Created a private instance of the `WindSensor` class in the physics engine node.
- Added a new private method: `__update_wind_sensor()`. This method updates the `wind` attribute of the `WindSensor` instance using the wind generator (`self.__wind_generator`).
- Integrated `__update_wind_sensor()` call right before `self.__update_boat_state()` in the `__publish()` function.
- Modified the `self.__update_boat_state()` method to use data from the `WindSensor` instead of the wind generator, introducing realistic noise into the simulation.
- Resolves #426

### DIFF
--- a/src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
+++ b/src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
@@ -110,6 +110,7 @@ class PhysicsEngineNode(Node):
         self.__rudder_angle = 0
         self.__sail_trim_tab_angle = 0
         self.__desired_heading = None
+        self._wind_sensor = WindSensor()
         self.__boat_state = BoatState(
             0.5, Constants.BOAT_PROPERTIES.mass, Constants.BOAT_PROPERTIES.inertia
         )
@@ -119,6 +120,12 @@ class PhysicsEngineNode(Node):
         self.__current_generator = FluidGenerator(
             generator=MVGaussianGenerator(np.array([1, 1]), np.array([[2, 1], [1, 2]]))
         )
+
+    def __update_wind_sensor(self):
+        """Updates the wind sensor with the latest wind data from the wind generator."""
+        wind_data = self.__wind_generator.next()
+        self.__wind_sensor.speed.speed = wind_data[0]
+        self.__wind_sensor.direction = wind_data[1]
 
     def __declare_ros_parameters(self):
         """Declares ROS parameters from the global configuration file that will be used in this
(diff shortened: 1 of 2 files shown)


src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
```diff
@@ -120,6 +121,12 @@ def __init_private_attributes(self):
             generator=MVGaussianGenerator(np.array([1, 1]), np.array([[2, 1], [1, 2]]))
         )
 
+    def __update_wind_sensor(self):
+        """Updates the wind sensor with the latest wind data from the wind generator."""
+        wind_data = self.__wind_generator.next()
+        self.__wind_sensor.speed.speed = wind_data[0]
```
chrischang5, line 127: I just happened to stumble upon this. You shouldn't have to assign the private members in this way. The update method is there to help you out, and you get additional functionality with noise/delays if you want. Try `self.update(wind=wind_data[0])`?

See the [wind sensor tests](https://github.com/UBCSailbot/sailbot_workspace/blob/058ebe4e9769b6e82230b492e7720e2e067e642a/src/boat_simulator/tests/unit/common/test_wind_sensor.py#L47) for an example of this type of call.

src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
```diff
@@ -120,6 +121,12 @@ def __init_private_attributes(self):
...
+        self.__wind_sensor.direction = wind_data[1]
```
chrischang5, line 128: `direction` doesn't look like a field in the `WindSensor` class, you'll have to add it by following the pattern of other fields in these classes (ex: `wind` field)

eomielan:
Issue with be addressed in a separate PR

PR #434, opened 2024-09-20
Land Data Generator Script

SPDonaghy:
### Description

This PR just includes a very short script to convert a GEOJSON file containing territorial water boundaries for CA and US to a shapley Multipolygon that we can query for state point validity.

I discarded the old script. 

The output file is `land.pkl` 

### Verification

All verification is visual in the jupyter notebook. I recommend pulling this branch and checking it out!


FireBoyAJ24 (approved):
Looks good. We will in the future have to make polygons of the land around the on water testing region. In addition, we can add obstacle courses to test how the sailbot deals with difficult situations.

PR #436, opened 2024-09-26
Implement Land Obstacles - 278

### LINKED ISSUE #278: Implement Land Obstacle Class
### Purpose

We need to implement a class for land obstacles. 

### Description

A Land class will be added to the `obstacles.py`. After implementation, we can decide if the Land and Boat classes are similar enough that  it would make sense to combine them.

### Resources

* [Raye Local Pathfinding](https://github.com/UBCSailbot/raye-local-pathfinding)
* [Raye Land Mass Creator](https://github.com/UBCSailbot/raye-local-pathfinding/blob/master/land_masses/land_mass_creator.py)
#### Land Mass Data ETL Pipeline
This is how we'll generate our data set of land polygons to avoid. 
Land Mass ETL (1)

#### Rough Idea for Land Mass Data Loading
WIP

SPDonaghy:
### Description

- Resolves #278 

### Verification

- Unit tests and visual testing in the same jupyter notebook


SPDonaghy:
some dependency hell going on, need to pin shapely and geopandas somewhere. Docker image will probably be the move.

SPDonaghy:
Okay, yeah this is done

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -101,7 +100,132 @@ def update_reference_point(self, reference: HelperLatLon):
...
+
+            if collision_zone.geom_type == "Polygon":
+                collision_zone = MultiPolygon([collision_zone])
+        else:
+            collision_zone = MultiPolygon()
+
+        self.collision_zone = collision_zone
+
+    @staticmethod
+    def _latlon_polygons_to_xy_polygons(
```
FireBoyAJ24, line 189: Could change the name to `_latlon_multipolygon_to_xy_polygons` as it's similar to the `_latlon_polygon_to_xy_polygon`.

src/local_pathfinding/land/generate_land_data.py
```diff
@@ -0,0 +1,70 @@
...
+    with open(file_path, "wb") as f:
+        pickle.dump(object, f, protocol=pickle.HIGHEST_PROTOCOL)
+
+
+def load_pkl(file_path: str) -> Any:
+    with open(file_path, "rb") as f:
+        return pickle.load(f)
+
+
+def pickle_land():
```
FireBoyAJ24, line 24: Should add a docstring for this function

src/local_pathfinding/land/generate_land_data.py
```diff
@@ -0,0 +1,70 @@
...
+    # the only useful data is in the id and geometry columns and in the first two rows
+    gdf = gdf[["id", "geometry"]]
+    gdf = gdf.head(2)
+    geoms = list(gdf.iloc[0]["geometry"].geoms) + list(gdf.iloc[1]["geometry"].geoms)
+
+    # convert to MultiPolygon for easier geometric manipulation
+    multi_poly = MultiPolygon(geoms)
+    multi_poly = multi_poly.buffer(0)  # repairs invalid geometry
+    # this is to prevent geometry the crosses -180 longitude from being distorted by transforms
+    with open(INLAND_POLYGON, "r") as f:
```
FireBoyAJ24, line 46: Could add error handling here

FireBoyAJ24 (changes requested):

FireBoyAJ24:
Will review the test cases later this week

SPDonaghy:
Im gonna take another look at the function using map and see if I can simplify it

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -101,7 +100,132 @@ def update_reference_point(self, reference: HelperLatLon):
...
+class Land(Obstacle):
+    """
+    Describes land and territorial waters that which Sailbot must avoid. During runtime,
+    the obstacle tracker will keep track of a single Land obstacle object and update its collision
+    zone with new geometry when required.
+
+    Attributes:
+        collision_zone (MultiPolygon): A collection of Polygons in (X,Y) that define regions of
+                                       land stored in the Land object.
+        next_waypoint (HelperLatLon): Lat/lon position of the next global waypoint.
```
jamenkaye, line 119: Doesn't the next global waypoint already exist as `reference` in the parent class?

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -101,7 +100,132 @@ def update_reference_point(self, reference: HelperLatLon):
...
+                )
+            )
+
+        if len(polygons) == 0:
+            return []
+
+        elif len(polygons) == 1:
+            return [_latlon_polygon_to_xy_polygon(polygons[0])]
+
+        return list(map(_latlon_polygon_to_xy_polygon, polygons))
```
jamenkaye, line 228: What's the rationale for the different returns? I'm concerned the calling code will need to be more complicated to handle these cases

jamenkaye (commented):
Looks good! I don't think any of this is blocking but I'm interested in discussing this for my own curiosity :)

src/local_pathfinding/local_pathfinding/obstacles.py
SPDonaghy: I named it `_latlon_polygon_list_to_xy_polygon_list` as the function does not expect a MultiPolygon parameter, rather just a list of polygons. What do you think?

src/local_pathfinding/land/generate_land_data.py
SPDonaghy: done

src/local_pathfinding/land/generate_land_data.py
SPDonaghy: Sure! which errors do you think we should plan for here?

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -1,46 +1,46 @@
...
 
+    Do not instantiate an Obstacle object directly, use one of the subclasses instead.
+
     Attributes:
         reference (HelperLatLon): Lat and lon position of the next global waypoint.
-        sailbot_position (XY): Lat and lon position of SailBot.
-        sailbot_speed (float): Speed of the SailBot in kmph.
-        collision_zone (Optional[Polygon]): Shapely polygon representing the
+        sailbot_position (HelperLatLon): Lat and lon position of SailBot.
+        collision_zone (Optional[Polygon or MultiPolygon]): Shapely geometry representing the
```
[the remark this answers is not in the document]
SPDonaghy: You're right there's no motivation to do that haha. I'll remove the Optional tag. I think back when I put that I thought it meant it was an option between the two types. but no, the collision zone will never be none once the sub class constructor has returned.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -49,48 +49,47 @@ def is_valid(self, point: XY) -> bool:
...
         if self.collision_zone is None:
-            raise ValueError("Collision zone has not been initialized")
-
-        # contains() requires a shapely Point object as an argument
-        point = Point(*point)
+            raise RuntimeError("Collision zone has not been initialized")
 
-        return not self.collision_zone.contains(point)
+        # could also use
+        # Point(point).within(self.collision_zone)
```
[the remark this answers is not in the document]
SPDonaghy: Sure will do

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -49,48 +49,47 @@ def is_valid(self, point: XY) -> bool:
...
+                state_space=kwargs.get("state_space"),
+                land_multi_polygon=kwargs.get("land_multi_polygon"),
+            )
 
-    def update_sailbot_data(self, sailbot_position: HelperLatLon, sailbot_speed: float):
-        """Updates the sailbot's position and speed.
+    def update_sailbot_data(
+        self, sailbot_position: HelperLatLon, sailbot_speed: Optional[float] = None
+    ) -> None:
+        """Updates Sailbot's position, and Sailbot's speed (if the caller is a Boat object).
```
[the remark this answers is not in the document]
SPDonaghy: Boat objects need Sailbot's speed to calculate the size of the collision_zone for the moving ship. Land objects don't need this information for any reason.

src/local_pathfinding/local_pathfinding/obstacles.py
SPDonaghy: YES! good point, I'll remove that next_waypoint parameter and use reference in its place.

src/local_pathfinding/local_pathfinding/obstacles.py
SPDonaghy: The rationale is that based on the proximity to land and the shape of the coastline and state-space, we could be dealing with multiple polygons, a single polygon, or none so the function needs to consider those three possible cases. Only `update_land_collision_zone` will call this function. I am thinking of moving this function to `coord_ssytems.py` thoughts?

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -49,48 +49,47 @@ def is_valid(self, point: XY) -> bool:
...
-        collision_zone = affine_transform(collision_zone, transformation)
+        Updates the collision zone of the obstacle to reflect updated attributes.
+        If attributes of the obstacle have not been changed, this function will have no effect.
+        """
+        if isinstance(self, Boat):
+            # Boat Obstacle
+            self._update_boat_collision_zone()
 
-        self.collision_zone = collision_zone.buffer(COLLISION_ZONE_SAFETY_BUFFER, join_style=2)
+        else:
```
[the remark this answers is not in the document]
SPDonaghy: Yeah the decision of when to update will be up to the obstacle tracker. You update it directly using the `update_collision_zone` function or by updating the reference point. In practice, for the land obstacle, we should only be updating its collision zone through the `update_reference_point` function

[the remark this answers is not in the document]
SPDonaghy: Clarification btw, at any time there will only be one land obstacle. A MultiPolygon containing 0 to n polygons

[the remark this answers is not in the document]
SPDonaghy: Yeah the decision of when to update will be up to the obstacle tracker. You update it directly using the update_collision_zone function or by updating the reference point. In practice, for the land obstacle, we should only be updating its collision zone through the update_reference_point function


Clarification btw, at any time there will only be one land obstacle. A MultiPolygon containing 0 to n polygons

src/local_pathfinding/local_pathfinding/obstacles.py
FireBoyAJ24: If the alternative to having nested functions makes it more unreadable, then it's fine to keep it as it is.

src/local_pathfinding/land/generate_land_data.py
FireBoyAJ24: The file cannot open exception error

FireBoyAJ24: You can add file open exception errors to files that needs to open during the deployment of the boat

FireBoyAJ24 (approved):

PR #445, opened 2024-10-26
Preliminary Resolution for #425 update low level control node

### LINKED ISSUE #425: Update low level control node with implemented controllers
### Purpose
The purpose of this issue is to update the low level control node (`boat_simulator/nodes/low_level_control/low_level_control_node.py`) to use the implemented rudder and sail actuation controllers.

### Changes
Update `__rudder_actuation_routine()` to use the implemented rudder controller and `__sail_actuation_routine()` to use the implemented sail controller.

### Resources

- [Controllers Implementation](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py)
- [Low Level Control Node](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py)

alberto-escobar:
### Description

- Preliminary implementation that aims to resolve #425
- This is still under works, I am publishing so code can be review by others while I still work on verification.

### Verification

- I will work on making more formal tests. The big hurdle is that testing of implementation relies on other packages and some are crashing during running resulting in my code never being reached. The other big hurdle is ROS2 is very complicated and I need to take some time understanding what I am doing.

### Resources

- N/A


alberto-escobar:
note to self: mock data node implement, can now test out my implementation here using the following command
`ros2 launch boat_simulator main_launch.py enable-mock-data:=true`

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -202,19 +219,36 @@ def __rudder_actuation_routine(
...
+        # set controller to disabled
+        if self._parameters["rudder.disable_actuation"].value:
+            # TODO change this block of code with correct logic of what to do when rudder is off.
+            self.get_logger().warn("Rudder actuation disabled.")
+
+        else:
+            self.get_logger().warn("Rudder actuation enabled.")
+            current_heading = self.gps.heading.heading
+            desired_heading = goal_handle.request.desired_heading.heading.heading
+            self.__rudder_controller.reset_setpoint(desired_heading, current_heading)
```
alberto-escobar, line 237: For some reason, when I call this method, the `RudderController` does not update its attribute `running_error` correctly so the wile loop below does not trigger.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -202,19 +219,36 @@ def __rudder_actuation_routine(
...
+            if self.__rudder_controller.is_target_reached:
+                self.get_logger().error(
+                    f"Somehow controller target is reached when controller desired heading is {self.__rudder_controller.desired_heading} and controller current heading is {self.__rudder_controller.current_heading}"
+                )
+            feedback_msg = SimRudderActuation.Feedback()
+            while not self.__rudder_controller.is_target_reached:
+                self.__rudder_controller.update_state()
```
alberto-escobar, line 244: For some reason, when I call this method, the `RudderController` does not update its attribute `running_error` correctly so the wile loop below does not trigger.

alberto-escobar: I tested this in an isolated environment and it seems to work. I created a script called `test.py` and it has the following:
```
from controller import RudderController

controller = RudderController(0.0, 0.0, 0, 1, 1, 0.1, 0.5)
print(controller.is_target_reached)
controller.reset_setpoint(-120.0, 0.0)
print(controller.is_target_reached)
while not controller.is_target_reached:
    controller.update_state()
```
and it prints the right values:
```
True
False
```

alberto-escobar:
image
this is my test, the warn statements were used to help with visability but have changed to debug statements now. Whenever new mock gps and sail command data is published, the actuation routines run and the controllers calculate new values for the trim tab and rudder.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -232,16 +282,37 @@ def __sail_actuation_routine(
...
+
+            feedback_msg = SimSailTrimTabActuation.Feedback()
+            self._is_sail_action_active = True
+            while not self.__sail_controller.is_target_reached:
+                self.__sail_controller.update_state()
+                i = self.__sail_controller.current_control_ang
+                feedback_msg.current_angular_position = float(i)
+                goal_handle.publish_feedback(feedback=feedback_msg)
+                self.__sail_trim_tab_angle = i
+                # self.sail_action_feedback_rate.sleep()
```
alberto-escobar, line 312: the Rate.sleep method is buggy, it causes permanent blocking.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -202,19 +233,38 @@ def __rudder_actuation_routine(
...
+            )
+            feedback_msg = SimRudderActuation.Feedback()
+            self._is_rudder_action_active = True
+            while not self.__rudder_controller.is_target_reached:
+                self.__rudder_controller.update_state()
+                i = self.__rudder_controller.current_control_ang
+                feedback_msg.rudder_angle = float(i)
+                goal_handle.publish_feedback(feedback=feedback_msg)
+                self.__rudder_angle = i
+                # self.rudder_action_feedback_rate.sleep()
```
alberto-escobar, line 263: the Rate.sleep method is buggy, it causes permanent blocking.

alberto-escobar:
Actuation routines are still work as before after last few commits I added.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -129,15 +135,11 @@ def __init_feedback_execution_rates(self):
         """
         self.get_logger().debug("Initializing rate objects...")
         self.__rudder_action_feedback_rate = self.create_rate(
-            frequency=self.get_parameter("rudder.actuation_execution_period_sec")
-            .get_parameter_value()
-            .double_value,
+            frequency=1.0 / self._parameters["rudder.actuation_execution_period_sec"].value,
             clock=self.get_clock(),
```
eomielan, line 139: Please use `self.get_parameter("rudder.actuation_execution_period_sec").get_parameter_value().double_value to avoid type errors.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -129,15 +135,11 @@ def __init_feedback_execution_rates(self):
...
         )
         self.__sail_action_feedback_rate = self.create_rate(
-            frequency=self.get_parameter("wingsail.actuation_execution_period_sec")
-            .get_parameter_value()
-            .double_value,
+            frequency=1.0 / self._parameters["wingsail.actuation_execution_period_sec"].value,
             clock=self.get_clock(),
```
eomielan, line 143: Same here

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -175,6 +177,27 @@ def __init_action_servers(self):
...
+        control_speed = 1 / time_step  # not sure if this is the right value to use
+        self.__rudder_controller = RudderController(
+            current_heading, desired_heading, current_control_ang, time_step, kp, cp, control_speed
+        )
+
+    def init_sail_controller(self):
+        target_angle = self._parameters["wingsail.fixed_angle_deg"].value
+        current_control_ang = self._parameters["wingsail.fixed_angle_deg"].value
+        time_step = self._parameters["wingsail.actuation_execution_period_sec"].value
+        control_speed = self._parameters["wingsail.actuation_speed_deg_per_sec"].value
```
eomielan, line 196: Please switch to `self.get_parameter(...)...` here as well

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -175,6 +177,27 @@ def __init_action_servers(self):
             callback_group=self.sail_action_callback_group,
         )
 
+    def init_rudder_controller(self):
+        current_heading = self.gps.heading.heading
+        desired_heading = self.gps.heading.heading
+        current_control_ang = self.rudder_angle
+        time_step = self._parameters["rudder.actuation_execution_period_sec"].value
+        kp = self._parameters["rudder.pid.kp"].value
+        cp = self._parameters["rudder.pid.kp"].value  # not sure if this is the right value to use
```
eomielan, line 186: Please add a new parameter to the `global_launch/config/globals.yaml` and update the README in the same directory with details such as range of values.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -202,19 +248,37 @@ def __rudder_actuation_routine(
...
-        feedback_msg = SimRudderActuation.Feedback()
-        for i in range(Constants.RUDDER_ACTUATION_NUM_LOOP_EXECUTIONS):
-            feedback_msg.rudder_angle = float(i)
-            self.get_logger().debug(f"Rudder Action Server feedback: {i}")
-            self.get_logger().debug(f"Is Rudder Action Active? {self.is_rudder_action_active}")
-            goal_handle.publish_feedback(feedback=feedback_msg)
-            self.rudder_action_feedback_rate.sleep()
-
+        if not self.__rudder_controller:
+            self.init_rudder_controller()
```
eomielan, line 252: Instead of this could you please call `self.__init_rudder_controller()` in `__init__()`. Also please add the `__` to the front to make it private.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -232,16 +296,35 @@ def __sail_actuation_routine(
...
-        # TODO Placeholder loop. Replace with sail ctrl once implemented.
-        feedback_msg = SimSailTrimTabActuation.Feedback()
-        for i in range(Constants.SAIL_ACTUATION_NUM_LOOP_EXECUTIONS):
-            feedback_msg.current_angular_position = float(i)
-            self.get_logger().debug(f"Sail Action Server feedback: {i}")
-            self.get_logger().debug(f"Is Sail Action Active? {self.is_sail_action_active}")
-            goal_handle.publish_feedback(feedback=feedback_msg)
-            self.sail_action_feedback_rate.sleep()
+        if not self.__sail_controller:
+            self.init_sail_controller()
```
eomielan, line 300: Please do the same thing here

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -202,19 +248,37 @@ def __rudder_actuation_routine(
...
+            while not self.__rudder_controller.is_target_reached:
+                self.__rudder_controller.update_state()
+                i = self.__rudder_controller.current_control_ang
+                feedback_msg.rudder_angle = float(i)
+                goal_handle.publish_feedback(feedback=feedback_msg)
+                self.__rudder_angle = i
+                self.rudder_action_feedback_rate.sleep()
+        self._is_rudder_action_active = False
+        self.get_logger().info(f"New rudder angle: {self.rudder_angle}")
+        self.get_logger().info("Rudder actuation complete.")
```
eomielan, line 280: Could you please move these to the else to avoid the situation where we get this print message ordering:
```
Rudder actuation disabled.
New rudder angle: {self.rudder_angle}
Rudder actuation complete.
```

which might be confusing.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -232,16 +296,35 @@ def __sail_actuation_routine(
...
+                self.__sail_controller.update_state()
+                i = self.__sail_controller.current_control_ang
+                feedback_msg.current_angular_position = float(i)
+                goal_handle.publish_feedback(feedback=feedback_msg)
+                self.__sail_trim_tab_angle = i
+                self.sail_action_feedback_rate.sleep()
+
+        self._is_sail_action_active = False
+        self.get_logger().info(f"New trim tab angle {self.__sail_controller.current_control_ang}")
+        self.get_logger().info("Trim tab actuation complete.")
```
eomielan, line 327: Same here

eomielan (commented):
Looks good, just want to double check `control_speed` calculation and then I'll approve

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -175,6 +186,44 @@ def __init_action_servers(self):
...
+        desired_heading = 0.0
+        current_control_ang = self.rudder_angle
+        time_step = (
+            self.get_parameter("rudder.actuation_execution_period_sec")
+            .get_parameter_value()
+            .double_value
+        )
+        kp = self.get_parameter("rudder.pid.kp").get_parameter_value().double_value
+        cp = self.get_parameter("rudder.pid.cp").get_parameter_value().double_value
+        control_speed = 1 / time_step  # not sure if this is the right value to use
```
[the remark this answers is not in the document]
SPDonaghy: curious why this one is not a parameter but the wingsail control speed is.

stevenxu27 (dismissed):

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
```diff
@@ -175,6 +186,44 @@ def __init_action_servers(self):
             callback_group=self.sail_action_callback_group,
         )
 
+    def __init_rudder_controller(self):
+        # initial heading value needed here
```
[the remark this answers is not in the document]
alberto-escobar: this can be deleted. The heading starts at zero but when GPS messages begin flowing, the current heading gets updated with the heading value on the gps.

src/boat_simulator/boat_simulator/nodes/low_level_control/low_level_control_node.py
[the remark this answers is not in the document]
alberto-escobar: I think its because the rudder controller config parameters in the yaml was defined to use time step long ago and then later on  control speed was used when adding config parameters for the sail controler. I can address this in a later issue.

alberto-escobar:
> Hi Alberto, just added a few questions but no real feedback. Can you please explain how you've tested these changes if they aren't suited to unit testing. Thanks!

sure, the way I tested this is by adding lines where it prints information about the node to the log. The information I print is if the action is executed or if actuation is occurring. then I ran the whole code base on ros2 and observed the output. You can see it here:



> image This is my test, the warn statements were used to help with visibility but have changed to debug statements now.
> 
> Whenever new `desired_heading` or `sail_cmd` is published, the actuation routines run and the controllers calculate new values for the trim tab angle and rudder angle.

SPDonaghy (approved):

stevenxu27 (approved):

PR #448, opened 2024-10-27
Global Path Bug

FireBoyAJ24:
### Description

Local pathfinding would break when initializing the reference_latlon in `OMPLPath` as the global path was not initialized as None in the `LocalPathState`.  To fix that I just initialized global path to none and checked it when initializing the reference latlon.

### Verification

-  Run the command `ros2 launch local_pathfinding main_launch.py` and ensured no errors was occuring.


jamenkaye (approved):
Seems good to me! Make sure the integration tests all pass before merging :)

PR #456, opened 2024-11-16
Fix dead link in setup instructions

eomielan:
### Description
- Dead link in setup instructions: https://docs.docker.com/desktop/faqs/linuxfaqs/#what-is-the-difference-between-docker-desktop-for-linux-and-docker-engine
Screenshot 2024-11-16 at 1 01 29 PM

- Replaced with https://www.docker.com/blog/how-to-check-docker-version

### Verification

- [x] Link check ci job


jamenkaye (approved):
Nice catch, thanks for fixing!

PR #461, opened 2024-11-23
Condition to Switch Path - Akash

### LINKED ISSUE #287: Conditions to switch the local path
### Purpose

This feature checks a set of conditions and compares them to determine whether to update to a new local path. It should also add logic to keep track of which global waypoint that local paths are trying to reach.

### Description

The feature allows the sailbot to update its path for various reasons, including choosing better, safer, or avoiding invalid paths.

### Resources

The list of conditions that are going to be implemented is consolidated here: [Conditions to switch](https://ubcsailbot.atlassian.net/wiki/x/_4Cnc)

akash-venkateshwaran:
### Description

- Resolves #287 
- Completed all the conditions in LocalPath class in local_path.py
- Completed all the tests for the functions in test_local_path.py except for function 'old_path_in_collision_zone()'

### Verification

- [ ] 

### Resources

-


jamenkaye (commented):
Also need to add logic to update the global waypoint when we reach it I think.

FireBoyAJ24 (changes requested):

PR #475, opened 2025-01-18
Remove AIS Ships from Local Transceiver

### LINKED ISSUE #460: Delete AISShips from Local transceiver sensors
We will not be sending AIS ships anymore from the local transceiver, thus we will need to remove including them in the `sensors` that deliver to the rockblock

Jng468:
### Description

- Resolves #460


samdai01:
There is still a call to `updateSensor(ais)` in the `sendData` test that is causing the build to fail, please remove that

samdai01 (approved):
lgtm

PR #478, opened 2025-01-18
Run virtual iridium when starting local transceiver in DEV mode

lross03:
### Description

- Runs virtual iridium script when local transceiver is started in DEV mode

### DIFF
--- a/src/network_systems/projects/local_transceiver/src/local_transceiver_ros_intf.cpp
+++ b/src/network_systems/projects/local_transceiver/src/local_transceiver_ros_intf.cpp
@@ -42,7 +42,22 @@ public:
             if (mode == SYSTEM_MODE::PROD) {
                 //TODO(Jng468) placeholder
             } else if (mode == SYSTEM_MODE::DEV) {
-                default_port = LOCAL_TRANSCEIVER_TEST_PORT;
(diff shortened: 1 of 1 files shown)


src/network_systems/projects/local_transceiver/src/local_transceiver_ros_intf.cpp
```diff
@@ -42,7 +42,22 @@ class LocalTransceiverIntf : public NetNode
             if (mode == SYSTEM_MODE::PROD) {
                 //TODO(Jng468) placeholder
             } else if (mode == SYSTEM_MODE::DEV) {
-                default_port = LOCAL_TRANSCEIVER_TEST_PORT;
+                default_port                = LOCAL_TRANSCEIVER_TEST_PORT;
+                std::string run_iridium_cmd = "./../../scripts/run_virtual_iridium.sh";
```
samdai01, line 46: Please refrain from using relative paths. We have an environment variable called `$ROS_WORKSPACE` that is linked to the `sailbot_workspace` root directory.

samdai01 (changes requested):
See comment on relative paths.

samdai01 (approved):

PR #484, opened 2025-02-03
all post requests working, need to make changes to generic sensors th…

JordanChen123:
…ough

### Description

Changes allows the website to handle incoming POST requests. 
- Still need to fix handling for generic sensors

### Verification

- [ ] 

### Resources

-


FireBoyAJ24 (approved):

JordanChen123:
Will not continue with this MR, since there are security risks the way it is right now, and we don't need post request handling anymore (for now)

If we do end up needing post request handling in the future, will need to add:

- Security Middleware
- Update endpoints with security

PR #486, opened 2025-02-06
publish kinematics data on physics engine node

### LINKED ISSUE #400: Publish physics engine data
### Purpose

The purpose of this issue is to update the physics engine node to publish real data obtained from the physics engine to the ROS network.

### Description

Update `__publish_kinematics()` to publish real data instead of hardcoded values to the ROS network. You can assume that the boat state (`self.__boat_state`) has been updated elsewhere. 
Please update the following parts of the ROS message to publish real data:
- msg.global_pose
- msg.wind_velocity
- msg.current_velocity

### Ressources

- [BoatState class](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py)

alberto-escobar:
### Description
- Resolves #400 
- Implemented publishing of physics engine data in the physics engine node

### Verification
image
- inserted logger warning statements to verify `msg` was getting values filled correctly and then ran `boat_simulator` package.  Warning statements deleted after verifying things are working.
- position and orientation not changing and always 0.0, I believe since the boat state updates is occurring elsewhere and the boat is not set to sail yet, they position and orientation does not change. **I would like confirmation of this to make sure my implementation here is complete.**

### Resources
N/A


src/boat_simulator/boat_simulator/nodes/physics_engine/physics_engine_node.py
```diff
@@ -387,20 +387,21 @@ def __publish_kinematics(self):
...
-        msg.global_pose.orientation.x = 0.0
-        msg.global_pose.orientation.y = 0.0
-        msg.global_pose.orientation.z = 0.0
+        # had to make elements into float or else I would get issues
+        msg.global_pose.position.x = float(self.__boat_state.global_position[0])
+        msg.global_pose.position.y = float(self.__boat_state.global_position[1])
+        msg.global_pose.position.z = float(self.__boat_state.global_position[2])
+        msg.global_pose.orientation.x = float(self.__boat_state.angular_position[0])
+        msg.global_pose.orientation.y = float(self.__boat_state.angular_position[1])
+        msg.global_pose.orientation.z = float(self.__boat_state.angular_position[2])
```
eomielan, line 396: Rather than casting and using `[]` to index, could you please use the numpy [item()](https://numpy.org/doc/2.2/reference/generated/numpy.ndarray.item.html) function. This should allow you to access an element and return it in the correct type.

alberto-escobar: changed this in my latest commit

eomielan (approved):
LGTM

PR #491, opened 2025-02-16
Raghumanimehta/485 removing ompl path state constructor

### LINKED ISSUE #485: Removing the redundant constructor for OMPLPathState
### Purpose

Getting rid of the redundant constructor to reduce technical debt. 

### Changes

No changes in functionality. 

### Resources

-

raghumanimehta:
### Description

- Resolves #485 

OMPLPathState class has been removed. Some of its fields have been moved as local variables to `_init_simple_setup` in `ompl_path.py` since only this method used those fields. The rest of the fields could be derived from `local_path_state` and have not been changed. Logic to calculate `state_domain` and `state_range` in `ompl_path.py` has been added. 

### Verification

- Tested the system before and after refactoring.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/ompl_path.py
+++ b/src/local_pathfinding/local_pathfinding/ompl_path.py
@@ -14,6 +14,9 @@ import pyompl
 from custom_interfaces.msg import HelperLatLon
 from rclpy.impl.rcutils_logger import RcutilsLogger
 
+# Fix the import
+from shapely.geometry import MultiPolygon, Point, Polygon, box
+
 import local_pathfinding.coord_systems as cs
 from local_pathfinding.objectives import get_sailing_objective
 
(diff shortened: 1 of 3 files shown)


src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -114,6 +98,13 @@ def get_waypoints(self) -> List[HelperLatLon]:
 
         return waypoints
 
+    def create_space(self, position) -> Polygon:
```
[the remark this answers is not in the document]
SPDonaghy: nit: maybe we could make the function name more descriptive also? Something that tells you right away that the function returns a box around the given point.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -122,21 +113,32 @@ def update_objectives(self):
...
-        """Initialize and configure the OMPL SimpleSetup object.
+        self.state = local_path_state
 
-        Returns:
-            og.SimpleSetup: Encapsulates the various objects necessary to solve a geometric or
-                control query in OMPL.
-        """
-        self.state = OMPLPathState(local_path_state, self._logger)
+        # Create buffered spaces and extract their centers
+        state_domain = self.create_space(self.state.position)
```
SPDonaghy, line 119: I think here we might be confusing what domain is (OMPL has its own different definition). You might rename this variable to something more clear that communicates its only a box around `position`.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -122,21 +113,32 @@ def update_objectives(self):
...
+        start_x, start_y = self.state.position  # Use original position for coordinates
```
SPDonaghy, line 120: `self.state.position` will be a `HelperLatLon` we have to transform it to XY coordinates before we can pass it to OMPL

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -122,21 +113,32 @@ def update_objectives(self):
...
+
+        if not self.state.global_path:
+            goal_polygon = self.create_space([0, 0])
+            goal_x, goal_y = (0, 0)
+        else:
+            goal_position = self.state.global_path[-1]
```
SPDonaghy, line 126: `self.state.global_path[-1]` will be a `HelperLatLon`, so same thing, we need to convert it to XY coordinates before passing it to OMPL. 

As a shortcut though, the XY position of the goal will always be (0,0) because the goal (in LatLon coords) is also our reference point which we use to convert `self.state.position` to XY coordinates.

SPDonaghy: For an example of how to convert coordinates from LatLon to XY see `src/local_pathfinding/test/test_coord_systems.py`

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -122,21 +113,32 @@ def update_objectives(self):
...
+            goal_x, goal_y = goal_position
 
         # create an SE2 state space: rotation and translation in a plane
         space = pyompl.SE2StateSpace()
 
         # set the bounds of the state space
         bounds = pyompl.RealVectorBounds(dim=2)
-        x_min, x_max = self.state.state_domain
-        y_min, y_max = self.state.state_range
+        big_polygon = box(*MultiPolygon([state_domain, goal_polygon]).bounds)
```
SPDonaghy, line 135: We can call this polygon the state space, as it fully defines the boundary of the 2D (XY) space in which we want our path planner to operate.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -156,8 +158,8 @@ def _init_simple_setup(self, local_path_state) -> pyompl.SimpleSetup:
         # set the goal and start states of the simple setup object
         start = pyompl.ScopedState(space)
         goal = pyompl.ScopedState(space)
-        start_x, start_y = self.state.start_state
-        goal_x, goal_y = self.state.goal_state
+        # start_x, start_y = start_state
```
SPDonaghy, line 161: can just remove these lines

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -170,6 +172,7 @@ def _init_simple_setup(self, local_path_state) -> pyompl.SimpleSetup:
         # Constructs a space information instance for this simple setup
         space_information = simple_setup.getSpaceInformation()
 
+        # figure this out
```
SPDonaghy, line 175: can leave this out for now, will be another task later

SPDonaghy (approved):
Great Work! Good to merge.

PR #493, opened 2025-02-24
User/adambrett40/259 global waypoint cache

### LINKED ISSUE #259: Global Waypoint Cache
### Purpose

Load last known global waypoints on start.

### Description

- Depends on UBCSailbot/sailbot_workspace#270 
- When we receive global waypoints, save the the data to a file `global_waypoint_cache` before de-serializing. If a file already exists, do not overwrite it immediately as we may lose power during an overwrite. If the file exists, then:
  1. Create a new file `global_waypoint_cache_temp`
  2. Write new global waypoint data to `global_waypoint_cache_temp`
  3. Move `global_waypoint_cache_temp` to `global_waypoint_cache`, overwriting it.
- On startup, check if a global waypoints file exists. If it does, read it, de-serialize it, and immediately publish it to ROS.

### Resources

- https://en.cppreference.com/w/cpp/filesystem/rename

adambrett40:
### Description

- Resolves #259 

- caching operations are done asynchronously
- updated ros2 node to run virtual iridium in the background, so the node starts properly in dev mode

### Verification

- Unit test ensures cache is correctly initialized and updated
- ros2 node retrieves and publishes cached data on startup
To view what's published on the global_path topic, need to do `export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/workspaces/sailbot_workspace/install/lib` in the terminal first as below (thanks Sean :))

### Resources

-

### DIFF
--- a/src/local_pathfinding/README.md
+++ b/src/local_pathfinding/README.md
@@ -12,11 +12,14 @@ Launch arguments are added to the run command in the format `<name>:=<value>`.
 
 | name        | description   | value                                                 |
 | ----------- | ------------- | ----------------------------------------------------- |
-| `log_level` | Logging level | A [severity level][severity level] (case insensitive) |
+| `log_level` | Logging level (default: `{INFO}`) | A [severity level][severity level] (case insensitive) `{DEBUG, INFO, WARN, ERROR, and FATAL}` |
+|`mode` | Mode (default: `{DEVELOPMENT}`) | `{DEVELOPMENT, PRODUCTION}`|
 
-[severity level]: <https://docs.ros.org/en/humble/Concepts/About-Logging.html#severity-level>
+## Some other important commands
 
-### Server Files
-
-The server files: `get_server.py` and `post_server.py` are basic http server files which are used for testing the
-global_path module's GET and POST methods.
+- `ros2 topic list`: Lists all the topics ([ROS 2 topic list documentation](https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Topics/Understanding-ROS2-Topics.html#ros2-topic-list))
+- `ros2 node list`: Lists all the nodes that are running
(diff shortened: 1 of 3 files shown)


SPDonaghy:
try running this command in the terminal before launching ros
`export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:/workspaces/sailbot_workspace/install/lib`

sailbot.code-workspace
```diff
@@ -508,19 +506,17 @@
...
                 "label": "setup",
                 "detail": "Set up the workspace",
                 "type": "shell",
                 "command": "./scripts/setup.sh",
                 "problemMatcher": []
             }
-
         ],
         "inputs": [
             {
```
samdai01, line 522: could you revert these small formatting changes? they dont seem to be related
if another lead request this be done please include it in another PR

src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
```diff
@@ -337,6 +369,18 @@ custom_interfaces::msg::Path LocalTransceiver::parseInMsg(const std::string & ms
     return soln;
 }
 
+std::optional<custom_interfaces::msg::Path> LocalTransceiver::getCache()
```
samdai01, line 372: please make this function async as well

src/network_systems/projects/local_transceiver/src/local_transceiver_ros_intf.cpp
```diff
@@ -91,6 +104,9 @@ class LocalTransceiverIntf : public NetNode
             sub_local_path_data = this->create_subscription<custom_interfaces::msg::LPathData>(
               ros_topics::LOCAL_PATH, ROS_Q_SIZE,
               std::bind(&LocalTransceiverIntf::sub_local_path_data_cb, this, std::placeholders::_1));
+
+            std::thread cache_thread(&LocalTransceiverIntf::getAndPublishCache, this);
```
samdai01, line 108: since the get function is now async, we do not need to launch it in a thread
please take a look at the 2019 answer in https://stackoverflow.com/questions/5645375/how-do-i-make-a-function-asynchronous-in-c

samdai01 (changes requested):
see comments

SPDonaghy:
@adambrett40 Is this still active?

PR #508, opened 2025-03-12
Raghumanimehta/492 mock wind sensor and gps local pathfinding

### LINKED ISSUE #492: Create Mock GPS and WindSensor Nodes
### Purpose

As we are almost ready to start testing the local_pathfinding package, we need some realistic mock data to calculate our paths with. The data left to mock that is relevant to local pathfinding is

- AIS
- GPS
- Wind Sensor

A mock AIS node is already in the works so we just need to get some GPS and Wind Sensor nodes running.
To keep things simple, we'll create a separate node to publish each topic, so this issue requires two nodes to be created.

### Description

The rough idea is that these nodes are structured very similarly to the navigate node but instead of calculating and publishing a desired heading, they publish realistic and arbitrary GPS and WindSensor data. 

@jamenkaye is working on test plans for pathfinding, so you can collaborate with him in determining how different mock scenarios will be specified/encoded and passed to your mock data nodes.

You may also want to explore how you could implement some sort of feedback loop that uses the published desired heading and the mock wind sensor data to generate more realistic values for the GPS topic that change over time.

### Resources

- [ROS2 Humble Writing simple publisher and subscriber](https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Publisher-And-Subscriber.html)
- Our navigate node is a good simple node to base your code off of: [node_navigate.py](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/node_navigate.py)
- The simulator team has a mockdata node that publishes to some of the same topics so you might find the code in this file useful: [mock_data_node.py](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/mock_data/mock_data_node.py
)

raghumanimehta:
### Description

- Resolves #492 (partially) 

Mock Wind Sensor now publishes data to the topic `filtered_wind_sensor`.
The wind speed is calculated using Weibull distribution.This distribution is a good fit for wind speed because is naturally positively skewed and has a long tail.
The wind direction is calculated using  von Mises distribution centered around the mean. This distribution is a circular distribution, which is perfect for wind direction.

### Verification

- [Verified using manual testing by observing that the node publishes to the ROS ] 

### Resources

-

### DIFF
--- a/src/local_pathfinding/global_paths/path_builder/path_builder.py
+++ b/src/local_pathfinding/global_paths/path_builder/path_builder.py
@@ -17,7 +17,6 @@ import re
 import webbrowser
 from os import walk
 from typing import Dict, List, Tuple
-
 import numpy as np
 import plotly.graph_objects as go
 from custom_interfaces.msg import HelperLatLon, Path
--- a/src/local_pathfinding/launch/main_launch.py
+++ b/src/local_pathfinding/launch/main_launch.py
@@ -70,6 +70,7 @@ def setup_launch(context: LaunchContext) -> List[Node]:
     launch_description_entities.append(get_navigate_node_description(context))
     if mode == "development":
         launch_description_entities.append(get_mock_global_path_node_description(context))
+    launch_description_entities.append(get_mock_wind_sensor_node_description(context))
     return launch_description_entities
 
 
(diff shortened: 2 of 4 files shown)


SPDonaghy (changes requested):
Nice work, love that bound to 180 makes an appearance. Could you please resolve the merge conflict with main and this is good to go.

SPDonaghy (approved):

PR #509, opened 2025-03-12
Obstacle Tracking and State Validity Checker

### LINKED ISSUE #474: Add Obstacle Collisions to the OMPL StateValidityChecker
### Purpose

After the land obstacles and geometry have been represented, we need to add the geometries created for StateValidtyChecker. 

### Description

[In file ompl_path.py](https://github.com/UBCSailbot/sailbot_workspace/blob/af1574073331b8fbcde8a1fcda228a0975c1884d/src/local_pathfinding/local_pathfinding/ompl_path.py#L203C5-L203C46) we have to use the land obstacle class and configure the OMPL Space Information to use the geomerties to check whether the selected points are within the geometry.

Update:
Obstacle tracking is implemented now. Unit tests pass, but I'd like to put the validity checker through some more thorough tests with ROS.

### Resources

-  [SpaceInformation class documentation](https://ompl.kavrakilab.org/classompl_1_1base_1_1SpaceInformation.html#a0449ec0893e02aca2bdfc63323b624e5)

SPDonaghy:
### Description

- Resolves #474 

Added a static `obstacles` attribute to the OMPLPath class that will store all obstacles within the state space.

The obstacles are in a static list so that the static function is_state_valid can access it during runtime.

A pointer to the is_state_valid function is passed to OMPL so it must also be static.

### Verification

- [x] Unit tests on basic functionality for OMPLPath class regarding initializing obstacles based off of a mock LocalPathState
- [x] ROS runs currently
- [x] Still would like to run more thorough tests to verify is_state_valid is able to access obstacles when it is called through bindings by OMPL.

My idea to verify the last point: choose a new magic number GPS point that is near enough to land to guarantee that some Land obstacles will be instantiated, log all state validity checks to a log file and verify that some states were valid and some were not along with the type of obstacle in which the invalid state lies.

Here is the log output verifying that while ROS is running, OMPL is able to use my state validity checker function to correctly invalidate states inside land obstacles.
```
State at (-7138.70,10031.93) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7147.07,10052.61) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7148.41,10055.14) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7132.68,10006.47) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7142.94,10062.80) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7144.61,9996.71) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
State at (-7143.91,10070.56) was invalidated by obstacle: <class 'local_pathfinding.obstacles.Land'>
```

### DIFF
--- a/src/local_pathfinding/local_pathfinding/ompl_path.py
+++ b/src/local_pathfinding/local_pathfinding/ompl_path.py
@@ -219,13 +219,12 @@ class OMPLPath:
         Returns:
             bool: True if state is valid, else false.
         """
-        # note: `state` is of type `SE2StateInternal`, so we don't need to use the `()` operator.
         state_is_valid = True
 
         for o in OMPLPath.obstacles:
             state_is_valid = o.is_valid(cs.XY(state.getX(), state.getY()))
             if not state_is_valid:
-                return state_is_valid  # return False immediately once state is found to be invalid
+                return state_is_valid
 
         return state_is_valid
 
--- a/src/local_pathfinding/local_pathfinding/ompl_path.py
+++ b/src/local_pathfinding/local_pathfinding/ompl_path.py
@@ -57,8 +57,6 @@ class OMPLPath:
         """
         self._box_buffer = 1
         self._logger = parent_logger.get_child(name="ompl_path")
-
-        self.obstacles = self.init_obstacles(local_path_state)
         self._simple_setup = self._init_simple_setup(local_path_state)  # this needs state
 
         self.solved = self._simple_setup.solve(time=max_runtime)  # time is in seconds
@@ -68,7 +66,9 @@ class OMPLPath:
         #     # try to shorten the path
         #     simple_setup.simplifySolution()
 
-    def init_obstacles(self, local_path_state: LocalPathState) -> List[Polygon]:
+    def init_obstacles(
+        self, local_path_state: LocalPathState, state_space_xy: Polygon
+    ) -> List[Polygon]:
         """Extracts obstacle data from local_path_state and compiles it into a list of Polygons
 
         Places Boats first in the list as states are more likely to conflict with Boats than Land.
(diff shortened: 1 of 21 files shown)


SPDonaghy:
Commenting so I don't forget, I am working on validating the current implementation but I would like to optimize the process of generating a path because right now it is wasteful (file I/O for every OMPLPath object created)

SPDonaghy:
Reminder: return true, false for better readability in isStateValide

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -189,19 +259,40 @@ def _init_simple_setup(self, local_path_state) -> pyompl.SimpleSetup:
...
-        bool: True if state is valid, else false.
+        for o in OMPLPath.obstacles:
+            state_is_valid = o.is_valid(cs.XY(state.getX(), state.getY()))
+            if not state_is_valid:
+                # uncomment this if you want to log which states are being labeled invalid
+                # its commented out for now to avoid unnecessary file I/O
+                # log_invalid_state(state=cs.XY(state.getX(), state.getY()), obstacle=o)
+                return state_is_valid
+
+        return state_is_valid
```
jamenkaye, line 282: Can likely simplify this a bit, if it finds a false state in the loop then return false, otherwise just return true at the end

jamenkaye (approved):
Looks good! Just the 1 suggestion :)

src/local_pathfinding/local_pathfinding/ompl_path.py
SPDonaghy: done

PR #514, opened 2025-03-16
Migrate back to using official OMPL whl files

SPDonaghy:
### Description

The purpose was to remove anything related to our custom ompl bindings and change all references back to the official ompl package.

I also added a command to install the appropriate whl file (for x86_64 or aarch64) within the build task.

### Verification

- [x] Ensured all unit tests pass
- [x] Ensured all local pathfinding nodes launched and ran properly
- [x] Kept the ompl logging level at debug for now, to validate that it is solving paths and using the state validity checker function during runtime
- [x] Verified via logging that state validity checker func is operating as expected while ros is running
- [x] Validate changes on RPI

### Note
Now that we have a whl file for ompl, we won't need to have the ompl installation be a part of our container image anymore!
I will make the neccessary changes to the base dockerfile in a separate pr.


jamenkaye (approved):
Thanks for taking care of this! Before you merge, is it worth double-checking that everything works on the raspberry pi?

SPDonaghy:
> Thanks for taking care of this! Before you merge, is it worth double-checking that everything works on the raspberry pi?

No problem! And yes definitely worth checking, I will do that!

PR #517, opened 2025-03-19
Mock Global Path Node to publish to ROS instead of http POST + node housekeeping

SPDonaghy:
### Description

Previously, the mock global path node was attempting to retrieve the GPS data and publish global path via http requests. 

I modified the node to do both operations through ROS with a gps subcriber and a global path publisher.

I also made some small readability changes while I was tooling around in the file.

### Verification

- [x] Once the mock GPS node is ready, I can test the functionality of the mgp node. Waiting on #492 
- [x] Verify that logging looks nice on launch


SPDonaghy:
This is done

PR #467, opened 2024-11-30
User/medhansh/desired heading

MC6527medhansh:
### **Description**
- Resolves #283  
- Implemented the functionality to compute and publish the desired heading from the current boat position to the next waypoint.

---

### **Verification**
- [ ] Ensure the desired heading is correctly calculated using the `Geod` class from the `pyproj` library for distance and azimuth computation.
- [ ] Validate that the system updates the waypoint index correctly when the boat reaches a waypoint using a threshold distance.
- [ ] Check scenarios where the path is updated to ensure the waypoints start from the boat's current position.

---

### **Resources**
- UBC Sailbot Documentation
- Pyproj Documentation for Geodesic Calculations (https://pyproj4.github.io/pyproj/stable/api/geod.html)

---

PR #527, opened 2025-03-24
Get desired heading

### LINKED ISSUE #283: Publish desired heading

jamenkaye:
### Description

- Resolves #283 by implementing `get_desired_heading()`
- Based on PR #467 ; this addresses the feedback and fixes merge conflicts

### Verification

- I'm hoping to discuss this more; not sure how to verify this.
   - Option 1: Put all the logic in a helper function that takes arguments rather than working with attributes, and test the helper
   - Option 2: test `node_navigate.py` with mocks?


jamenkaye:
This should be ready to go! I've moved most of the logic from get_desired_heading to a helper function, which I've added tests for. Also added `bound_to_180()` and tests :)

PR #529, opened 2025-03-24
Only run docs workflow when docs is modified plus once weekly

### LINKED ISSUE #526: Docs update CI workflow is failing constantly
#### Expected Behavior

workflow doesn't fail. Especially when no changes have been made to docs.

#### Actual Behavior

Workflow fails at first and then is successful after 1-2 retries.

### Steps to Reproduce

1.  open a PR or merge a PR into main

### Potential Solutions

Looks like exit code 137 which means the host is exceeding the maximum memory usage when running the workflow.
I am going to configure docs checks to only run on prs that modify the docs from now on.

SPDonaghy:
### Description

- Resolves #526 

I am just going to run the docs workflow much less frequently so we don't have to keep rerunning the action when it runs out of memory.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,12 +1,18 @@
 name: Docs Site
 
 on:
-  push:
-    branches:
-      - main
-  pull_request:
-    # default types and closed
-    types: [opened, synchronize, reopened, closed]
+    push:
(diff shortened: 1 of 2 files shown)


SPDonaghy:
need to make the docs CI not required before merging this

Issue #476, opened 2025-01-18
Create Mock AIS for Local Pathfinding


FireBoyAJ24:
### Purpose

The mock AIS will be used to run the local pathfinding under development. This will allow us to couple the publisher data to the OMPL path classes and have the full pathfinding system function for deployment.

### Description

There can be files (JSON, YAML) containing the publisher AIS that will be parsed and published in the ROS network. The Pathfinding subscribers will receive these inputs, generate the OMPL path, and publish it. 

### Resources

- https://github.com/UBCSailbot/raye-local-pathfinding/
- [ROS2 Humble Writing simple publisher and subscriber](https://docs.ros.org/en/humble/Tutorials/Beginner-Client-Libraries/Writing-A-Simple-Py-Publisher-And-Subscriber.html)
- Our navigate node is a good simple node to base your code off of: [node_navigate.py](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/node_navigate.py)
- The simulator team has a mockdata node that publishes to some of the same topics so you might find the code in this file useful: [mock_data_node.py](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/mock_data/mock_data_node.py
)
- The other mock data node task for pathfinding [Mock GPS and WindSensor](https://github.com/UBCSailbot/sailbot_workspace/issues/492)
- An example of creating a mock AIS ship in our unit tests is [here](https://github.com/UBCSailbot/sailbot_workspace/blob/59392457dfdb52b967e76d08d7115a6c86c192ae/src/local_pathfinding/test/test_obstacles.py#L297-L305)

PR #541, opened 2025-03-30
Implemented ability to add or remove AIS ships with CSV

rnar17:
### Description
- Adds to #476 
- Updated the Mock AIS Node to allow for addition and removal of AIS ships through a CSV file during runtime. 

### Verification
- Manual verification that the addition / removal of a row in the csv file changes the output logs.


SPDonaghy (approved):
nice

PR #551, opened 2025-04-21
Install ompl from PyPi in build.sh

SPDonaghy:
Remove scripts for downloading and installing ompl from github as ompl is now available on PyPi.


jamenkaye (approved):
Looks good to me!

SPDonaghy:
Waiting on this because ompl 1.7.0 has a memory error that is not in 1.6.0, but 1.6.0 is not available on pypi

SPDonaghy:
I can't figure out why thats happening, I am just going to have ompl install using pip3 in the build.sh script instead.

PR #553, opened 2025-04-21
Raghumanimehta/531 upgraded mock gps wind sensor

### LINKED ISSUE #531: Upgrade mock wind sensor and gps nodes
### Purpose

Upgrade the mock wind sensor and gps nodes to publish more realistic data that depends more on previous values.

### Mock WindSensor
**Goal:** we want to be able to pass in arbitrary mean wind speed and mean direction values from our test plans.

**Idea:** Configure the mock wind sensor node's mean wind speed and mean direction to be read from a csv file whose file path is determined from a ROS parameter so that we can change it during runtime and switch to different wind conditions without editing any code.

Other ideas?

### Mock GPS
Assuming sailbot perfectly achieves the previous desired heading as its new current heading is good enough.

**Goal:** We want the boat speed to update dynamically depending on mocked wind conditions.

There is a table of predetermined values for expected boat speed depending on wind speed and direction. You can find this table [HERE](https://github.com/UBCSailbot/sailbot_workspace/blob/1f2c97b3c528366b8ac7cb7899b3ac557db5c43c/src/local_pathfinding/local_pathfinding/objectives.py#L21-L32)

There is a function in `objectives.py` called get_sailboat_speed that interpolates this data to a continuous function.

**Idea:** We can calculate the mock gps boat speed by using the random mock windsensor data and an interpolation of the boat speed data provided above.

raghumanimehta:
### Description

- Resolves #531
- Mock wind sensor data can be set during runtime now by using `ros2 param set /mock_wind_sensor mean_wind_speed {value}` and  `ros2 param set /mock_wind_sensor mean_direction {value}`.
- Fixed a bug with `ensure_correct_range` in `node_mock_wind_sensor.py` that was not handling direction values less than -179.


src/local_pathfinding/local_pathfinding/node_mock_wind_sensor.py
```diff
@@ -97,9 +105,32 @@ def ensure_correct_range(self, direction_value: int) -> int:
...
-            return (360 - (direction_value % 360)) * -1
-        return direction_value
+
+        normalized_direction = direction_value % 360
+        if normalized_direction > 180:
+            normalized_direction -= 360
+
+        return normalized_direction
+
+    def get_latest_speed_and_direction_values(self) -> None:
```
SPDonaghy, line 115: I think you can remove the docstring from this one, its pretty self explanatory.

src/local_pathfinding/local_pathfinding/node_mock_wind_sensor.py
```diff
@@ -97,9 +105,32 @@ def ensure_correct_range(self, direction_value: int) -> int:
         Returns:
             int: The corrected wind direction in degrees.
         """
-        if direction_value > 180:
-            return (360 - (direction_value % 360)) * -1
-        return direction_value
+
```
SPDonaghy, line 108: Can replace this function with a call to `coord_systems.bound_to_180`

src/local_pathfinding/local_pathfinding/node_mock_wind_sensor.py
```diff
@@ -75,6 +81,7 @@ def get_mock_wind_speed(self) -> ci.HelperSpeed:
         """
         # Shape=2 (typical for wind under normal calm conditions), Scale=7 knots (mean)
 
+        self.get_latest_speed_and_direction_values()
```
SPDonaghy, line 84: can call this function once, in `mock_wind_sensor_callback` before calling `get_mock_wind_speed`

src/local_pathfinding/local_pathfinding/node_mock_wind_sensor.py
```diff
@@ -22,7 +22,11 @@ def __init__(self):
             __mock_wind_sensor_timer (Timer): Timer to call the mock wind sensor callback function.
             __wind_sensors_pub (Publisher): Publisher for the filtered wind sensor data.
             __mean_wind_speed (float): Mean wind speed in the Pacific Ocean near Vancouver in July.
+            This parameter can be set during runtime using
```
SPDonaghy, line 25: could you move info like this to a brief confluence doc page on how to use this node please? Same with the notes on the types of distributions used for random data. Would be good to have that all on confluence, just a few bullet points.

raghumanimehta: Sure!

raghumanimehta (commented):
I have made the change you requested.

SPDonaghy:
Hey @raghumanimehta I'm not seeing the new changes, I think you forgot to push.

SPDonaghy (approved):

PR #555, opened 2025-04-26
Add GitHub Action to Validate Website Build on All Branches

JordanChen123:
### Description

Adding a GitHub Actions workflow to automatically build the website for every commit and push.

This workflow ensures that the website is always in a buildable state by running npm run build, catching any type errors, lint issues, or build failures early during development.

It triggers on pushes to main as well as to any branches branching off main.

### Verification

- [ ] 

### Resources

-

### DIFF
--- a/.github/workflows/website-build.yml
+++ b/.github/workflows/website-build.yml
@@ -1,0 +1,29 @@
+name: Website Build Check
+
+on:
+  push:
+    branches:
+      - main
+      - '**'
+  pull_request:
+
+jobs:
+  build-website:
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v4
+        with:
+          node-version: 20
+
+      - name: Install dependencies
+        run: npm install
+        working-directory: src/website
+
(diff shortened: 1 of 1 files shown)


.github/workflows/website-build.yml
```diff
@@ -0,0 +1,29 @@
+name: Website Build Check
+
+on:
```
SPDonaghy, line 3: Hey Jordan, my suggestion is we restrict what triggers this workflow a bit more.
My thinking is that it should only run when files related to the website have been changed.
I did this for the `Docs Site` workflow and it really sped things up for PRs and pushes that didn't involve docs src files.

What are your thoughts?

.github/workflows/website-build.yml
```diff
@@ -0,0 +1,29 @@
...
+  push:
+    branches:
+      - main
+      - '**'
```
SPDonaghy, line 7: I also don't think it should run on pushes to any other branch than `main` since being triggered on pull requests should catch anything before we merge it in.

.github/workflows/website-build.yml
JordanChen123: Yes I completely agree, this would be a lot more efficient. Will add this change

SPDonaghy:
@JordanChen123 Is this still active?

PR #558, opened 2025-05-14
Base Pathfinding Visualizer 

### LINKED ISSUE #533: Pathfinding Visualizer
### Purpose

We need a tool to visualize our pathfinding software in action so we can make better informed decisions and properly evaluate the current system.

### Description

Ideas:

It will be some sort of GUI with an API that supports loading in test plans and possibly allows for changes to test conditions on the fly. 

It should also support some form of logging.

It should be able to run in a headless mode for long term tests.

### Suggestions
From my previous research in creating a GUI for the global path builder, Flask was simple enough to use and we know it works within the docker container.

Other python GUI frameworks are more difficult to configure from within a container as they are not really designed for such an application and don't have the ability to display windows on your desktop right out of the box.

FireBoyAJ24:
### Description

- Resolves #533 mostly
   

-  The visualizer uses plotly with Dash's Flask app as the GUI. The app runs in parallel with the sailbot observer node through the use of multiprocessing
- The `VisualizerState` stores all the sailbot's GPS and local waypoint locations and visualizes it as an animated graph every 10 seconds.

### Verification

- [ ] Run all nodes and flask app for 10-20 minutes (minimum)
- [ ] The waypoints change if the wind speed and direction change.

### Resources

- [Visualizer Design Specification](https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/2338455564/Visualizer+Design+Specification)


FireBoyAJ24:
The `VisualizerState` will need to be refactored into multiple smaller classes to improve readability and scalability. The question of how to store the converted lat/lons and future message types can be a nice data structure question.

FireBoyAJ24:
@JordanChen123, should I delete the `package-lock.json` changes from this PR?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
...
+from local_pathfinding.coord_systems import latlon_to_xy
+
+app = dash.Dash(__name__)
+
+queue: Optional[Queue] = None  # type: ignore
+
+
+class VisualizerState:
+    """
+    Converts the ROS message to a format that can be used by the visualizer.
```
SPDonaghy, line 20: Could you explain the function of the class a bit more?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
+from multiprocessing import Queue
```
SPDonaghy, line 1: Could you add a docstring at the top to give a very high level overview of the structure of the visualizer? 
Eg. how the observer node, dash app, and the VisualizerState class/queue work together?

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
```diff
@@ -25,22 +37,86 @@ class SailbotObserver(Node):
...
+        self.prev_msg = None
+        self.queue = queue
+
+    def local_path_callback(self, msg):
+        if not self.is_msg_different(msg):
+            self.get_logger().debug("Message is the same as previous, skipping update.")
+            return
+
+        self.prev_msg = msg
+        self.msgs.append(msg)
```
SPDonaghy, line 65: reminder to add a size limit to the msgs collection

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
```diff
@@ -25,22 +37,86 @@ class SailbotObserver(Node):
...
+
+        self.get_logger().debug(f"Received new local path message: {msg}")
+
+        self.state = visualizer.VisualizerState(msgs=self.msgs)
```
SPDonaghy, line 69: ditto on the size limit

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
```diff
@@ -25,22 +37,86 @@ class SailbotObserver(Node):
...
+        if self.prev_msg.global_path.waypoints != msg.global_path.waypoints:
+            return True
+        if self.prev_msg.gps != msg.gps:
+            return True
+
+        return False
+
+
+def main():
+    """Creates two processes, one for the ROS node and one for the Dash app.
```
SPDonaghy, line 98: could maybe move this comment up to the top of the file

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
```diff
@@ -25,22 +37,86 @@ class SailbotObserver(Node):
...
+    The ROS node subscribes to the local_path topic and the Dash app visualizes
+    the data.
+    """
+    manager = Manager()
```
SPDonaghy, line 102: a comment to explain what this is would help readability

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
...
+queue: Optional[Queue] = None  # type: ignore
+
+
+class VisualizerState:
+    """
+    Converts the ROS message to a format that can be used by the visualizer.
+    """
+
+    def __init__(self, msgs: List[ci.LPathData]):
+        if not msgs:
```
SPDonaghy, line 24: can this ever happen? I think at worst the msgs list would be empty right?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
...
+        If not, it raises a ValueError.
+        """
+        if not msg.local_path:
+            raise ValueError("local path must not be None")
+        if not msg.global_path:
+            raise ValueError("global path must not be None")
+        if not msg.gps:
+            raise ValueError("gps must not be None")
+
+    def _convert_to_xy(self, lat_lon_list: List[ci.HelperLatLon]) -> List[cs.XY]:
```
SPDonaghy, line 59: can you move this function to coord systems?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
...
+
+
+def live_update_plot(state: VisualizerState) -> go.Figure:
+    """
+    Updates the live graph to the latest path planning data.
+    """
+
+    fig = initial_plot()
+
+    # Plotting Local Waypoints
```
SPDonaghy, line 145: can you change this to # start point

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -0,0 +1,359 @@
...
+    start_trace = go.Scatter(
+        x=[state.final_local_wp_x[0]],
+        y=[state.final_local_wp_y[0]],
+        mode="markers",
+        marker=dict(color="green", size=10),
+        name="Start",
+    )
+
+    # Intermediate waypoints (blue)
```
SPDonaghy, line 154: and this to # local path waypoints. I think it would be more clear

SPDonaghy (commented):
Great execution with this. I really like the use of the queue and background processes to get everything working together.

It would be really nice if there was a way to have the top chart stay zoomed out when the plot updates but I have a hunch that might not be possible as I've struggled with animated charts before...

src/local_pathfinding/local_pathfinding/visualizer.py
FireBoyAJ24: This would unlikely happen unless the navigate node shutdown before sending any LPathData message,

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
FireBoyAJ24: I will set the size limit to 100 messages for now.

PR #560, opened 2025-05-17
Obstacle Initialization and Tracking Rework

rnar17:
### Description

This PR implements the ability to inject mock land obstacles via a ROS parameter (`use_mock_land`) to the navigate node, which reads mock land coordinates (in latlon format) from [mock_land.json](https://github.com/UBCSailbot/sailbot_workspace/blob/c69bf6ba3bf5c8af479a3f63fbdf05b00f9ebbf1/src/local_pathfinding/land/mock/mock_land.json ). 

The PR edits the navigate node to read the mock land data as shown below 
```Python
# Initialize mock land obstacle
self.land_multi_polygon = None
if self.use_mock_land:
    with open(self.mock_land_file, "r") as f:
        data = json.load(f)
        polygons = [Polygon(p) for p in data.get("land_polygons", [])]
        self.land_multi_polygon = MultiPolygon(polygons)
        self.get_logger().info("Loaded mock land data.")
```

It also modifies OMPLPath and LocalPath to make the mock land data accessible at the obstacle initialization scope. 

This PR also tackles obstacle persistence across calls to `desired_heading_callback` which I accomplished by adding the variables `ais_ships` and `current_ship_ids` to OMPLPath to keep track of previously used obstacles so that we only add new obstacles (logic shown below).

```Python
for ship in ais_ships:
  if ship.id in OMPLPath.obstacles:
      # If the ship is already in the obstacles, update its information
      existing_boat = OMPLPath.obstacles[ship.id]
      existing_boat.update_sailbot_data(
          sailbot_position=local_path_state.position,
          sailbot_speed=local_path_state.speed,
      )
      existing_boat.update_collision_zone()
  else:
      # Otherwise, create a new Obstacle object
      new_boat = ob.Boat(
          local_path_state.reference_latlon,
          local_path_state.position,
          local_path_state.speed,
          ship,
      )
      OMPLPath.obstacles[ship.id] = new_boat

```

### Verification

- [x] Main launch runs with no errors
- [x] Mock land obstacles appear with correct coordinates in obstacle list during runtime
- [x] AIS ship obstacles still appear in obstacle list
- [x] State validity checker works as expected (injected mock land obstacles near boat which logged invalid message)
- [x] Relevant unit tests pass, including a newly added one for ompl path initialization regarding updating obstacles and avoiding duplicates


src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -89,21 +95,36 @@ def init_obstacles(
                                        obstacles
 
         """
+
         OMPLPath.obstacles = []
 
         # BOATS
         ais_ships = local_path_state.ais_ships
+        ship_map = {ship.id: ship for ship in ais_ships}
```
SPDonaghy, line 103: I think we could just iterate through `ais_ships` no? I don't think converting to a map will have any performance gains.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -228,8 +237,19 @@ def get_desired_heading(self) -> float:
         Returns:
             float: The desired heading
         """
+
+        # Initialize mock land obstacle
+        multi_land_polygon = None
+        if self.use_mock_land:
+            with open(self.mock_land_file, 'r') as f:
```
SPDonaghy, line 244: Can you find a way to do the file I/O only once on launch?

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -89,21 +95,35 @@ def init_obstacles(
...
+        for ship_id, ship in ship_map.items():
+            if ship_id in existing_obstacles:
+                # If the ship is already in the obstacles, update its information
+                existing_boat = existing_obstacles[ship_id]
+                existing_boat.update_sailbot_data(
+                    sailbot_position=local_path_state.position,
+                    sailbot_speed=local_path_state.speed,
+                )
+                existing_boat.update_collision_zone()
+                OMPLPath.obstacles.append(existing_boat)
```
SPDonaghy, line 115: My concern is this will cause duplicates in the `obstacles` list. I am also guilty of this with my initial code.

I like that you made `existing_obstacles` a dictionary for efficient lookups, so what about this:

What if we made OMPLPath's `obstacles` a dictionary and then we wouldn't need `existing_obstacles`. 

This would also require giving the one land obstacle some arbitrary id so that we can update it rather than creating a new one to insert each time this function is called.

Could you also add some logic to evict Boat obstacles from the `obstacles` dictionary if it didn't appear in the new `ais_ships` message?

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -117,18 +137,17 @@ def init_obstacles(
 
         if OMPLPath.all_land_data is None:
             try:
-                LAND = load_pkl(
-                    "/workspaces/sailbot_workspace/src/local_pathfinding/land/pkl/land.pkl"
-                )
-            except RuntimeError as e:
+                OMPLPath.all_land_data = load_pkl("../land/pkl/land.pkl")
```
SPDonaghy, line 140: nice catch

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -288,10 +310,10 @@ def is_state_valid(state: Union[base.State, base.SE2StateInternal]) -> bool:
                 # uncomment this if you want to log which states are being labeled invalid
                 # its commented out for now to avoid unnecessary file I/O
 
-                # if isinstance(state, base.State):  # only happens in unit tests
-                #     log_invalid_state(state=cs.XY(state().getX(), state().getY()), obstacle=o)
-                # else:  # happens in prod
-                #     log_invalid_state(state=cs.XY(state.getX(), state.getY()), obstacle=o)
+                if isinstance(state, base.State):  # only happens in unit tests
```
SPDonaghy, line 313: can comment this out again so we don't have this logging happening in main branch

SPDonaghy (approved):
Really nice work

PR #576, opened 2025-06-01
Pathfinding Mock GPS node fixed 

raghumanimehta:
### Description

- Resolves no issue (minor fix)

- Made updates to how get_sailbot_speed was used by converting apparent wind to true wind

### Verification

- All tests passed and manual verification that the ros doesn't crash

### DIFF
--- a/src/local_pathfinding/local_pathfinding/node_mock_gps.py
+++ b/src/local_pathfinding/local_pathfinding/node_mock_gps.py
@@ -5,16 +5,107 @@ Mock class for the GPS. Publishes basic GPS data to the ROS network.
 import math
 
 import custom_interfaces.msg as ci
+import numpy as np
 import rclpy
 from geopy.distance import great_circle
 from rclpy.node import Node
-from local_pathfinding.objectives import SpeedObjective
 
+from local_pathfinding.objectives import get_true_wind
 
 MEAN_SPEED = ci.HelperSpeed(speed=15.0)  # mean boat speed in kmph
 
 START_POINT = ci.HelperLatLon(latitude=48.46, longitude=-125.1)  # Starting location of the mock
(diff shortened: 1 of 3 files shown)


src/local_pathfinding/local_pathfinding/node_mock_gps.py
```diff
@@ -5,18 +5,109 @@
...
+    # Find the nearest windspeed values above and below the true windspeed
+    lower_windspeed_index = max([i for i, ws in enumerate(WINDSPEEDS) if ws <= true_wind_speed])
+    upper_windspeed_index = (
+        lower_windspeed_index + 1
+        if lower_windspeed_index < len(WINDSPEEDS) - 1
+        else lower_windspeed_index
+    )
+
+    # Find the nearest angle values above and below the sailing angle
+    lower_angle_index = max([i for i, ang in enumerate(ANGLES) if ang <= sailing_angle])
```
FireBoyAJ24, line 65: This finds the maximum nearest angle to the sailing angle. For example, if the sailing angle is 29 then instead of taking 30 it will take 20 degrees. Is that correct?

FireBoyAJ24: Could have found the angle which had the minimum distance from the sailing angle and done min(min_ANGLE) array, like in this [question](https://stackoverflow.com/questions/2566412/find-nearest-value-in-numpy-array).

FireBoyAJ24: I don't think it would be a good idea to select the larger angle if the sailing angle is close to the smaller angle.

raghumanimehta: I get your concern. It seems the better solution would be to round to the closest angle and not always going with the larger angle. However, I don't have the sailing knowledge to gauge what the original author intended. If you want, I can change this function.

PR #588, opened 2025-06-19
Visualize Wind Vectors and Land Obstacles on Visualizer

rnar17:
### Description

This PR enhances the Sailbot path planning visualizer by adding visual representations of wind vectors (apparent, true, and boat wind) at the bottom right as well as land obstacles. These updates allow for more informed debugging of the boat’s movement. 

image

A mock land obstacle is plotted near the centre of the grid too. The speed of each wind vector is displayed when hovering over its tip, however it can be somewhat difficult to position the mouse precisely enough to trigger the display.

### Verification

- [x] ros2 launch local_pathfinding main_launch.py runs with no errors
- [x] Mock land obstacles appear with correct coordinates on visualizer 
- [x] Apparent wind ≈ true wind + boat wind (tip-to-tail vector addition)


src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -82,8 +85,35 @@ def __init__(self, msgs: List[ci.LPathData]):
...
+        true_wind = get_true_wind(
+            self.curr_msg.filtered_wind_sensor.direction,
+            self.curr_msg.filtered_wind_sensor.speed.speed,
+            boat_heading,
+            boat_sog,
+        )
+        self.true_wind_vector = self._process_true_wind_vector(true_wind)
+
+        # boat wind vector
+        boat_wind_radians = math.radians(cs.bound_to_180(boat_heading + 180))
```
FireBoyAJ24, line 110: It would a good idea to add a comment over why we are adding 180 to the boat heading. For example, the boat's wind direction is 180 to the heading of the boat.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -208,14 +280,164 @@ def live_update_plot(state: VisualizerState) -> go.Figure:
...
+        ay=actual_y0,
+        xref="x",
+        yref="y",
+        axref="x",
+        ayref="y",
+        showarrow=True,
+        arrowhead=3,
+        arrowsize=0.5,
+        arrowwidth=3,
+        arrowcolor="purple",
```
FireBoyAJ24, line 332: The purple arrow made me confused over whether I was looking at the boat's heading or apparent wind vector. The apparent wind vector could be added to the box with the true and boat wind, and we can replace the apparent wind vector (purple arrow) with the boat's heading since we cannot illustrate that visually with the triangle icon.

FireBoyAJ24 (commented):
Should the wind vector arrows go towards the boat rather than away the boat? For example, in this image:
image

PR #589, opened 2025-06-21
Update lookup table and publisher description

kristatraboulay:
### Description
Updated lookup table to include 0.2 error for alpha. Added short description of this error and a publisher description in the wingsail control node

### Verification
- Successfully built and tested controller package, ran flake8 successfully


eomielan (approved):
LGTM

Issue #582, opened 2025-06-10
Add GPS Data to the Pathfinding Visualizer


SPDonaghy:
### Purpose

Ultimately we want to represent all the data in the [local path topic](https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/custom_interfaces/msg/LPathData.msg) in the visualizer and this task is one part of that larger goal.

### Changes

Implement the code required to display the GPS data in the pathfinding visualizer GUI. 

Files which need to be modified:
* `visualizer.py`

### Resources
The visualizer in visualizer.py shares a queue with the sailbot observer node. The observer puts VisualizerState objects in the queue, these objects contain all the data packaged in the local_path ros topic message. And then a dash app pulls a VisualizerState off the queue, parses it for all the pathfinding data. and plots that data in the graph and updates the graph

PR #591, opened 2025-06-21
Raghumanimehta/582 advanced gps visualizer

raghumanimehta:
# Description
- Resolves https://github.com/UBCSailbot/sailbot_workspace/issues/582
- Boat marker points in the direction of boat heading
- The goal marker shows the angle it is at from the boat's perspective
# Verification
- Changes to the plot were manually verified
- The new coordinate system function was thoroughly tested.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/coord_systems.py
+++ b/src/local_pathfinding/local_pathfinding/coord_systems.py
@@ -42,7 +42,7 @@ def true_bearing_to_plotly_cartesian(true_bearing: float) -> float:
         Increases in the clockwise direction till 180 degrees.
         Decreases in the counter-clockwise direction till -180 (exclusive)
     Returns:
-        float:  Angle where 0 is north and values increase counter-clockwise.
+        float:  Angle where 0 is north and values increases clockwise.
     """
     assert -180 < true_bearing <= 180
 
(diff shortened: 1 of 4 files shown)


src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -34,6 +34,24 @@ def cartesian_to_true_bearing(cartesian: float) -> float:
...
 
+def true_bearing_to_plotly_cartesian(true_bearing: float) -> float:
+    """Convert a true bearing angle to the equivalent cartesian angle .
+
+    Args:
+        true_bearing (float): Angle where 0 is true north. Range: -180 < heading <= 180.
+        Increases in the clockwise direction till 180 degrees.
+        Decreases in the counter-clockwise direction till -180 (exclusive)
+    Returns:
+        float:  Angle where 0 is north and values increase counter-clockwise.
```
FireBoyAJ24, line 45: Did you mean clockwise? If the angle increases in a counter-clockwise then 90 degrees would be west and not east.

raghumanimehta: Yeah, I messed up the documentation. Plotly has a strange convention that goes from 0 to 360, where 0 is North, 90 is East, 180 is South and 270 is East. My bad, I'll fix it.

FireBoyAJ24 (approved):

PR #598, opened 2025-07-11
Added AIS ships to visualizer with ID

rnar17:
### Description
This PR enhances the Sailbot path planning visualizer by adding AIS ships with their XY coordinates along with their ID.

### Verification

- [x] ros2 launch local_pathfinding main_launch.py runs with no errors
- [x] AIS Ships appear with correct coordinates on visualizer


FireBoyAJ24 (commented):
Could we have the AIS ship's symbol be the same as the sailbot and point in the direction of the heading?

PR #642, opened 2025-10-01
Raghumanimehta/conditions to switch path

### LINKED ISSUE #287: Conditions to switch the local path

raghumanimehta:
- Resolves  #287 
### Changes made
- `update_if_needed` in `LocalPath` is complete now. 
- Implemented `get_cost` in `OMPLPath`
- Refactored `calculate_desired_heading_and_waypoint_index` such that it is now in `LocalPath`
- Implemented and testsed `in_collision_zone`
- Updated documentation of a couple of things
### Verification
- `in_collision_zone` was tested using unit tests
- manual verification of `update_if_needed` as a whole 
- TODO (separate task): test`update_if_needed` and its helpers except `in_collision_zone`

### DIFF
--- a/src/local_pathfinding/local_pathfinding/ompl_path.py
+++ b/src/local_pathfinding/local_pathfinding/ompl_path.py
@@ -17,11 +17,11 @@ from ompl import base
 from ompl import geometric as og
 from ompl import util as ou
 from rclpy.impl.rcutils_logger import RcutilsLogger
-from shapely.geometry import MultiPolygon, Point, Polygon, box
+from shapely.geometry import MultiPolygon, Polygon, box
 
 import local_pathfinding.coord_systems as cs
 import local_pathfinding.obstacles as ob
-from local_pathfinding.ompl_objectives import get_sailing_objective
+from local_pathfinding.ompl_objectives import get_sailing_objective, create_buffer_around_position
 
 if TYPE_CHECKING:
     from local_pathfinding.local_path import LocalPathState
@@ -30,6 +30,7 @@ if TYPE_CHECKING:
 ou.setLogLevel(ou.LOG_WARN)
 
 BOX_BUFFER_SIZE = 1.0  # km
+
 LAND_KEY = -1
 CURRENT_DIR = os.path.dirname(os.path.abspath(__file__))
 LAND_PKL_FILE_PATH = os.path.join(CURRENT_DIR, "..", "land", "pkl", "land.pkl")
@@ -229,13 +230,6 @@ class OMPLPath:
 
         return ci.Path(waypoints=waypoints)
 
-    def create_buffer_around_position(self: OMPLPath, position: cs.XY) -> Polygon:
-        """Create a space around the given position. Position is the center of the space and
-        is a tuple of x and y.
-        """
-        space = Point(position.x, position.y).buffer(self._box_buffer, cap_style=3, join_style=2)
-        return space
-
     def update_objectives(self):
         """Update the objectives on the basis of which the path is optimized.
         Raises:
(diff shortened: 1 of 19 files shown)


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -4,10 +4,16 @@
...
 from rclpy.impl.rcutils_logger import RcutilsLogger
-from shapely.geometry import MultiPolygon
+from shapely.geometry import MultiPolygon, LineString
+from pyproj import Geod
 
 import local_pathfinding.obstacles as ob
 from local_pathfinding.ompl_path import OMPLPath
+import local_pathfinding.coord_systems as cs
+
+LOW_WIND_SPEED_THRESHOLD = 9.26  # 5 knots
```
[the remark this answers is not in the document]
raghumanimehta: -	Code Modification: Removed unnecessary logic for better performance.
-	Justification: The constant used in the check was too low and the logic was redundant as the program already chooses the least cost path.

src/website/tests/simulation/linear_interpolation.py
```diff
@@ -1,11 +1,9 @@
-"""Interpolates points on a line between two waypoints. Used to create a series of GPS points based on the local path."""
+"""Interpolates points on a line between two waypoints. Used to create a series of GPS points
```
SPDonaghy: I'm assuming this is just the linter, was a workflow failing without this?

src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -64,6 +64,10 @@ def bound_to_180(angle_degrees: float) -> float:
     return ((angle_degrees + 180) % 360) - 180
 
 
+def calculate_heading_diff(boat_heading: float, desired_heading):
+    return abs((desired_heading - boat_heading + 180) % 360 - 180)
```
SPDonaghy: I think just call `bound_to_180` here right?

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -117,12 +209,84 @@ def update_if_needed(
...
+        new_cost_normalized = new_cost / max_cost
+
+        max_heading_diff = max(
+            math.fabs(heading_diff_new_path), math.fabs(heading_diff_old_path), 1.0
+        )
+
+        heading_diff_new_normalized = heading_diff_new_path / max_heading_diff
+        heading_diff_old_normalized = heading_diff_old_path / max_heading_diff
+
+        w_h = 0.6
```
SPDonaghy: Could you define these constants at the very top of the file with more descriptive upper snake case names instead?

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -85,28 +91,114 @@ def __init__(self, parent_logger: RcutilsLogger):
...
-            ais_ships (ci.AISShips): AIS ships data.
-            global_path (ci.Path): Path to the destination.
-            filtered_wind_sensor (ci.WindSensor): Wind data.
+        Evaluates whether to update the current path based on several criteria:
+        - Receipt of a new global waypoint
+        - Absence of an existing path
+        - Current path intersecting with collision zones
+        - New path having a better cost-heading metric than the old path
+
+        The decision metric combines normalized heading difference (60% weight) and
```
SPDonaghy: I would suggest not specifying the actual values of the weights here and instead just mentioning that the heading diffs and path cost are weighted separately. Just so its marginally easier to maintain if we want to change the weights later. See my other comment below about moving the definition of the weights to top of file.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -165,13 +165,41 @@ def init_obstacles(
...
+
+        Returns:
+            float: The total cost of the path from the given waypoint index. Returns `float('inf')`
+                   if the solution path does not exist.
         """
-        raise NotImplementedError
+        try:
+            solution_path = self._simple_setup.getSolutionPath()
+        except Exception as e:
+            self._logger.debug(f"Solution path does not exist. Exception thrown: {e}")
```
SPDonaghy: Nit: maybe warn here? Is this the thing that would just happen on startup when there is no old path?

PR #644, opened 2025-10-04
Added state space to the visualizer

AMaharaj16:


raghumanimehta (changes requested):
use the variables goal_x, goal_y etc. on lines 532 to 535

AMaharaj16:
Since I didn't write a description before here it is:

- In visualizer, used create_buffer_around_position to generate boat state space in live_update_plot.

- Since create_buffer_around_position was now being used by both ompl_path and visualizer, I refactored it out of ompl_path and into ompl_objectives. Both ompl_path and visualizer now import it from ompl_objectives as a helper function.

- Moved test_create_space from test_ompl_path to test_ompl_objectives.

- Added comments with values for further testing.

src/local_pathfinding/local_pathfinding/ompl_objectives.py
```diff
@@ -32,6 +33,16 @@
 WINDSPEEDS = [0, 9.3, 18.5, 27.8, 37.0]  # The row labels
 ANGLES = [0, 20, 30, 45, 90, 135, 180]  # The column labels
 
+BOX_BUFFER_SIZE = 1.0  # km
```
raghumanimehta: For reusabiity (in case we need it), don't remove self._box_buffer from OMPLPath class and don't set the BOX_BUFFER_SIZE as a constant in ompl_objectives. It would be better if we keep the BOX_BUFFER_SIZE in the OMPL module and change the signature of create_buffer_arounds_space from (position: cs.XY) to (position: cs.XY, box_buffer_size: float).  So visualizer, can pass in the constant referred from it's module and ompl can pass the stored attribute. This way, if we need to ever reuse this function, we could just call it with the required buffer size which might or might not be BOX_BUFFER_SIZE

PR #660, opened 2025-10-11
User/adambrett40/566 power off

### LINKED ISSUE #566: CAN power off frame
### Purpose

Add power off frame from CAN documentation

### Resources

-

adambrett40:
### Description

- Resolves #566 

- Added the power off frame from the can frames documentation, can command to shutdown gracefully when power cutoff is imminent.
- Use ssh over localhost to access host's shutdown command (this requires some setup for the host which is documented in the script). Also mounts in an ssh key from the host to avoid git tracking an ssh key (might be better to put in a docker volume to avoid hardcoding the path?).

### Verification

- Works to send shutdown command to rpi using CAN loopback, including after rebuilding the container.

### Resources

-


SPDonaghy (approved):
I skimmed over the cpp files and just mainly wanted to see the compose file and bash script. Looks solid to me! 

Could this issue perhaps have a sister issue for auto launch (with cron or something) when the pi is powered back on?

Jng468 (approved):

Issue #575, opened 2025-06-01
Improving Compilation time and Reduce Resource Utilization


FireBoyAJ24:
### Resources

-

PR #667, opened 2025-10-18
Create release image with ROS binaries pre-built for deployment on RPI

digcheese:
### Description

- Resolves first task of #575

- Created a release docker image layer that builds off of the base-dev image and runs setup.sh and build.sh
- Created deployment script and associated VSCode workspace task to build the docker release image.
- Created GitHub workflow to also build the release image

### Verification

- Built and pushed a release image with vscode task.
- Pulled the release image on the raspberry pi and verified that binaries were available without building.

### DIFF
--- a/scripts/deployment/start_containers.sh
+++ b/scripts/deployment/start_containers.sh
@@ -117,8 +117,7 @@ MONGO_TAG=$MONGO_TAG $DOCKER_COMPOSE_ARGS up --build --detach --pull never
 if [[ "$INTERACTIVE" = true ]]; then
     $DOCKER_COMPOSE_ARGS exec --interactive --tty sailbot-workspace /bin/bash
 else
-    $DOCKER_COMPOSE_ARGS exec --no-TTY sailbot-workspace /bin/bash -c \
-        "source ./install/local_setup.bash && ros2 launch src/global_launch/main_launch.py"
+    $DOCKER_COMPOSE_ARGS exec --no-TTY sailbot-workspace /bin/bash -c "./scripts/run_software.sh"
 fi
 
 # stop containers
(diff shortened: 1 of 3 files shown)


.devcontainer/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ghcr.io/ubcsailbot/sailbot_workspace/dev:deferred-path-variable-evaluation
+FROM ghcr.io/ubcsailbot/sailbot_workspace/dev:with-binaries
```
SPDonaghy: I think we want to undo this change because the dev image should be unchanged right?

scripts/deployment/start_containers.sh
```diff
@@ -117,7 +117,8 @@ MONGO_TAG=$MONGO_TAG $DOCKER_COMPOSE_ARGS up --build --detach --pull never
 if [[ "$INTERACTIVE" = true ]]; then
     $DOCKER_COMPOSE_ARGS exec --interactive --tty sailbot-workspace /bin/bash
 else
-    $DOCKER_COMPOSE_ARGS exec --no-TTY sailbot-workspace /bin/bash -c "./scripts/run_software.sh"
+    $DOCKER_COMPOSE_ARGS exec --no-TTY sailbot-workspace /bin/bash -c \
```
SPDonaghy: Can we leave dealing with the deployment scripts for another PR? or is this change required?

SPDonaghy: Got it, in that case we can remove it from this PR. We can overhaul those deployment scripts at some point if we need to

SPDonaghy (approved):

Issue #271, opened 2023-11-24
Sensor Filters


hhenry01:
### Purpose

Raw sensor data is not perfect. Filtering can help reduce outliers and improve reliability.

### Description

Currently, the only filter that we want for certain is a moving average of the wind sensors. There are multiple different moving averages, so we should have multiple implementations. Additionally, we want the filters to work on _any_ numerical data, not just wind sensors, for the sake of future extensibility.

The requirements for this closing this issue are:
- Add a new `filters` library under `lib/` (since it should operate any numerical data, not just specific sensors)
  - Add headers, implementation, and tests
  - Add to CMake (Make sure it can run in CAN Transceiver)
- Create a templated base class `Filter` that looks something like:
```
#include <concepts>

template<typename T>
concept arithmetic = std::integral<T> or std::floating_point<T>;
// A concept restricts the acceptable template types
template<arithmetic T> 
class Filter {
public:
    Filter(T initial_value) : stored_value_(initial_value) {}
    virtual void update(T new_value) = 0;
    T value() const {
        return stored_value_;
    };
protected:
    T stored_value_;
}
```
- Derived classes for the following filters:
    - simple moving average
    - cumulative moving average
    - weighted moving average
    - exponential weighted moving average
ex.
```
template<arithmetic T>
class MovingAvg : public Filter<T> {
// implementation
}
```
- An nice-to-have would be to make the filters operate on containers (vectors, arrays, etc...) as well
```
template<template <typename... > class Container, typename... T>
class MovingAvg : public MovingAvg<T> {
// implementation
}
```

### Resources

- Wikipedia page on different moving averages: https://en.wikipedia.org/wiki/Moving_average
- Info on the C++ template syntax I used (good chance it won't compile as is):
  - Using `concept` to restrict a template to a numerical value: https://stackoverflow.com/a/14294277
  - Using variadic template arguments to specify container templates: https://stackoverflow.com/a/44064626

PR #682, opened 2025-10-30
User/adambrett40/605 filtered wind sensor

### LINKED ISSUE #605: Verify/modify filtered wind sensor
### Description

Pathfinding wants only filtered wind sensor to be published - meaning filter out erroneous data and only give the average. We need to filter out bad data   and only send valid. So in the event of 1/2 sensors sends bad data, send just the other sensor.

adambrett40:
### Description

- Resolves #605, starts on #271 

- implements filtered wind sensor as simple moving average of the last n (currently NUM_WIND_SENSORS = 2) wind sensor readings
- adds elec -> soft convention conversion, since ELEC uses [0, 359] for wind direction while PATH uses [-179, 180]
- filter should be extracted to its own library along with other possible filters for other sensors as mentioned in #271 

### Verification

- tested on rpi with loopback that directions get converted according to the conventions that PATH expects
- tested that averages across 180/360 boundaries work correctly 

### Resources

-

### DIFF
--- a/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
+++ b/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
@@ -368,12 +368,6 @@ private:
             wind_sensor_msg                   = wind_sensor.toRosMsg();
             wind_sensors_pub_->publish(wind_sensors_);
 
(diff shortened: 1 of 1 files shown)


Jng468 (approved):
Very cool

PR #687, opened 2025-11-01
User/rnar17/683 update wind objectives

### LINKED ISSUE #683: Investigate and update WindObjectives
### Purpose

- I noticed that in `ompl_objectives.py` we are multiplying disance to the next waypoint with upwind and downwind multiplier. This seems to add a bias since we are already handling distance metrics in `DistanceObjective`; ideally each objective class should handle only one objective, if that's not the case an objective might unintentionally get extra weight.  
- Investigate if my intuition is right and remove the multipliers if it is. 

### Changes

- Changes to class WindObjective in https://github.com/UBCSailbot/sailbot_workspace/blob/7374e7870abba5536ccee18ac2c40e8cc958ab88/src/local_pathfinding/local_pathfinding/ompl_objectives.py#L230

### Resources

- https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/ompl_objectives.py#L1

rnar17:
### Description

- Resolves #683 

This PR changes the wind cost calculation from a binary system (upwind/downwind zones have fixed costs) to a continuous cost function that smoothly scales based on how aligned the boat is with the wind. It takes the relative angle between the boat's heading and the wind direction, normalizes to [-pi, pi), then takes the cosine of that angle. The cost is computed by multiplying the `UPWIND_MULTIPLIER` or `DOWNWIND_MULTIPLIER` (depending on whether the boat is facing upwind or downwind) by the cosine of the angle. 

Set UPWIND_MULTIPLIER to 1.0 and DOWNWIND_MULTIPLIER to 0.75 (i.e. we prefer downwind but not by much due to potential damage)

- cos(angle) = 1 when heading directly into wind → apply full upwind cost
- cos(angle) = 0 when perpendicular to wind → zero cost (optimal)
- cos(angle) = -1 when heading with wind → apply full downwind cost

### Verification

- Various unit tests for `wind_direction_cost()` containing various boat and wind directions contained in `test_ompl_objectives.py`

### DIFF
--- a/src/local_pathfinding/test/test_coord_systems.py
+++ b/src/local_pathfinding/test/test_coord_systems.py
@@ -78,21 +78,6 @@ def test_bound_to_180(unbounded: float, bounded: float):
 
 
 @pytest.mark.parametrize(
-    "unbounded,bounded",
-    [
-        (0.0, 0.0),
-        (-math.pi, -math.pi),
-        (math.pi - 0.01, math.pi - 0.01),
-        (math.pi, -math.pi),
-        (-math.pi - 0.01, math.pi - 0.01),
-        (10 * math.pi + 0.001, 0.001),
-    ],
-)
-def test_bound_to_pi(unbounded: float, bounded: float):
-    assert cs.bound_to_pi(unbounded) == pytest.approx(bounded), "incorrect angle conversion"
-
-
-@pytest.mark.parametrize(
     "ref_lat,ref_lon,true_bearing_deg,dist_km",
     [
         (30.0, -123.0, 0.00, 30.0),


src/local_pathfinding/test/test_ompl_objectives.py
```diff
@@ -9,11 +9,14 @@
...
+from local_pathfinding.ompl_objectives import (
+    create_buffer_around_position,
+    get_true_wind,
+)
 
 # Upwind downwind cost multipliers
-UPWIND_MULTIPLIER = 3000.0
-DOWNWIND_MULTIPLIER = 3000.0
+UPWIND_MULTIPLIER = 1.0
+DOWNWIND_MULTIPLIER = 0.75
```
raghumanimehta: ~Are these constants arbitrary? or did you tune them?~ Can you please import these constants from the `ompl_objectives`?

src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -64,6 +64,10 @@ def bound_to_180(angle_degrees: float) -> float:
     return ((angle_degrees + 180) % 360) - 180
 
 
+def bound_to_pi(angle_rad: float) -> float:
```
raghumanimehta: Can you add documentation for this function please?

raghumanimehta: I didn't see any tests for this. Can you add them? Please let me know, if i missed them.

raghumanimehta (changes requested):
Please update `bound_to_pi`. Thanks

src/local_pathfinding/test/test_ompl_objectives.py
raghumanimehta: I meant explicitly using `from local_pathfinding.ompl_objectives import UPWIND_MULTIPLIER, DOWNWIND_MULTIPLIER`

src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -75,6 +75,19 @@ def bound_to_180(angle_degrees: float) -> float:
...
+    """
+    Bounds an angle to the interval (-pi, pi].
+
+    Args:
+        angle_rad (float): The input angle in radians to be bounded
+
+    Returns:
+        float: The equivalent angle bounded to (-pi, pi] in radians
+    """
+    return ((angle_rad + math.pi) % (2 * math.pi)) - math.pi
```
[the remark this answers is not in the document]
rnar17: bound_to_pi returns (-pi, pi] but we can use bound_to_180 too. Either is fine.

PR #691, opened 2025-11-05
Raghumanimehta/updating mocks 678

raghumanimehta:
### Description

- Resolves a lot of things.

- Changed the mock wind sensor to have the following: 
- - constant mode that outputs constant value (through ros2 param set)
   - Added the feature that winds dont change for 60 to 120 seconds (randomly decided in the runtime) 
   - we can now control the distribution of the speeds and direciton
   - Earlier the mock_wind_sensor was outputting true winds. Now the use can specify true winds (for simplicity) and the mock node would convert it to apparent wind, by using mock gps' published data
- Refactored the mock nodes, such that they exist in a separate directory in local_pathfinding module
- Due to the growing number of wind conversion helpers, created a new module `wind_coord_systems.py` which has these helpers. 
- Fixed a calculation in WindObjectives where the code assumed in get_true_wind function that the inputs are wrt global true bearing. This wasn't the case. This has now been fixed by first fixing the coordinate system and then passing this as input to get_true_wind.  
 
 

### Verification

- Tests for wind_coord_systems.py
- Manual verification of mock nodes.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/global_path.py
+++ b/src/local_pathfinding/local_pathfinding/global_path.py
@@ -218,7 +218,7 @@ def generate_path(
 ) -> ci.Path:
     """Returns a path from the current GPS location to the destination point.
     Waypoints are evenly spaced along the path according to the interval_spacing parameter.
-    Path does not include pos, but does include dest as the final element.
+    Path does not include pos, but does include dest as the last element.
 
     If write is True, the path is written to a new csv file in the same directory as file_path,
     with the name of the original file, appended with a timestamp.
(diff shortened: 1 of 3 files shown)


src/website/views/components/TimestampFilter/timeStampFilter.module.css
SPDonaghy: Can you remove this from the PR please?

fyang151: just wondering why we are removing this file?

raghumanimehta: That was not my intention. I don't know how this happened, lol.

src/local_pathfinding/local_pathfinding/mock_nodes/shared_constants.py
```diff
@@ -0,0 +1,11 @@
+import custom_interfaces.msg as ci
+
+MEAN_SPEED = ci.HelperSpeed(speed=15.0)  # mean boat speed in kmph
+START_POINT = ci.HelperLatLon(latitude=48.46, longitude=-125.1)  # Starting location of the mock
+START_HEADING = ci.HelperHeading(heading=180.0)  # in degrees, heading of the boat
+
+'''
+For further tests:
+START_POINT = ci.HelperLatLon(latitude=49.308157, longitude=-123.244801)
+Change last line of mock_global_path.csv to: 49.289686,-123.195877
```
SPDonaghy: What's this for?

PR #733, opened 2025-12-07
Raghumanimehta/mock start up fix

raghumanimehta:
Re-implemented the `node_mock_wind_sensor` from scratch. The `constant` model works fine. The program might crash because of the old implementation of `get_sailbot_speed`; this is not the fault of the new method. 

I believe we can merge this for any sort of testing in the meanwhile you'd like to do. 
I will be working on getting the variable model and a planned voyage mode ready after the exams.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
+++ b/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
@@ -5,8 +5,6 @@ This node publishes mock data that mimics the boat's wind sensors.
 It takes **true wind** inputs in the **global frame** via ROS parameters and publishes
 **apparent wind** outputs in the **boat frame** on ``filtered_wind_sensor``.
 
-Important design notes:
-
 Node name:
 * ``mock_wind_sensor``
 


src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
```diff
@@ -93,125 +125,93 @@ def __init__(self):
...
-
         self.__gps_sub = self.create_subscription(
             msg_type=ci.GPS, topic="gps", callback=self.gps_callback, qos_profile=10
         )
-
-        self.__last_direction = int(
-            self.get_parameter("mean_direction").get_parameter_value().double_value
+        self.__boat_heading = sc.START_HEADING.heading
+        self.__boat_speed = sc.MEAN_SPEED.speed
+        self.__mode = "constant"
```
SPDonaghy: nit: should we derive this from the parameter instead?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
```diff
@@ -93,125 +125,93 @@ def __init__(self):
...
+        msg = ci.WindSensor(
+            speed=ci.HelperSpeed(speed=aw_speed_kmph),
+            direction=aw_direction_boat_coord_deg
         )
-        self.__last_direction = result
-        return result
+        self.get_logger().debug(f"Publishing to {self.__wind_sensors_pub.topic}: {msg}")
+        self.__wind_sensors_pub.publish(msg)
 
     def get_latest_speed_and_direction_values(self) -> None:
```
SPDonaghy: I know the name of this fn was not changed in this PR but maybe we can change it to something like `update_params` ? since a more general name makes the function more easily extensible to update other parameters (like mode)

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
```diff
@@ -21,70 +21,102 @@
...
-    """Publishes mock apparent wind speed and direction at a fixed rate.
+class WindModel(ABC):
+    @abstractmethod
+    def get_wind_sensor_data(self):
+        pass
+
+
+class ConstantWindModel(WindModel):
+
+    def __init__(self, true_wind_heading_deg, true_wind_speed_kmph,
```
SPDonaghy: This could probably be defined in WindModel right?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
```diff
@@ -21,70 +21,102 @@
...
+                 boat_heading, boat_speed):
+
+        self.__true_wind_heading_deg = true_wind_heading_deg
```
SPDonaghy: Are you intending to use name mangling by adding a double underscore to the attribute names? I'm wondering why not use just a single underscore?

SPDonaghy (commented):
Left a few questions. Also just to confirm, you have tested this works correctly?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
raghumanimehta: No reason, the old wind sensor code had double underscore and i copied parts of it and didn't change. I prefer single underscores, so i can change that

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
raghumanimehta: Yes. will make that change

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
raghumanimehta: sure

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
```diff
@@ -21,70 +21,102 @@
...
+        self.__true_wind_speed_kmph = true_wind_speed_kmph
+        self.__boat_heading = boat_heading
```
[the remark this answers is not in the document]
raghumanimehta: yes

SPDonaghy (approved):
Thanks Raghumani, have you tested this to ensure everything works as desired?

PR #743, opened 2025-12-30
AMaharaj16/test-conditions-switch-path-669

AMaharaj16:
Refactored the `update_if_needed` function to improve testability by extracting the metric calculation logic (which combines heading difference cost and path cost) into a separate, testable unit called `calculate_metric`. Test suite for `calculate_metric` fully passes. No actual computation changes made.

Also moved `test_calculate_desired_heading_and_waypoint_index` from `test_node_navigate.py` to `test_local_path.py` and added testing for `calculate_heading_diff`.

Next, I plan to test `update_if_needed` altogether using a deterministic function that creates OMPLPath mock objects with predetermined paths to enable controlled testing.

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -7,11 +7,6 @@
     "indent": 4
   },
   "MD010": false,
-  "MD013": {
-    "line_length": 120,
-    "code_blocks": false,
-    "tables": false
-  },
   "MD033": false,
   "MD035": {
     "style": "---"


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -239,30 +246,17 @@ def update_if_needed(
...
 
-        heading_diff_new_normalized = heading_diff_new_path / max_heading_diff
-        heading_diff_old_normalized = heading_diff_old_path / max_heading_diff
-
-        w_h = HEADING_WEIGHT
-        w_c = COST_WEIGHT
-
-        metric_old = w_h * heading_diff_old_normalized + w_c * old_cost_normalized
-        metric_new = w_h * heading_diff_new_normalized + w_c * new_cost_normalized
+        metric_old, metric_new = self.calculate_metric(heading, heading_old_path, heading_new_path,
```
raghumanimehta: Instead of returning a tuple of old and new, you can make function reusable such that it doesn't care if you pass the old or new path. You can call compute metric on the old_path and new_path separately to get old and new metric respectively. 
For example:
```
new_metric = calculate_metric(heading_new_normalized, heading_old_normalized)
```
This function right now is very tightly coupled with the caller. Also, I don't like the assymetric behaviour where we are normalizing the cost outside the function but normalizing the heading inside.

raghumanimehta (changes requested):
Nice work on the PR! Requesting a bunch of refactoring the code before I merge this to the main branch. As soon as you are done with the changes, let me know and move on to the next half of testing.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -16,6 +16,12 @@
 COST_WEIGHT = 0.4
 
 
+def normalize(x: float, y: float):
```
SPDonaghy: Please use a more descriptive function name and docstring to explain what this function is for and exactly what it's doing.

Keep in mind, an item can be normalized to lots of different conditions so just saying x and y are normalized is not a complete description.

An example would be:
```
def normalize_sigmoid(x: int):
    """Normalizes x to a value in the range [0, 1] using the sigmoid function.
    """
    return 1 / (1 + math.exp(-x))
```

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -239,50 +246,56 @@ def update_if_needed(
...
-        max_heading_diff = max(
-            math.fabs(heading_diff_new_path), math.fabs(heading_diff_old_path), 1.0
-        )
-
-        heading_diff_new_normalized = heading_diff_new_path / max_heading_diff
-        heading_diff_old_normalized = heading_diff_old_path / max_heading_diff
-
-        w_h = HEADING_WEIGHT
-        w_c = COST_WEIGHT
+        heading = self.state.heading if self.state else 0.0
```
SPDonaghy: When would self.state be None?

AMaharaj16: Not entirely sure, I just remember it was causing errors in testing so I adjusted it. In my next PR I will give testing `update_if_needed` another go and this may change again.

raghumanimehta:
@AMaharaj16 I refactored the function `update_if_needed` to make it more readable. I renamed the function `calculate_metric` and changed how ti computed the metric. 
Please address the remaining comments and "Resolve" them once you have addressed them. 
Then, please provide a detailed description of what you have tested already, what you learned and what you would be testing next.

raghumanimehta (dismissed):
The task is yet not complete. Merging because, there are notable improvements in `update_if_needed`

PR #745, opened 2026-01-07
Refactored visualizer and add tests

### LINKED ISSUE #738: Visualizer refactor
### Purpose

Visualizer class has a lot of technical debt.

### Changes

Refactoring the `visualizer.py` in `local_pathfiding` in the following ways:
- Ensure proper constants, functions, and variable naming. Reference: https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/2736685057/Style+guide+for+functions+variables+and+constants
- Split long functions to smaller atomic functions to increase readibility and usability. 
- Fix all the bugs along the way. Ensure the code for visualizer is as expected by writing the unit tests for helper functions as and when required and then verifying that the visualizer is displaying the right information (i.e. matches the tests).

jb3982:
### Description

- Resolves #738
This PR refactors the **local pathfinding visualizer** to improve readability, maintainability, and testability, while preserving existing visual behaviour and user interactions.

#### Key changes
- Refactored `visualizer.py` into smaller, well-scoped helper functions.
- Broke the previous monolithic update logic into composable builders (paths, goal, boat, wind box, state space, etc.).
- Introduced structured helpers for:
  - Goal change detection
  - Unit vector normalization
  - Path, goal, and intermediate waypoint traces
- Added unit tests covering the newly introduced helper functions.

### Visualization & UX Improvements
- **Preserves user zoom and pan** across updates by leveraging `uirevision="constant"` and setting axis ranges only on the initial render.
- On first render, the view automatically fits the **state-space bounds** (boat + goal buffers).
- Subsequent updates no longer reset the view, preventing the “zoom snapping back” behaviour seen previously.

**Attached images**
- **Image 1**: Initial render - axes automatically fit the state-space bounds.
- **Image 2**: After user zoom/pan  view is preserved across updates.

---

### Verification

- Ran the `local_pathfinding` package and visually verified:
  - Path rendering
  - Goal updates
  - Wind vectors (apparent, true, and boat wind)
  - AIS and obstacle overlays
  - Zoom/pan persistence across updates
- Executed the newly added unit tests for visualizer helper functions.

## Image 1:
Image 1

## Image 2:
Image 2


src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,64 +15,110 @@
...
 import local_pathfinding.wind_coord_systems as wcs
 from local_pathfinding.ompl_path import OMPLPath
 
-app = dash.Dash(__name__)
+# -----------------------------
+# Constants
+# -----------------------------
+UPDATE_INTERVAL_MS = 2500
+
+BOX_BUFFER_SIZE_KM = 1.0  # km
```
raghumanimehta: Remove the coment please.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,64 +15,110 @@
...
-        final_local_wp_y (List[Tuple[float, float]]): Y coordinates of the final local waypoint
+        sailbot_pos_x (List[float]): X coordinates (km) of the boat track.
+        sailbot_pos_y (List[float]): Y coordinates (km) of the boat track.
+
+        final_local_wp_x (List[float]): X coordinates (km) of the latest local path waypoints.
+        final_local_wp_y (List[float]): Y coordinates (km) of the latest local path waypoints.
+
+        sailbot_gps (List[ci.Gps]): GPS messages used for heading and speed display.
+
+        ais_pos_x (List[float]): X coordinates (km) of AIS ships.
```
raghumanimehta: add units in the definitions here.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -520,139 +635,251 @@ def live_update_plot(state: VisualizerState) -> go.Figure:
...
+) -> Tuple[go.Figure, Tuple[float, float]]:
+    fig = initial_plot()
+
+    local_x = list(state.final_local_wp_x)
+    local_y = list(state.final_local_wp_y)
+
+    # Boat and goal info
+    boat_xy = (state.sailbot_pos_x[-1], state.sailbot_pos_y[-1])
+
+    # DEBUG: inspect local path structure
```
raghumanimehta: Remove these comments.

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: This was marked as resolved but hasn't been resolved. I meant something like ais_pos_x_km should be the variable name instead of ais_pos_x

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -81,246 +127,279 @@ def __init__(self, msgs: deque[ci.LPathData]):
...
-        """Checks if the sailbot observer node received any messages.
+    @staticmethod
+    def _validate_message(msg: ci.LPathData) -> None:
+        """
+        Checks if the sailbot observer node received any messages.
         If not, it raises a ValueError.
+        Args:
+            msg: Latest `ci.LPathData` message to validate.
+        """
+        if msg.global_path is None or len(msg.global_path.waypoints) == 0:
```
raghumanimehta: What was the reason for changing this code here from the original implementation? I don't see why an error should be raised if the list is empty.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -81,246 +127,279 @@ def __init__(self, msgs: deque[ci.LPathData]):
...
+    Args:
+        last_goal_xy: Previously stored (x, y) goal in km (already rounded), or None if rendered
+                      for the first time.
+        goal_xy: Current (x, y) goal in km.
+
+    Returns:
+        GoalChange containing the rounded goal coordinates and an optional popup message.
     """
-    Initializes the plot with default settings.
+    rounded = (
```
raghumanimehta: What happens if you don't round?

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: But why do you think an empty list is a case in which we should raise an error? What makes an empty list wrong? We can be at the goal where the list is empty? It's a rare case, but this shouldn't raise an error. We should raise an error only if we know something has horribly gone wrong because errors make the system panick.

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: You should push the changes if you mark them resolved. I will let you know of more changes if I find any in the final PR after the chagnes. 
The workflow should be `make PR -> review the review and fix -> push the fixes -> repeat till the PR is deemed fit`

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -107,8 +107,7 @@ def get_next_location(self) -> None:
             self.__current_location.latitude,
             self.__current_location.longitude,
         )
-        heading_radians = math.radians(self.__heading_deg.heading)
-        destination = great_circle(kilometers=distance_km).destination(start, heading_radians)
+        destination = great_circle(kilometers=distance_km).destination(start, self.__heading_deg.heading) # noqa
```
SPDonaghy: please actually remove this change from this PR, its resolved here with clearer variable name `heading_degrees`

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
+        ais_pos_x_km (List[float]): X coordinates (km) of AIS ships.
+        ais_pos_y_km (List[float]): Y coordinates (km) of AIS ships.
+        ais_headings_deg (List[float]): Headings (degrees) of AIS ships.
+        ais_ship_ids (List[int]): AIS ship identifiers.
+        ais_speeds_kmph (List[float]): Speeds (km/h) of AIS ships.
+
+        land_obstacles_xy (List[Polygon]): Land obstacle polygons in XY (km).
+        boat_obstacles_xy (List[Polygon]): Boat collision-zone polygons in XY (km).
+
+        aw_wind_vector (cs.XY): Apparent wind vector in global XY.
```
SPDonaghy: These variable names should have units in them (kmph if they are in km/h)

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
+        sailbot_pos_x_km (List[float]): X coordinates (km) of the boat track.
+        sailbot_pos_y_km (List[float]): Y coordinates (km) of the boat track.
+
+        final_local_wp_x_km (List[float]): X coordinates (km) of the latest local path waypoints.
+        final_local_wp_y_km (List[float]): Y coordinates (km) of the latest local path waypoints.
+
+        sailbot_gps (List[ci.Gps]): GPS messages used for heading (degrees) and speed (km/h)
+                                    display.
+
+        ais_pos_x_km (List[float]): X coordinates (km) of AIS ships.
```
SPDonaghy: Can we store all the ais data in a dictionary with ship ids as keys instead of having data for each ship distributed over multiple lists?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
-        aw_speed = self.curr_msg.filtered_wind_sensor.speed.speed
-        aw_dir_boat = self.curr_msg.filtered_wind_sensor.direction
+        # Wind Vectors
+        boat_speed = self.latest_msg.gps.speed.speed
+        boat_heading = self.latest_msg.gps.heading.heading
+        aw_speed = self.latest_msg.filtered_wind_sensor.speed.speed
+        aw_dir_boat = self.latest_msg.filtered_wind_sensor.direction
 
         # Convert Apparent wind to global frame
         aw_dir_global = wcs.boat_to_global_coordinate(boat_heading, aw_dir_boat)
```
SPDonaghy: please add units to this var name while we're at it

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
         aw_dir_global_rad = math.radians(aw_dir_global)
-
         # Compute apparent wind vector (in global frame)
         self.aw_wind_vector = cs.polar_to_cartesian(aw_dir_global_rad, aw_speed)
```
SPDonaghy: units

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
         """
-        processed_obstacles = []
+        processed: List[Polygon] = []
         if obstacles is None:
-            return processed_obstacles
+            return processed
 
-        for obs in obstacles:
-            if obs.obstacle_type != "Boat":
+        for ob in obstacles:
```
SPDonaghy: Can you use something other than `ob` since we use the import:
`import local_pathfinding.obstacles as ob` in other files.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
 
-def initial_plot() -> go.Figure:
+def compute_goal_change(
+    last_goal_xy: Optional[Tuple[float, float]], goal_xy: Tuple[float, float]
+) -> GoalChange:
+    """
+    Determine whether the local goal moved since the last update (with some jitter tolerance).
+
+    Args:
+        last_goal_xy: Previously stored (x, y) goal in km (already rounded), or None if rendered
```
SPDonaghy: Nit: This should work regardless of the units right? ie (x, y) don't need to be in km? If so, the docstring over constrains the function usage by stating (x, y) is in km. Just something to think about from a maintainability standpoint.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
-    goal_trace = go.Scatter(
-        x=goal_x,
-        y=goal_y,
+
+def build_goal_trace(goal_xy: Tuple[float, float], angle_deg: float) -> go.Scatter:
+    """
+    Create the marker trace for the local goal waypoint.
+
+    Args:
+        goal_xy: (x, y) goal position in km.
```
SPDonaghy: units

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
-        x=[state.sailbot_pos_x[-1]],
-        y=[state.sailbot_pos_y[-1]],
+
+def build_path_trace(local_x: List[float], local_y: List[float],
+                     boat_xy: Tuple[float, float]) -> Optional[go.Scatter]:
+    """
+    Create a dotted line trace connecting the local waypoints to the goal.
+
+    Args:
+        local_x: Local path X coordinates in km.
```
SPDonaghy: units

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
+
+def build_boat_trace(
+    state: VisualizerState, boat_xy: Tuple[float, float], dist_to_goal_km: float
+) -> go.Scatter:
+    """
+    Create the boat marker trace (filled arrow-head/ triangle) at the current boat position.
+
+    Args:
+        state: VisualizerState containing GPS heading/speed history.
+        boat_xy: (x, y) current boat position in km.
```
SPDonaghy: units

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -329,150 +409,191 @@ def live_update_plot(state: VisualizerState) -> go.Figure:
...
-                    y=y,
-                    fill="toself",
-                    mode="lines",
-                    line=dict(color="lightgreen"),
-                    fillcolor="lightgreen",
-                    opacity=0.5,
-                    name="Land Obstacle",
-                )
+
+def add_polygon(
```
SPDonaghy: I really like that you have broken out tracing logic for each type of object into a separate function that generates a go.scatter that can be added to the main figure, but I am curious as to why you discontinue that pattern starting from here and chose to make some functions add their trace directly to the figure. Wouldn't it be nice if every function returned a trace and then in the main plotting function we can just cleanly call add trace a bunch of times?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -14,313 +15,392 @@
...
+        self.true_wind_vector = cs.polar_to_cartesian(true_wind_angle_rad, true_wind_mag)
+
         # Boat wind vector
         boat_wind_radians = math.radians(cs.bound_to_180(boat_heading + 180))
         self.boat_wind_vector = cs.polar_to_cartesian(boat_wind_radians, boat_speed)
 
-    def _validate_message(self, msg: ci.LPathData):
-        """Checks if the sailbot observer node received any messages.
+    @staticmethod
+    def _validate_message(msg: ci.LPathData) -> None:
```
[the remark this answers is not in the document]
raghumanimehta: I believe this check is fine. I don't see how we might get `LPathData` with incorrect fields because ROS is strictly typed.

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: You can make a note that intended use involves x, y in the units of km.

src/local_pathfinding/local_pathfinding/visualizer.py
jb3982: Thanks for the feedback. Keeping consistency is a fair point but my reasoning for the mixed design choice was:
1. For add_polygon: We're iterating over a collection of polygons and handling fig.add_trace() inside the function kept the iteration logic cohesive and avoided adding redundant looping in build_figure or else I feared we would end up with the same clutter like the old visualizer's live_update_plot function.

2. For add_wind_box and compute_and_add_state_space: I treated them more like a tightly-integrated system like wind box required coordinated setup (setting up the axes, vector arrow calculations, annotations, labeling), and state space computes geometry and adds the overlay as a single visual unit so keeping it in a single function call looked much better to me rather than splitting in between build_figure and a helper function.

The goal was to minimize clutter in build_figure and keep closely-related elements together, rather than fragmenting them across the script.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -1,9 +1,10 @@
 """
 Sailbot Path Planning Visualizer
 
-This script sets up a Dash web application to visualize the sailbot's path planning data.
-It processes ROS messages such as waypoint coordinates and GPS data, converting them from
-latitude/longitude to XY coordinates in km.
+This script provides visualization tools for the sailbot path planning module. This script sets up
```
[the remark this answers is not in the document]
jb3982: okay

src/local_pathfinding/local_pathfinding/visualizer.py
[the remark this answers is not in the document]
raghumanimehta: @SPDonaghy, do you have any more thoughts about this?

raghumanimehta (approved):

raghumanimehta:
@SPDonaghy, please review the PR and see if your issues have been resolved. I feel this PR is now ready to be merged.

PR #748, opened 2026-01-10
Raghumanimehta/circular dependency fix for mock nodes

raghumanimehta:
### Description

- Removed the cyclical dependency between the `mock_wind_sensor` and the `mock_gps` by giving direct access of true wind to `mock_gps` 
- To ensure that the updates are synchronous, added a shell script to change parameters rather than using `ros2 param set` command. 
- Fixes the forward drift of the boat on the visualizer.

### Verification

- Manual verification that the nodes output the right vlaue and the values are changed synchronously.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
+++ b/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
@@ -113,7 +113,7 @@ class MockGPS(Node):
         """
         self.__mean_speed_kmph = ci.HelperSpeed(
             speed=float(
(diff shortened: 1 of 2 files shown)


src/global_launch/config/globals.yaml
```diff
@@ -89,4 +89,3 @@ mock_data_node:
       mock_sail_trim_tab: True
       mock_sail_trim_tab_lower_bound: -7.0
       mock_sail_trim_tab_upper_bound: 7.0
-      pub_period_sec: 5.0
```
SPDonaghy: @alberto-escobar you added this line last year. Can we remove this?

PR #755, opened 2026-01-16
Limiting Visualizer Process Queue Size

FireBoyAJ24:
### Description

Currently, the `node_navigate_observer` node recieves the messages and immediately adds the messages to the queue. This causes a large increase in `VisualizerState` states whiles a few older states are visualized at the same time. We can see an observable lag in the visualizer's `desired_heading` compared to the ros2 topic.

In this PR, I have added queue size constraints and reduced the number of times the queue is updated to prevent it from increasing indefinitely. 

### Verification

- [x] Run the code for 5-10 minutes and ensure the following conditions are met
    - [x] Ensure the queue size does not exceed 1
    - [x] Ensure the Dash app's boat state is upto date with the latest information
- [x] CI tests passes

### DIFF
--- a/src/global_launch/config/globals.yaml
+++ b/src/global_launch/config/globals.yaml
@@ -89,4 +89,3 @@ mock_data_node:
       mock_sail_trim_tab: True
       mock_sail_trim_tab_lower_bound: -7.0
       mock_sail_trim_tab_upper_bound: 7.0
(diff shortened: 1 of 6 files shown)


raghumanimehta (approved):
Thanks for the PR! Nicely done

PR #759, opened 2026-01-22
Add parameter callback to sync true wind updates in MockGPS

### LINKED ISSUE #757: PATH: Fix boat drift under `true wind = 0`
### Description

- The boat keeps moving if I set the `tw_speed_kmph` to 0.
- This make no sense.  

#### Expected Behavior

- The boat should not move at all. We follow a simple kinematic model, where we don't consider waves. 
- As a result, the expectation is that the boat will stop moving the moment we set the `tw_speed_kmph` to 0. 
#### Actual Behavior

- Mentioned above

### Steps to Reproduce

1. Start the local pathfinding ros node with mock nodes
2. Set the true wind speed to 0 

### Relevant Code

- `node_mock_gps.py`
- `node_mock_wind_sensor.py`
- `ompl_objectives.py`
- The bug could be somewhere else as well. 

### Potential Solutions

- This perhaps be caused by the `get_sailbot_speed` interpolation. It's worthwhile checking that out. 
- This could also be how winds are handled or how kinematic model in `node_mock_gps.py` works. 

### Deadline: 
- Please try to implement a fix latest by Thursday (ideally Wednesday). 
- This is a critical bug which is hindering the testing, failing to implement a fix Friday would result in a reassignment.

jb3982:
### Description

- Resolves #757

Added a parameter callback (`_on_set_parameters`) to the MockGPS node to properly handle dynamic updates to true wind parameters (`tw_speed_kmph` and `tw_dir_deg`). Previously, when the true wind speed was set to `0` via parameter updates, the MockGPS node continued using its cached value, causing the boat to keep drifting. Now both MockGPS and MockWindSensor stay synchronized when wind parameters are updated at runtime.

### Verification

- [ ] Launch local pathfinding with mock nodes and set `tw_speed_kmph` parameter to `0` using the `wind_params.sh` script.
- [ ] Verify that the boat speed drops to `0` and stops moving in the dash app.

### Image before setting `tw_speed_kmph` and `tw_dir_deg` to `0`:
Screenshot 2026-01-22 at 12 52 59 PM

### Image after setting `tw_speed_kmph` and `tw_dir_deg` to `0`:
Screenshot 2026-01-22 at 1 11 07 PM

### DIFF
--- a/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
+++ b/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
@@ -125,7 +125,7 @@ class MockGPS(Node):
                     self.__tw_speed_kmph = p.value
             return SetParametersResult(successful=True)
         except Exception:
-            reason = f"Please enter the direction in (-180, 180]."
+            reason = "Please enter the direction in (-180, 180]."
             return SetParametersResult(successful=False, reason=reason)
 
     def update_speed(self):
--- a/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
+++ b/src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_wind_sensor.py
@@ -112,7 +112,7 @@ class MockWindSensor(Node):
                     self._tw_speed_kmph = p.value
             return SetParametersResult(successful=True)
         except Exception:
-            reason = f"Please enter the direction in (-180, 180]."
+            reason = "Please enter the direction in (-180, 180]."
             return SetParametersResult(successful=False, reason=reason)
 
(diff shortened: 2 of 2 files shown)


src/local_pathfinding/local_pathfinding/mock_nodes/shared_utils.py
```diff
@@ -13,3 +13,11 @@
 START_POINT = ci.HelperLatLon(latitude=49.308157, longitude=-123.244801)
 Change last line of mock_global_path.csv to: 49.289686,-123.195877
 """
+
+
+def _validate_tw_dir_deg(value: int) -> None:
```
raghumanimehta: Not a private function, should not have an underscore.

raghumanimehta (changes requested):

Issue #761, opened 2026-01-24
Controller logic for stop state


alberto-escobar:
### Purpose

The goal of this issue is to implement a "safety" or "pause" state for the boat's trim tab control. Currently, the controller node continuously calculates and publishes trim tab angles based on wind data. However, when the pathfinding node fails to calculate a valid route, the boat should cease active sailing to prevent unpredictable behavior. This change will allow the controller to monitor the pathfinding status and zero out the sail command when necessary.

### Changes

We will modify the controller node to subscribe to the `desired_heading` topic and implement conditional logic for the trim tab output:

- Add a subscriber to the 'desired_heading' topic.
- Monitor a specific boolean field within the messages in the `desired_heading` topic 
- If True: Override the standard calculation and publish a trim tab angle of 0 to the `sail_cmd`
- If False: Continue publishing the calculated trim tab angle based on the `filtered_wind_sensor` data.
- the main file you will work in in https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/controller/controller/wingsail/wingsail_ctrl_node.py

### Resources

- #765 
- https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Topics/Understanding-ROS2-Topics.html
- https://docs.ros.org/en/humble/Tutorials/Beginner-CLI-Tools/Understanding-ROS2-Nodes/Understanding-ROS2-Nodes.html
- https://docs.ros.org/en/humble/Concepts/Basic.html
- https://www.youtube.com/watch?v=q98d9Qd3iwE&list=PLu5CQvEq08tAoWAVa5HSTC7eKE7-05WLg&index=5
- https://www.youtube.com/watch?v=iJdLFAo-QYM&list=PLu5CQvEq08tAoWAVa5HSTC7eKE7-05WLg&index=6

PR #780, opened 2026-02-03
Fix #765: Add error handling for OMPL path hard crashes and stop sign…

### LINKED ISSUE #765: PATH: Stoping the boat
### Purpose

The PATH software can hard crash at a lot of places. One such example is https://github.com/UBCSailbot/sailbot_workspace/blob/2baa660a6008aa9cec087292f65f80820161cb15/src/local_pathfinding/local_pathfinding/local_path.py#L216 
Here, when the OMPL can't figure out a state space state (defined by (x, y, yaw)), OMPL hard crashes and an exception is thrown which is uncaught. There are other plroblems with handling it this way.

If there is no where for the boat to go, the boat should ideally stop moving hoping things would improve sometime.

### Tasks

Steps to achieve this: 

- [ ] Update the `DesiredHeading` msg to include a new field of type boolean that stops the boat.
- [ ] Update the code linked above in the purpose that does this unsafe operation on the `OMPLPath`. Make Simular changes everywhere to add `try except` to handle this operation safely.
- [ ] Make sure that the logic to stop is properly handled and please document the changes in https://github.com/UBCSailbot/sailbot_workspace/tree/main/src/custom_interfaces#readme

### Resources
- #761

jb3982:
### Description
- Resolves #765 

This PR addresses the issue where the pathfinding node would crash when OMPL fails to compute a valid state space solution. The changes introduce error handling and a new stop signaling mechanism.

**Changes made:**
1. Added `sail` boolean field to `DesiredHeading` message to signal when the boat should stop
2. Implemented try-except blocks in `ompl_path.get_path()` to catch OMPL exceptions
3. Added error handling in `local_path.update_if_needed()` for both new and old path retrievals
4. Updated `node_navigate.desired_heading_callback()` to detect pathfinding failures and set `sail=False`
5. Documented the new `sail` field behavior and error handling in custom_interfaces `README.md`

**Behavior:**
- When pathfinding fails, `DesiredHeading` is published with `sail=False`, `heading=0.0`, and `steering=0`
- This prevents hard crashes and provides an explicit stop signal to downstream nodes

### Verification

- Created temporary error handling test file (`test_error_handling.py`) locally and verified all tests pass
  - Tests verify exceptions are caught when `get_path()` fails
  - Tests confirm `update_if_needed()` returns `(None, waypoint_index)` on failure
  - Tests validate `desired_heading_callback()` sets `sail=False` when heading is `None`

### Resources

- #761

### DIFF
--- a/src/local_pathfinding/local_pathfinding/local_path.py
+++ b/src/local_pathfinding/local_pathfinding/local_path.py
@@ -212,17 +212,11 @@ class LocalPath:
         )
         old_ompl_path = self._ompl_path
 
-        try:
-            new_path = ompl_path.get_path()
-            if new_path is None:
-                self._logger.error("New OMPL path is None")
-                return None, local_waypoint_index
-            heading_new_path, wp_index = self.calculate_desired_heading_and_waypoint_index(
-                new_path, 0, gps.lat_lon
-            )
-        except Exception as e:
-            self._logger.error(f"Failed to get new OMPL path: {e}")
-            return None, local_waypoint_index
+        heading_new_path, wp_index = self._try_get_heading(
+            ompl_path, 0, gps.lat_lon, "new"
+        )
+        if heading_new_path is None:
+            return None, None
 
         if received_new_global_waypoint:
             self._logger.debug("Updating local path because we have a new global waypoint")
@@ -235,17 +229,11 @@ class LocalPath:
             self._update(ompl_path)
             return heading_new_path, wp_index
 
-        try:
-            old_path = old_ompl_path.get_path()
-            if old_path is None:
-                self._logger.error("Old OMPL path is None")
-                return None, local_waypoint_index
-            heading_old_path, updated_wp_index = self.calculate_desired_heading_and_waypoint_index(
-                old_path, local_waypoint_index, gps.lat_lon
-            )
-        except Exception as e:
-            self._logger.error(f"Failed to get old OMPL path: {e}")
-            return None, local_waypoint_index
+        heading_old_path, updated_wp_index = self._try_get_heading(
+            old_ompl_path, local_waypoint_index, gps.lat_lon, "old"
+        )
+        if heading_old_path is None:
+            return None, None
         # check if the current path goes through a collision zone.
         # No need to check for new path since it's fresh and ompl doesn't generate path that
         # go through a collision zone
@@ -259,6 +247,8 @@ class LocalPath:
         heading_diff_old_path = cs.calculate_heading_diff(self.state.heading, heading_old_path)
         heading_diff_new_path = cs.calculate_heading_diff(self.state.heading, heading_new_path)
 
+        assert updated_wp_index is not None
+        assert wp_index is not None
         old_cost = old_ompl_path.get_cost(updated_wp_index)
         new_cost = ompl_path.get_cost(wp_index)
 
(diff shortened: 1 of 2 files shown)


src/local_pathfinding/local_pathfinding/ompl_path.py
raghumanimehta: Change the return type to `Optional[ci.Path]`. Handle this `none` return everywhere.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -180,7 +180,18 @@ def desired_heading_callback(self):
...
+
+        # If desired_heading is None, pathfinding failed - signal boat to not sail
+        if desired_heading is None:
+            self.get_logger().error("Pathfinding failed, signaling boat to stop")
+            msg.heading.heading = 0.0
+            msg.steering = 0
+            msg.sail = False
+        else:
+            msg.heading.heading = desired_heading
+            msg.steering = 0
```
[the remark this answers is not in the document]
raghumanimehta: Also, I don't think we change `steering`. Confirm this, please.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -212,9 +212,17 @@ def update_if_needed(
...
+            new_path = ompl_path.get_path()
+            if new_path is None:
+                self._logger.error("New OMPL path is None")
+                return None, local_waypoint_index
+            heading_new_path, wp_index = self.calculate_desired_heading_and_waypoint_index(
+                new_path, 0, gps.lat_lon
+            )
+        except Exception as e:
+            self._logger.error(f"Failed to get new OMPL path: {e}")
+            return None, local_waypoint_index
```
raghumanimehta: Should perhaps return `None` for both.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -227,9 +235,17 @@ def update_if_needed(
...
+            old_path = old_ompl_path.get_path()
+            if old_path is None:
+                self._logger.error("Old OMPL path is None")
+                return None, local_waypoint_index
+            heading_old_path, updated_wp_index = self.calculate_desired_heading_and_waypoint_index(
+                old_path, local_waypoint_index, gps.lat_lon
+            )
+        except Exception as e:
+            self._logger.error(f"Failed to get old OMPL path: {e}")
+            return None, local_waypoint_index
```
raghumanimehta: Same as above comment.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -286,4 +302,7 @@ def update_if_needed(
     def _update(self, ompl_path: OMPLPath):
 
         self._ompl_path = ompl_path
-        self.path = self._ompl_path.get_path()
+        local_path = self._ompl_path.get_path()
+        if local_path is None:
```
raghumanimehta: You should update `self.path` still. You have logged the error but in the next cycle, the reference to `self.path` would have a valid path which is incorrect behaviour.

jb3982: `self.path = local_path` runs after the `if` block, since there is no return statement. So, won't it set the `self.path` to `None` and in this would trigger `if old_ompl_path is None or self.path is None:` check. where it would call `self._update(omple_path)`

raghumanimehta:
> > There are other things that are missing still. Consider the scenarios:
> > 
> > * What happens when the boat was set to`sail == False`, when would this be next checked? Have you tested this?
> > * What happens if the `OMPL` failed becuase of other reasons? I believe we could handle this by after the `sail = False`, try again immediately a couple of times. If we can't get a path, wait for, let's say, 10 minutes before attempting to generate a new path.
> >   What are your thoughts on these points? @jb3982 @SPDonaghy @FireBoyAJ24
> 
> 1. The current flow is like this:
>    
>    * `desired_heading_callback` runs and calls`get_desired_heading`, which then calls`update_if_needed`
>    * If OMPL fails => `desidered_heading = None` => sets `sail = False`
>    * On next tick, `desired_heading_callback` runs again => If OMPL succeeds in finding path => publishes `sail = True`
>      So the boat's sailing state is re-evaluated every cycle automatically.
> 2. Just to confirm I understand correctly: you're proposing something like:
>    1. On OMPL failure, immediately retry 2-3 times
>    2. If all retries fail, set `sail = False` and enter a cooldown period (~10 min)
>    3. After the cooldown, resume normal pathfinding attempts
> 
> I think this is a good idea. Could we use a shorter cooldown than 10 minutes though? Short enough so that our boat isn't drifting aimlessly for too long. What do you think is reasonable?

Yes. No, just set it to 10 minutes for now. We can discuss this later.

Issue #776, opened 2026-02-01
PATH Sending data for changing parameters through the visualizer ui


raghumanimehta:
### Purpose

- Currently, we have to use `wind_params.yaml` and `wind_params.sh` from the terminal to change the mock node characteristics.
- This is slow and cumbersome.
- Ideally, we should be able to send data through UI and do checks on it (like range, or type)
- this will make testing easier. 

### Description

- Update the `visualizer.py` file to allow the dash app to take inputs, using the inputs update the `wind_params.yaml` and run the `wind_params.sh` script to update the nodes. 
- I honestly am not sure how you would do this; so you'll probably have to do a bit of research on this. I belive you can use asynchronous callbacks from ROS for example. 
### Resources

- https://docs.ros.org/en/galactic/Tutorials/Intermediate/Monitoring-For-Parameter-Changes-CPP.html
- https://robotics.stackexchange.com/questions/111633/ros2-parameter-dynamic-update-and-node-restart

PR #785, opened 2026-02-07
Sending data for changing parameters through the visualizer ui

AMaharaj16:
Issue Link: https://github.com/UBCSailbot/sailbot_workspace/issues/776

### Summary

Adds inline wind controls to the existing visualizer UI, allowing true wind direction and speed to be adjusted live without launching a separate dashboard.

### Details

Embedded numeric inputs for true wind direction and speed directly into the visualizer

Updates wind_params.yaml and applies changes via the existing script

Keeps all changes contained to visualizer.py

### DIFF
--- a/src/local_pathfinding/local_pathfinding/visualizer.py
+++ b/src/local_pathfinding/local_pathfinding/visualizer.py
@@ -944,8 +944,9 @@ def apply_wind_params():
     subprocess.run(
         ["bash", str(WIND_PARAMS_SH)],
         check=True,
-      
-      
+    )
+
+
 def get_state_space_bounds(
     vs: VisualizerState,
 ) -> Tuple[cs.XY, cs.XY]:
@@ -1143,7 +1144,7 @@ def update_graph(
             yaxis=dict(range=y_range, autorange=False),
             uirevision="reset",
         )
-        
+
     return fig, [new_goal_xy[0], new_goal_xy[1]], last_range
 
 


AMaharaj16:
Something to note is that adjusting the wind direction parameter doesn't change the arrow angle in the wind speed/direction visualization in the bottom right of the visualizer. Should I investigate that?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -933,3 +992,30 @@ def live_plot(_: int, last_goal_xy_km: Optional[List[float]]) -> Tuple[go.Figure
...
+    State("tw-dir-input", "value"),
+    State("tw-speed-input", "value"),
+    prevent_initial_call=True,
+)
+def update_wind(_, tw_dir_deg, tw_speed_kmph):
+    try:
+        if tw_dir_deg is None or tw_speed_kmph is None:
+            return "Invalid input"
+
+        if not (0 <= tw_dir_deg <= 360):
```
raghumanimehta: Direction must be (-180, 180] not [0, 360]

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -933,3 +992,30 @@ def live_plot(_: int, last_goal_xy_km: Optional[List[float]]) -> Tuple[go.Figure
...
+            return "Direction must be 0–360°"
+
+        if tw_speed_kmph < 0:
+            return "Speed must be ≥ 0"
+
+        write_wind_params(tw_dir_deg, tw_speed_kmph)
```
raghumanimehta: A cleaner separation of concerns would be, 
you do what you are doing here but you check for the value ranges in the `write_wind_params()` function. Then you should `apply_wind_params()` after all your checks pass in the same function.
So it would be like: 
`update_wind` calls (`write_wind_params` calls (`apply_wind_params`)).

raghumanimehta (changes requested):
- Add the name of the text fields. 
- Remove "Applied wind ... " message after sometime. 
- Can you also move the fields to the bottom left cornere and move the main plot a bit above for cleanlineess. I will be adding more controls there soon.
Thanks!

raghumanimehta (dismissed):

Issue #764, opened 2026-01-25
Local Transceiver: update send() and debugsend() to check connection before using satellite and look into clearing buffer before reading


lross03:
### Purpose

We need to check the connection of the satellite before attempting to send/receive messages. It also might need to clear the buffer before beginning the send message loop.

### Changes

Check the satellite connection before sending/receiving and gracefully fail if connection is inadequate. `AT+CSQ` command is used to check connection: output of 5 indicates good connection, 0 indicates no connection. Currently unsure what value is the minimum for messages to be sent/received.

We should also look into if clearing the message buffer before beginning the send loop is necessary.

PR #786, opened 2026-02-07
User/Apps247/714-new-create-debug-send-at-function

### LINKED ISSUE #714: Create debug send function for local tansceiver satellite testing
### Purpose

It is difficult to verify the local transceiver send functionality using the deployment send function. It would be good to have a simpler function that follows the same process flow as send(), but it sends some test payload to avoid sending all sensor data (expensive). (e.g. 1 byte)

### Description

Write a new debug_send() function that is called by the send_data ros service which sends a test data payload following the same steps as regular send(). Feel free to add any debug prints or other helpful additions for debugging.

### Resources

- AT command manual: https://cdn-shop.adafruit.com/product-files/4521/4521-AT%20command.pdf
- 5.144 +SBDIX[A] - Short Burst Data: Initiate an SBD Session Extended
- 5.154 +SBDWB - Short Burst Data: Write Binary Data to the ISU

Apps247:
### Description
- Resolves #714
- Also resolves part of #764 (see last point below)

- Created debugSendAT function, with a ros2 service debug_send_data and ros2 parameter debug_data_to_send

- Updated local transceiver to use a log callback for debug and errors rather than cout and cerr. This is set by whoever initializes LocalTransceiver(). Use setLogCallbacks() to set. Currently the unit tests set it to use cout and cerr, and the ros2 service sets it to use ros logs.

- Fixed \r extra character in issue with the clearSerialBuffer() function. This should also resolve the issue of stale buffer input from previous attempts affecting the current attempt. This resolves the clearing buffer part of #764 too

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -43,7 +43,7 @@ trap 'signal_handler' EXIT
 if [ -f install/local_setup.bash ]; then source install/local_setup.bash; fi
 
 if [[ "$PACKAGE" == "network_systems" || "$PACKAGE" == "" ]]; then
-    # Change MONGODB_PASSWORD password
+    # Change MONGODB_PASSWORD password to do tests on the database
     export MONGODB_PASSWORD="placeholder"  # DO NOT PUSH THE ACTUAL PASSWORD TO GITHUB
 
     NET_DIR=src/network_systems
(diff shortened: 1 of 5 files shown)


lross03:
Merged in by #764 can close

PR #791, opened 2026-02-08
Initializing Pathfinding Visualizer Axes to Axes of State Space

### LINKED ISSUE #775: PATH: Change the visualizer such that, it is always resized to the state space
### Purpose

- It's a pain to resize the visualizer to always zoom in. 
- If I need to look at the boats, I can change this using the zoom out functionality and this is quite a rare thing to do personally.  

### Description

- Resize the display figure layout to match state space of the boat.

### Resources

-https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/visualizer.py
- https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/ompl_path.py

MilcToast:
### Description

- Resolves #775 

This PR addresses the issue where the pathfinding visualizer would initialize on x and y-axes [-100, 100], which was very inconvenient, since we were mostly interested in what was going on in the state space.

#### Changes Made:
1. Added a `zoom_needed` boolean in `build_figure()` so that on first load, `compute_and_add_state_space` would be zoomed into the state space.
2. Added `stored_range` input and `last_range` output to update_graph to store the previous axes as a result of a user panning, zooming or autoscaling.

The second change was made due to the visualizer either:
- snapping back with the state space on the top right
- having axes that were hard-coded during the second rendering

### Verification

- Tested out on loading the visualizer to ensure that axes were set to the bounds of the state space
- Tested out on the visualizer with panning the plot, zooming in/out, autoscaling to ensure that the plot didn't snap back

### DIFF
--- a/src/local_pathfinding/local_pathfinding/visualizer.py
+++ b/src/local_pathfinding/local_pathfinding/visualizer.py
@@ -722,8 +722,6 @@ def compute_and_add_state_space(
     boat_xy_km: Tuple[float, float],
     goal_xy_km: Tuple[float, float],
     fig: go.Figure,
-    zoom_needed: bool,
-    last_range: Optional[Dict[str, List[float]]],
 ):
     """
     Build the visualization state-space overlay around the boat and goal. Then, add the built
@@ -737,8 +735,6 @@ def compute_and_add_state_space(
         boat_xy_km: (x, y) boat position in km.
         goal_xy_km: (x, y) goal position in km.
         fig: Target Plotly figure.
-        zoom_needed: whether we want to zoom into the state space
-        last_range: previously stored axis ranges to maintain axes if zoom not needed.
     """
     boat_pos = cs.XY(boat_xy_km[0], boat_xy_km[1])
     goal_pos = cs.XY(goal_xy_km[0], goal_xy_km[1])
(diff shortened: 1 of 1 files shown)


src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -714,7 +718,12 @@ def add_arrow(origin, dx, dy, color, title, mag):
 
 
 def compute_and_add_state_space(
-    boat_xy_km: Tuple[float, float], goal_xy_km: Tuple[float, float], fig: go.Figure
+    vs: VisualizerState,
```
raghumanimehta: The function is doing two sort-of related tasks. `compute_and_state_space` is doing something that is unexpected given the name of the function. Also, a function should ideally do one thing. 
You should handle the zoom logic somewhere else.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -877,12 +903,39 @@ def build_figure(
         fig.add_annotation(annotation)
 
     # Computing State space overlay and adding it to the plot
-    compute_and_add_state_space(boat_xy_km, goal_xy_km, fig)
+    zoom_needed = last_goal_xy_km is None or last_range is None
+    compute_and_add_state_space(state, boat_xy_km, goal_xy_km, fig, zoom_needed, last_range)
```
raghumanimehta: Perhaps the zooming logic can go to `apply_layout`.

raghumanimehta (changes requested):
Good work! Please make changes mentioned in the review and as discussed.

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -714,7 +718,10 @@ def add_arrow(origin, dx, dy, color, title, mag):
...
 
 def compute_and_add_state_space(
-    boat_xy_km: Tuple[float, float], goal_xy_km: Tuple[float, float], fig: go.Figure
+    vs: VisualizerState,
+    boat_xy_km: Tuple[float, float],
+    goal_xy_km: Tuple[float, float],
+    fig: go.Figure,
 ):
     """
     Build the visualization state-space overlay around the boat and goal. Then, add the built
```
raghumanimehta: You should make a note that this function changes the `vs`

PR #798, opened 2026-02-12
Raghumanimehta/778 get cost

### LINKED ISSUE #778: PATH: Fix get_cost
### Description

`get_cost` is not tracking the index of the waypoint to get the `cost`

raghumanimehta:
- Added a new function `get_remaining_cost` which calculates the cost of the remaining OMPL Path. 
- The function assumes monotonically linearly decreasing cost as the path is consumed/traversed and interpolates it accordingly. For example, if the cost between two points is `x` and the boat has covered half the distance between the two points then the function `get_remaining_cost` would return `0.5x`.
- This function is to be used temporarily for the `conditions_to_switch_paths` which will ultimately be refactored to not use this function. However, this function is very important for the purpose of testing and data collection. 
- There is an edge case to look out for. There is a chance that the boat doesn't linearly travel between two points and may deviate from the linear path; this could result in fraction remaining be `> 1`. For such cases, I have assumed that in doing so the cost would not have increased beyond the start point and as a result I project this vector (from start to current boat position) on to the vector from start to the end of the segment. 
### Description

- Resolves #778 
### Verification

- [x] Added tests which are passing 
- [x] Ran the visualizer to ensure that it wasn't crashing

### DIFF
--- a/src/network_systems/projects/can_transceiver/inc/can_frame_parser.h
+++ b/src/network_systems/projects/can_transceiver/inc/can_frame_parser.h
@@ -111,7 +111,9 @@ enum class CanId : canid_t {
     PRESSURE_14           = 0x13E,
     PRESSURE_SENSOR_END   = 0x13F,
     GENERIC_SENSOR_START  = 0x140,
-    GENERIC_SENSOR_END    = 0x1FF
+    GENERIC_SENSOR_END    = 0x1FF,
+    DEBUG_START           = 0x200,
+    DEBUG_END             = 0x2FF
 };
 
 inline bool isValidCanId(canid_t id)
@@ -135,7 +137,9 @@ inline bool isValidCanId(canid_t id)
             id <= static_cast<canid_t>(CanId::PRESSURE_SENSOR_END)) ||
 
            (id >= static_cast<canid_t>(CanId::GENERIC_SENSOR_START) &&
-            id <= static_cast<canid_t>(CanId::GENERIC_SENSOR_END));
+            id <= static_cast<canid_t>(CanId::GENERIC_SENSOR_END)) ||
+
+           (id >= static_cast<canid_t>(CanId::DEBUG_START) && id <= static_cast<canid_t>(CanId::DEBUG_END));
 }
 
 /**
--- a/src/network_systems/projects/can_transceiver/src/can_transceiver.cpp
+++ b/src/network_systems/projects/can_transceiver/src/can_transceiver.cpp
@@ -27,13 +27,21 @@ void CanTransceiver::onNewCanData(const CanFrame & frame) const
         return;
     }
     CanId id{frame.can_id};
-    if (read_callbacks_.contains(id)) {
-        read_callbacks_.at(id)(frame);  // invoke the callback function mapped to id
+
+    if (id >= CanId::DEBUG_START && id <= CanId::DEBUG_END) {
+        // ELEC debug frame, do nothing
+        return;
     }
 
     //0x100 - 0x1FF: Generic sensor data frame range
     if (id >= CanId::GENERIC_SENSOR_START && id <= CanId::GENERIC_SENSOR_END) {
         read_callbacks_.at(CanId::GENERIC_SENSOR_START)(frame);
+        return;
+    }
+
+    if (read_callbacks_.contains(id)) {
+        read_callbacks_.at(id)(frame);  // invoke the callback function mapped to id
+        return;
     }
 }
 
--- a/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
+++ b/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
@@ -368,9 +368,8 @@ private:
             wind_sensor_msg                   = wind_sensor.toRosMsg();
             wind_sensors_pub_->publish(wind_sensors_);
 
-            // NUM_WIND_SENSORS is a placeholder,
-            // replace with number of data points wanted in the moving average
-            double k = NUM_WIND_SENSORS;
+            // Arbitrary number of data points for moving average
+            double k = 20;  //NOLINT(readability-magic-numbers)
             // convert deg to rad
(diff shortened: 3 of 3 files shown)


src/local_pathfinding/test/test_ompl_path.py
```diff
@@ -269,3 +278,146 @@ def test_create_space(
...
+    waypoints = fresh_ompl_path.get_path().waypoints
+    boat_latlon = waypoints[wp_index]
+    next_wp_latlon = waypoints[wp_index + 1] if wp_index + 1 < len(waypoints) else None
+
+    def mid_point(start_latlon: HelperLatLon, end_latlon: HelperLatLon):
+        if end_latlon is None:
+            return start_latlon
+        return HelperLatLon(
+            latitude=(random.uniform(start_latlon.latitude, end_latlon.latitude)),  # noqa
+            longitude=(random.uniform(start_latlon.longitude, end_latlon.longitude)),  # noqa
```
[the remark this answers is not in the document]
raghumanimehta: It's capped; even though the final value is undeteministic, it should work fine @copilot .

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -103,7 +132,7 @@ def calculate_desired_heading_and_waypoint_index(
         Args:
             path (ci.Path): Array of waypoints
             waypoint_index (int): Pointer to the current local waypoint index in path array
```
SPDonaghy:
```suggestion
            waypoint_index (int): The index of the waypoint sailbot most recently reached in the path array.
```

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -199,53 +230,72 @@ def update_if_needed(
...
+            self.state = new_state
+            self._update(new_ompl_path)
             return heading_new_path, wp_index
 
         heading_diff_old_path = cs.calculate_heading_diff(self.state.heading, heading_old_path)
         heading_diff_new_path = cs.calculate_heading_diff(self.state.heading, heading_new_path)
 
-        old_cost = old_ompl_path.get_cost(updated_wp_index)
-        new_cost = ompl_path.get_cost(wp_index)
+        old_last_wp_index = max(updated_wp_index - 1, 0)
```
SPDonaghy: ditto on my comment about using _last_

src/local_pathfinding/local_pathfinding/ompl_objectives.py
```diff
@@ -12,9 +12,9 @@
 UPWIND_COST_MULTIPLIER = 1.0
 DOWNWIND_COST_MULTIPLIER = 1.0
 ZERO_SPEED_COST = 1.0
-ACCEPTABLE_COST_THRESHOLD = 0.0
-WIND_OBJECTIVE_WEIGHT = 1.0
-TIME_OBJECTIVE_WEIGHT = 1.0
+ACCEPTABLE_COST_THRESHOLD = 0.85
```
SPDonaghy: The description didn't mention tuning these values, what work was done to obtain these values? That would be something to explain with a comment

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -52,6 +52,9 @@ class Sailbot(Node):
         global_path (ci.Path): Path that we are following.
         filtered_wind_sensor (ci.WindSensor): Filtered data from the wind sensors.
         desired_heading (ci.DesiredHeading): current desired heading.
+        last_lp_wp_index: index of the last local waypoint seen. This is the waypoint that has
```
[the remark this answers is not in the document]
raghumanimehta: I am sorry, I don't get what you are quoting here. I have streamlined this comment anyway.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -206,13 +203,101 @@ def get_cost(self, waypoint_index: int) -> float:
...
+        if total_seg_dist >= DISTANCE_THRESHOLD and dist_boat_to_start >= DISTANCE_THRESHOLD:
+            # Project the boat's position onto the segment vector to compute traveled distance.
+            # dot(boat_vec, seg_vec) / ||seg_vec|| gives the scalar projection length
+            projected_dist = (dx_seg * dx_boat + dy_seg * dy_boat) / total_seg_dist
+
+            # clamp to [0, total_seg_dist]
+            projected_dist = max(0.0, min(projected_dist, total_seg_dist))
+            fraction_travelled = projected_dist / total_seg_dist
+        else:
+            # segment is degenerate or boat is at start
```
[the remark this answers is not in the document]
raghumanimehta: degenerate in the sense that the segement length between state space pos 1 and state space pos 2 is < 1e-9. Doesn't ever really happen. I am going to remove the entire commnent.

src/local_pathfinding/local_pathfinding/local_path.py
SPDonaghy: @raghumanimehta Thoughts on this? its a nit bc its not really a pointer

src/local_pathfinding/local_pathfinding/node_navigate.py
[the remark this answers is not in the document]
raghumanimehta: Yes, it does make sense now. I editted the comment but it's still quite verbose.
The reason I added this verbose comment is because earlier `local_wp_index` was a source of confusion. Some places in the codebase were using it to track the waypoint Polaris was heading towards; this is fundamentally not how OMPL works. I wanted to ensure that this doesn't ever confuse us. Lmk what you think after this justification.

src/local_pathfinding/local_pathfinding/ompl_objectives.py
raghumanimehta: I didn't mean to change. I didn't mention it in the PR because this was going to be addressed in another task as soon as Aayush is done with wind cost model.

PR #800, opened 2026-02-12
752 protobuf encode decode script

### LINKED ISSUE #752: Create a script (or similar) that serializes global path data
### Purpose

We want to manually send data to the satellite modem to test the receive function. Receive expects the data to be protobuf, so we want to be able to generate a waypoint protobuf object and send it manually via the ROCKBLOCK website.

### Purpose
Write a script or other tool that can generate protobuf global waypoint data so we can send it via the ROCKBLOCK website.

lross03:
### Description

- Resolves #752
- This script is able to encode global path protobuf objects, and decode protobuf objects.
- This will be used for manually testing Iridium satellite communication.

### Verification

- [x] Manually verified encoding/decoding protobuf objects

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -29,8 +29,8 @@ while getopts "hp:q" flag; do
     esac
 done
 
-# Assign build type debug
-BUILD_TYPE="Debug"
+# Assign build type Release (full optimizations and no debug symbols)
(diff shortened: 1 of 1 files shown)


Jng468 (approved):
Very cool

PR #814, opened 2026-02-21
install vim and add ros user to dialout permission group

digcheese:
### Description

- Adds QOL updates to the release devcontainer
- installed vim
- added the ros user to the dialout permission group

### Verification

- container builds properly


SPDonaghy:
LGTM, approval pending confirmation from Liam that new image meets requirements/ works

PR #815, opened 2026-02-21
Fix #808: Added helper function that determines if wind conditions are different enough

### LINKED ISSUE #808: PATH: Conditions to switch paths -> wind helper
### Purpose

This will be the first part of the conditions to switch paths task. 
Read the parent issue and the documentation attached to this issue. 
For this part, write and test a helper function that takes 2 sets of wind conditions i.e. (`tw_dir_deg` (aka true wind direction in degrees in global coordinate) and `tw_speed_kmph` (aka true wind speed in kmph)) and compares the second with respect to the first set and determines if the wind conditions are different enough. 

Set a particular threshold to consider if the wind condition is different or not. Read the documentation to get an idea about this. Approach me if you have any questions. 
### Changes

- the funciton shoud be added in `Local_path` i.e. a method of the class 
- the funciton should be well tested 
### Resources

- Link to the documentation with the design is available above.

MilcToast:
### Description

- Resolves #808 

#### Changes made
1.  Added `Wind` abstract data class with fields `speed_kmph` and `dir_deg` 
2. Created helper function `significant_wind_change` 

#### Behavior 
- Returns `true` if the difference between the user-provided average wind speed over the past *x* seconds and the wind speed at path creation time exceeds **30% of the wind speed at path creation**.
- Returns `true` if the difference between the user-provided average wind direction over the past *x* seconds and the wind direction at path creation time exceeds **10 degrees**.

The thresholds were decided according to proposed values from: 
image

#### Verification

- Added unit tests in `test_local_path.py` covering:
  - Wind speed change above and below the 30% threshold
  - Wind direction change above and below the 10° threshold
  - 10° threshold upholds circular angle logic, correctly handling wraparound at ±180°

### DIFF
--- a/src/local_pathfinding/test/test_local_path.py
+++ b/src/local_pathfinding/test/test_local_path.py
@@ -249,10 +249,10 @@ def test_in_collision_zone(local_wp_index, reference_latlon, path, obstacles, re
         ),
     ],
 )
-def test_significant_wind_change(new_tw_data,
-                                 previous_tw_data,
-                                 result):
-    assert PATH.significant_wind_change(new_tw_data, previous_tw_data) == result
+def test_is_significant_wind_change(new_tw_data,
+                                    previous_tw_data,
+                                    result):
+    assert PATH.is_significant_wind_change(new_tw_data, previous_tw_data) == result
 
 
 def test_LocalPathState_parameter_checking():


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -157,6 +158,40 @@ def in_collision_zone(
...
+        """
+
+        # Check for significant changes
+        tw_speed_kmph = previous_tw_data.speed_kmph
+        tw_dir_deg = previous_tw_data.dir_deg
+
+        speed_change_percent = abs(tw_speed_kmph - avg_tw_speed_kmph) / tw_speed_kmph
+        dir_change = abs(cs.bound_to_180(avg_tw_dir_deg - tw_dir_deg))
+
+        return speed_change_percent >= 0.3 or dir_change >= 10
```
raghumanimehta: Magic numbers. Please make them constants and define them at the top of the file.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -157,6 +158,40 @@ def in_collision_zone(
...
+    @staticmethod
+    def significant_wind_change(
+        avg_tw_speed_kmph: float,
+        avg_tw_dir_deg: float,
+        previous_tw_data: wcs.Wind,
+    ):
+        """Returns true if there is a significant change in the wind warranting a change in path
+
+        Evaluates a short history of wind speed compared to the wind condition used when previous
+        path was made. Assume that the current wind speed is above MINIMUM_WIND_SPEED.
```
raghumanimehta: Where does the assumption come from?

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -157,6 +158,40 @@ def in_collision_zone(
                     return True
         return False
 
+    @staticmethod
+    def significant_wind_change(
+        avg_tw_speed_kmph: float,
```
raghumanimehta: Average wind data should also be a `Wind` Object like the current wind data.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -157,6 +158,40 @@ def in_collision_zone(
...
+        avg_tw_dir_deg: float,
+        previous_tw_data: wcs.Wind,
+    ):
+        """Returns true if there is a significant change in the wind warranting a change in path
```
raghumanimehta: Explicitly mention that although this function would work with all sorts of wind, the orignal intention was to use this for `true wind`

src/local_pathfinding/test/test_local_path.py
AMaharaj16: Try to update the testing to import  `WIND_SPEED_CHANGE_THRESH_PROP` and `WIND_DIRECTION_CHANGE_THRESH_DEG` so that tests don't fail if these values change as we move forward with pathfinding.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -174,6 +177,44 @@ def in_collision_zone(
                     return True
         return False
 
+    @staticmethod
+    def significant_wind_change(
```
raghumanimehta: Rename the function to `is_significant_wind_change`. Reader might not realize that this is a boolean returning function.

raghumanimehta (approved):

PR #822, opened 2026-02-23
User/rnar17/new update collision zones

### LINKED ISSUE #458: Updating the Boat Collision Zone's Size
### Purpose

Adjust the boat obstacle size after testing and validating that the obstacle class is functioning correctly.
https://github.com/UBCSailbot/sailbot_workspace/blob/70669dbc4f453aab8f1e0d3c210c483135d92584/src/local_pathfinding/local_pathfinding/obstacles.py#L159

### Changes

We will just need to adjust the COLLISION_ZONE_STRETCH_FACTOR or the formula if the boat collision zone is too large or too small.

rnar17:
### Description
 Implemented three distinct collision zone shapes for Boat obstacles based on the ship's motion relative to Sailbot. Added `rot_to_rad_per_sec()` as a shared helper in coord_systems.py since its logic is needed in both obstacle calculation and mock AIS ship simulation.

### Case 1: No collision projected (Circle)
  When `_calculate_projected_distance()` returns zero, no intersection between the AIS ship and Sailbot is possible. A circular zone is placed around the ship's current position using radius = max(width, length), then buffered by radius + BOAT_BUFFER. The double-radius buffer provides an outer safety margin scaled to the ship's size.

 ### Case 2: Straight-line motion (Rectangle)
  When ROT ≈ 0, the ship travels in a straight line. A rectangle extends from the stern to the projected collision point. Only BOAT_BUFFER is applied; the rectangle already covers the full collision path exactly, so only a small physical buffer is needed.

 ### Case 3: Turning motion (Triangle)
  When ROT is non-zero and a collision is still projected after DT = 10 seconds of turning, a triangle is constructed in the ship's local frame:
  - A — current bow position
  - B — projected collision point if the ship continued on its current heading
  - C — projected collision point after applying ROT over the 10-second horizon
  
  No buffer is applied to the triangle. The geometry itself already encodes the swept uncertainty region: the two forward vertices (B and C) fan outward as the ship turns, naturally capturing the lateral spread introduced by the rotation. Adding a buffer artificially distorts its shape.

 ### Case 3 fallback: Turning boat turns away (Circle)
  If the ship was on a collision course but, after turning for DT seconds, is no longer projected to reach Sailbot, a circular zone is placed, similar to Case 1.

  ## Changes

  - `obstacles.py` — `_set_circular_zone()`, `_set_straight_line_zone()`, `_set_turning_zone()` replacing the previous single-path collision zone logic
  - `coord_systems.py` — `rot_to_rad_per_sec()` with full AIS spec handling (special values ±127, −128)
  - `mock_nodes/node_mock_ais.py` — updated to use the shared `rot_to_rad_per_sec()`
  - `test/test_obstacles.py` — unit tests for all three cases, including geometry verification for the straight-line and turning zones
  - `test/test_coord_systems.py` — unit tests for rot_to_rad_per_sec()

### Verification
  Unit tests in test_obstacles.py verify:
  - No-collision case produces a circular zone
  - Straight-line case (ROT = 0) produces a rectangular zone and verifies stern vertex position
  - Turning motion case (ROT ≠ 0) produces a 3-vertex triangular zone and verifies all three triangle vertices match the expected world-frame geometry
  
Unit tests in test_coord_systems.py verify `rot_to_rad_per_sec()` for normal values, special AIS values (±127, −128), and out-of-range inputs.

## Video

The following video shows the effect turning motion (ROT = 50) has on two AIS ships in the vicinity of Sailbot and the goal path. Since ROT > 0, we are oscillating between Case 1, Case 3, and Case 3 fallback. 

https://github.com/user-attachments/assets/ecb02659-eecb-4d50-9192-8c78f79ae39f

## Clockwise Rotation Formula Visual
(Via Claude). The diagram was generated by running the exact same clockwise rotation formula used in _set_turning_zone() — center of curvature placement, the rotation of (dx, dy) by turn_angle, and the future heading update — and then plotting the result. The four printed checks (both radius distances, arc length, future heading) are computed from the output coordinates, not from the formula inputs, so they independently confirm the formula produced geometrically consistent results.
image


src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -233,33 +236,88 @@ def update_collision_zone(self, **kwargs) -> None:
...
+
+        if projected_distance == PROJ_DISTANCE_NO_COLLISION:
+            # If no collision detected, create small collision zone around the boat
+            RADIUS_MULTIPLIER = 5
+            radius = cs.km_to_meters(max(self.width, self.length)) * RADIUS_MULTIPLIER
+            boat_collision_zone = Point(-self.width / 2, -self.length / 2).buffer(
+                radius, resolution=16
+            )
+            raw = self._translate_collision_zone(boat_collision_zone)
+            self._raw_collision_zone = raw
```
raghumanimehta: I would suggest assigning `self._raw_collision_zone` directly.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -233,33 +236,88 @@ def update_collision_zone(self, **kwargs) -> None:
...
+            prepared.prep(self.collision_zone)
 
-        # Points of the boat collision cone polygon before rotation and centred at the origin
-        boat_collision_zone = Polygon(
-            [
-                [-self.width / 2, -self.length / 2],
-                [-collision_zone_width, self.length / 2 + projected_distance],
-                [collision_zone_width, self.length / 2 + projected_distance],
-                [self.width / 2, -self.length / 2],
+        elif abs(rot_rps) < 1e-4:
```
raghumanimehta: How did you come up with this threshold? + Magic number

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -233,33 +236,88 @@ def update_collision_zone(self, **kwargs) -> None:
...
+            self._raw_collision_zone = raw
+            self.collision_zone = raw.buffer(BOAT_BUFFER, join_style=2)
+            prepared.prep(self.collision_zone)
+        else:
+            # Turning motion
+            R = speed_kmps / abs(rot_rps)
+            turn_angle = rot_rps * dt
+
+            # Center of rotation in world frame
+            cx = x - R * math.sin(cog_rad) * math.copysign(1, rot_rps)
```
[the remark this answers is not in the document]
raghumanimehta: Also, can you please explain the math here. Especially why does `cx = x - R ...` whereas `cy = y + R...`? That is, why is there a discrepency in the signs of those calculations. It is not immediately obvious to me.

raghumanimehta (changes requested):
Thanks for the PR @rnar17 !! This task was indeed quite complicated. 

I have pointed a couple of changes, and I have also requested a review on coordinate systems given PATH's troubles with it. I am in the process of updating the documentation in code and confluence. I will notify you when that's done. 

Please use that to verify the coordinates on this PR. You can start working on the other changes I have requested on this PR.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -233,33 +236,88 @@ def update_collision_zone(self, **kwargs) -> None:
...
+        dt = DT
+        speed_kmps = self.ais_ship.sog.speed / 3600.0
+        cog_rad = math.radians(self.ais_ship.cog.heading)
+        x, y = cs.latlon_to_xy(self.reference, self.ais_ship.lat_lon)
+        projected_distance = self._calculate_projected_distance(cog_rad)
+        bow_y = self.length / 2
+
+        if projected_distance == PROJ_DISTANCE_NO_COLLISION:
+            # If no collision detected, create small collision zone around the boat
+            RADIUS_MULTIPLIER = 5
```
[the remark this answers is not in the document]
rnar17: Just with testing on the visualizer, I thought that it was a good multiplier to make the collision zone around the boat a reasonable size. This is only for the case where there is no projected colllision though.

src/local_pathfinding/local_pathfinding/obstacles.py
rnar17: Changed it to a constant. I just wanted it to represent the smallest possible ROT that we could categorize into straight line motion.

raghumanimehta (changes requested):
Thanks for the work on this PR! This task is really complex and it took me a lot of time to understand what was going on and I haven't made it all the way through. I will review more once you make the requested changes. 
The most important thing to consider is the coordinate system differences I have pointed out. Full disclosure: I didn't go super deep into the calculation to verify it's incorrect, I just want you to review it again.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
         ais_ship = kwargs.get("ais_ship", None)
         if ais_ship is not None:
             if ais_ship.id != self.ais_ship.id:
                 raise ValueError("Argument AIS Ship ID does not match this Boat instance's ID")
             self.ais_ship = ais_ship
 
-        projected_distance = self._calculate_projected_distance()
+        # -- Shared values used across all cases --
+        rot = self.ais_ship.rot.rot
+        rot_rps = cs.rot_to_rad_per_sec(rot)
```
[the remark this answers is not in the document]
raghumanimehta: This comment was resolved but the errorr is still not handled gracefully. What we might want to do is to have the rot as 0 if the function throws a `ValueError`.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -204,6 +207,7 @@ def __init__(
         self.ais_ship = ais_ship
         self.width = cs.meters_to_km(self.ais_ship.width.dimension)
```
raghumanimehta, line 208: to improve readibility, can you please append the unit to name of the variable/field? Eg. this would then become `self.width_m`.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
-        boat_collision_zone = Polygon(
-            [
-                [-self.width / 2, -self.length / 2],
-                [-collision_zone_width, self.length / 2 + projected_distance],
-                [collision_zone_width, self.length / 2 + projected_distance],
-                [self.width / 2, -self.length / 2],
+        # --- Case 1: No collision projected ---
+        if projected_distance == PROJ_DISTANCE_NO_COLLISION:
+            # If no collision detected, create small collision zone around the boat
+            radius = cs.km_to_meters(max(self.width, self.length)) * speed_kmps
```
raghumanimehta, line 256: Can you confirm this is not a unit mismatch? We have `m * km / s`? Raidus should end up with meter or km. Also, wheny you fix this append the unit here as well like `radius_m`.

raghumanimehta (changes requested):
Hi, I have requested some more changes. Please note, I made some comments on this PR that might not appear as part of this PR review (but those comments are part of this PR review).

src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -246,3 +244,39 @@ def _latlon_point_to_xy_point(latlon_point: tuple) -> Point:
 def latlon_list_to_xy_list(reference_latlon, lat_lon_list: List[ci.HelperLatLon]) -> List[XY]:
     """Converts a list of lat/lon coordinates to x/y coordinates."""
     return [latlon_to_xy(reference=reference_latlon, latlon=pos) for pos in lat_lon_list]
+
+
+def rot_to_rad_per_sec(rot: int) -> float:
+    """
+    Convert an AIS rate-of-turn (ROT) value into radians per second via the formula:
```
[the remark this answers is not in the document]
FireBoyAJ24: A quick return note over the formula and the units would be nice.

src/local_pathfinding/local_pathfinding/obstacles.py
FireBoyAJ24: I second this. `self.width`  and `self.length` are in km and need to be that for the transformation matrix, as `cs.latlon_to_xy` converts x and y to km. However, we are mixing it up here with this line as suggested by Raghumani.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
+            boat_collision_zone = Point(-self.width / 2, -self.length / 2).buffer(
+                radius, resolution=16
+            )
+            self._raw_collision_zone = self._translate_collision_zone(boat_collision_zone)
+            self.collision_zone = self._raw_collision_zone.buffer(
+                radius + BOAT_BUFFER, join_style=2
+            )
+            prepared.prep(self.collision_zone)
```
FireBoyAJ24: This line is repeated for each case. We should place this outside of the if statement, as it is needed for all cases.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
+            C = [
+                future_projected_distance * math.sin(turn_angle),
+                self.length / 2 + future_projected_distance * math.cos(turn_angle),
             ]
-        )
 
-        # this code block translates and rotates the collision zone to the correct position
+            boat_collision_zone = Polygon([A, B, C])
+            self._raw_collision_zone = self._translate_collision_zone(boat_collision_zone)
+            self.collision_zone = self._raw_collision_zone
```
FireBoyAJ24: We need to add a buffer here as the ship can drift variably as it rotates around the center.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
+
+            if future_projected_distance == PROJ_DISTANCE_NO_COLLISION:
+                radius = cs.km_to_meters(max(self.width, self.length)) * RADIUS_MULTIPLIER
+                boat_collision_zone = Point(-self.width / 2, -self.length / 2).buffer(
+                    radius, resolution=16
+                )
+                self._raw_collision_zone = self._translate_collision_zone(boat_collision_zone)
+                self.collision_zone = self._raw_collision_zone.buffer(
+                    radius + BOAT_BUFFER, join_style=2
+                )
```
FireBoyAJ24: We should add a `prepared.prep(self.collision_zone)` here before the return statement of this sub-case.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,124 @@ def update_sailbot_data(
...
+        # --- Case 1: No collision projected ---
+        if projected_distance == PROJ_DISTANCE_NO_COLLISION:
+            # If no collision detected, create small collision zone around the boat
+            radius = cs.km_to_meters(max(self.width, self.length)) * speed_kmps
+            boat_collision_zone = Point(-self.width / 2, -self.length / 2).buffer(
+                radius, resolution=16
+            )
+            self._raw_collision_zone = self._translate_collision_zone(boat_collision_zone)
+            self.collision_zone = self._raw_collision_zone.buffer(
+                radius + BOAT_BUFFER, join_style=2
```
FireBoyAJ24: The buffer radius varies by case. This can cause some confusion when testing and dealing with potential edge cases, as we continuously test it across different scenarios.

For example,
Case 1: AIS ship's collision zone buffer is `radius + BOAT_BUFFER` where the `radius = cs.km_to_meters(max(self.width, self.length)) * speed_kmps`
Case 2: AIS ships' collision zone buffer is `BOAT_BUFFER`
Case 3: AIS ship's collision zone buffer is `radius + BOAT_BUFFER` where the `radius = cs.km_to_meters(max(self.width, self.length)) * RADIUS_MULTIPLIER`

The radius formula uses different multipliers, and the buffer zone varies by case. Is there a reason why? We should add a docstring or additional comments explaining why these differ and why they work. It does make sense why they may differ due to the unique uncertainty in each case, but it took a bit of time to figure that out.

FireBoyAJ24 (commented):
The `update_collision_zones` is doing a lot of stuff. We should break it down into multiple smaller functions where each smaller function deals with each case, and the `update_collision_zones` just dispatches it to those cases.

src/local_pathfinding/custom-rosdep.yaml
```diff
@@ -12,3 +12,8 @@ python3-anywidget:
   ubuntu:
     pip:
       packages: [anywidget]
+
+# python3-ompl:
```
raghumanimehta, line 16: Why were these comments added?

src/local_pathfinding/package.xml
```diff
@@ -20,6 +20,7 @@
   <exec_depend>python3-dash</exec_depend>
   <exec_depend>python3-anywidget</exec_depend>
   <exec_depend>python3-scipy</exec_depend>
+  <!-- <exec_depend>python3-ompl</exec_depend> -->
```
raghumanimehta, line 23: why was this added?

src/local_pathfinding/local_pathfinding/coord_systems.py
```diff
@@ -92,10 +93,7 @@ def polar_to_cartesian(angle_rad: float, magnitude: float) -> XY:
             - x → east component
             - y → north component
     """
-    return XY(
-        x=magnitude * math.sin(angle_rad),
-        y=magnitude * math.cos(angle_rad)
-    )
+    return XY(x=magnitude * math.sin(angle_rad), y=magnitude * math.cos(angle_rad))
```
raghumanimehta: @copilot fix indentation here to make it more readable.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -4,18 +4,21 @@
...
 from shapely.geometry import MultiPolygon, Point, Polygon
 
+import custom_interfaces.msg as ci
 import local_pathfinding.coord_systems as cs
 
 # Constants
+DT = 10
 PROJ_DISTANCE_NO_COLLISION = 0.0
-BOAT_BUFFER = 0.25  # km
+BOAT_BUFFER = 0.1  # km
```
raghumanimehta:
```suggestion
BOAT_BUFFER_KM = 0.1
```

raghumanimehta: and consequent changes in the rest of file

raghumanimehta: I would like to see similar changes on other constants and variables. Thanks

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -220,46 +224,157 @@ def update_sailbot_data(
...
             self.ais_ship = ais_ship
 
-        projected_distance = self._calculate_projected_distance()
+        try:
+            rot_rps = cs.rot_to_rad_per_sec(self.ais_ship.rot.rot)
+        except ValueError:
+            rot_rps = 0.0  # invalid ROT: treat as no rotation information
+
+        speed_kmps = self.ais_ship.sog.speed / 3600.0
+        # COG is measured in degrees clockwise from North (0° = North, 90° = East)
```
raghumanimehta:
```suggestion
        # COG is measured in degrees clockwise from North (0° = North, 90° = East) and is [0°, 360°)
```

raghumanimehta (changes requested):
Just a couple more changes for me. I skimmed the PR, I believe the logic is the same which looks good to me.

src/local_pathfinding/custom-rosdep.yaml
raghumanimehta: @rnar17, can you let me know why you added the commented code? I don't see it on main.

src/local_pathfinding/mock_ais_files/ais_ships.csv
```diff
@@ -1,6 +1,2 @@
 id,latitude,longitude,heading,speed,rot,width,height
-1,49.01261318191024,-135.65593922412154,90.0,20.0,30,1.0,1.0
-2,48.67919212345747,-132.94724986187234,90.0,25.0,50,2.0,2.0
-3,48.11352031871804,-135.67990496461405,30.0,15.0,-50,3.0,3.0
-4,48.1970229368345,-128.93710143805356,4.0,10.0,-30,4.0,4.0
-5,48.51027819947852,-126.15925374786544,5.0,5.0,20,5.0,5.0
+1,49.27907901001712,-123.37626882688295,90.0,10.0,0.0,1.0,1.0
```
raghumanimehta, line 2: Why reduce the test cases?

raghumanimehta (changes requested):
Just a couple of questions before i merge this in.

src/local_pathfinding/mock_ais_files/ais_ships.csv
rnar17: This was a while ago, I believe the other ships weren't close enough to Sailbot to show meaningful collision zones on the visualizer. But ais ships are now pulled from basic.yaml instead of here, right?

src/local_pathfinding/custom-rosdep.yaml
rnar17: Forgot to remove that sorry. Was trying to fix a dependency issue which wasn't allowing my visualizer to run - those lines are not needed.

raghumanimehta:
@copilot+gpt-5.4-mini src/local_pathfinding/local_pathfinding/obstacles.py:254:52: W291 trailing whitespace

src/local_pathfinding/local_pathfinding/obstacles.py:313:75: W291 trailing whitespace

src/local_pathfinding/local_pathfinding/obstacles.py:314:73: W291 trailing whitespace

src/local_pathfinding/local_pathfinding/obstacles.py:317:72: W291 trailing whitespace

src/local_pathfinding/local_pathfinding/obstacles.py:332:76: W291 trailing whitespace
Fix these

FireBoyAJ24 (approved):

PR #823, opened 2026-02-23
Added a script to main_launch.py and a test to explore_bag_data.ipynb to verify that data published to local_path is collected.

### LINKED ISSUE #753: Pathfinding data plotting workflow
### Purpose

- We want to plot the data that the `local_path` node publishes. This will allow us to verify pathfinding algorithms and evaluate how good a path is in long voyages. 
- The idea is we run pathfinding algorithm and let it run for a long time (perhaps a day or two) and then plot all the data published. 

### Description

- This will allow us to verify pathfinding algorithm and draw insights on whether path is as expected, doesn't change too often (isn't jittery) and is decently effecient.

ethanhu912:
### Description
- Resolves #753 
- Added script to `main_launch.py` to automatically create a rosbag in that stores data published to local_path.
- This rosbag is then converted into a pandas DataFrame which allows us to access and read the data (2nd cell in `explore_bag_data.ipynb`).
- We then use this DataFrame to create plots, tables, etc. for us to analyze the performance of our algorithm.
- Added a code cell to `explore_bag_data.ipynb` to display usable information from the first 5 points stored in the rosbag while the system is running
- To view this data, run the two first code cells in `explore_bag_data.ipynb`. You can run the 3rd code cell to display the first 5 messages published to the topic, and the 4th cell takes each message and extracts the data into a number for easier analysis.
- Added a launch argument `record` that only records the data when its value is set to true `record:=true.`
- Added a launch argument `save_path` that allows the user to specify the folder where the rosbag will be stored if `record` is already set to true. 
- Note: `record` is defaulted to false and `save_path` is defaulted to workspaces/sailbot_workspace/notebooks/local_pathfinding/session_recordings/.

### Verification
- Tested using the added code cell in `explore_bag_data.ipynb` to verify all fields are included and the correct data is being recorded.
- Verified both launch arguments by running the main launch with each launch argument set to true, false, etc. and confirmed that behavior is correct.

### DIFF
--- a/notebooks/README.md
+++ b/notebooks/README.md
@@ -13,3 +13,10 @@ To get started using notebooks in the devcontainer:
 ## Standards
 
 1. Notebooks should be organized into directories named like the UBC Sailbot repositories they correspond to
+
+# Introducing rosbags
+
+- Rosbags are sqlite db files which contain messages published to a topic that are being recorded. These are not natively supported.
+- By converting these db files into a DataFrame, we can use data to create plots and tables.
+- To start, begin by executing the code cells in `explore_bag_data.ipynb` in order. You can see the DataFrame by running the `df.head()` command in the third cell. These are the messages that are being published to local_path by gps, wind_sensors, etc.
+- By running the 4th code cell, you extract the data from each message into numerical values that are being published to the topic.


src/global_launch/main_launch.py
```diff
@@ -78,7 +80,33 @@ def setup_launch(context: LaunchContext) -> List[IncludeLaunchDescription]:
     """
     mode = LaunchConfiguration("mode").perform(context)
     ros_packages = get_running_ros_packages(mode)
-    return get_include_launch_descriptions(ros_packages)
+    include_launch_descriptions = get_include_launch_descriptions(ros_packages)
+
+    # Specifies the directory that we want the data to be placed in
```
SPDonaghy: comments here are not needed as its pretty clear what's being done

src/global_launch/main_launch.py
```diff
@@ -78,7 +80,33 @@ def setup_launch(context: LaunchContext) -> List[IncludeLaunchDescription]:
...
+    target_dir = "/workspaces/sailbot_workspace/notebooks/local_pathfinding/session_recordings"
+    os.makedirs(target_dir, exist_ok=True)
+
+    # Creates the file name with date & time
```
SPDonaghy: ditto

src/global_launch/main_launch.py
```diff
@@ -78,7 +80,33 @@ def setup_launch(context: LaunchContext) -> List[IncludeLaunchDescription]:
     """
     mode = LaunchConfiguration("mode").perform(context)
     ros_packages = get_running_ros_packages(mode)
-    return get_include_launch_descriptions(ros_packages)
+    include_launch_descriptions = get_include_launch_descriptions(ros_packages)
+
+    # Specifies the directory that we want the data to be placed in
+    target_dir = "/workspaces/sailbot_workspace/notebooks/local_pathfinding/session_recordings"
```
SPDonaghy: can you use the environment variable instead of the hardcoded path please?

```
ros@96135fa2-5ffc-4dee-a6bc-253940e86713:/workspaces/sailbot_workspace$ echo $ROS_WORKSPACE
/workspaces/sailbot_workspace
```

PR #839, opened 2026-03-02
Raghumanimehta/807-partial:Timeout implementation + cleanup of update_if_needed, LocalPath and NodeNavigate

raghumanimehta:
### Description

- Resolves a part of the conditions to switch path issue, not mentioning it here otherwise github would close it. 
- I had intended the whole conditions to switch paths be one PR but it seems making small incremental changes will allow more people to asynchronously work on this eg. @MilcToast worked on the wind thresholding.
- This PR adds a field `path_generation_time` to `LocalPathState` to keep a track of when the path was generated. 
- I also added a function `is_path_expired` in `LocalPath` that checks if a path is expired or not using the field mentioned in the previous point. 
- The funciton has been tested
- I also updated the `LocalPath`'s `update_if_needed` to have a cleaner flow. It still largely follows the old algorithm. 
- I have updated the documentation in `node_navigate.py` and `local_path.py` to reflect that waypoint index is not the targetted waypoint but the waypoint we have already passed. Think of the situation of what happens when OMPL is called. The first item (0 index) is the waypoint at which OMPL was called and the boat is moving away from this waypoint. Following this intuition, the waypoint index is the waypoint that has already been visited. The variable name has been updated to reflect this behaviour. 
- I also added some error checking to ensure that we never access out of bound access in the waypoints array stored  (localPath array from OMPL).

### Verification

- [x] Unit tests 

### Resources 
- The point about timeout in https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1890025723/Conditions+to+Update+Local+Path

### DIFF
--- a/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
+++ b/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
@@ -56,8 +56,8 @@ class BoatState:
                 kinematic data in the relative reference frame, and the second element representing
                 data in the global reference frame, both using SI units.
         """
-        rel_wind_vel = glo_wind_vel - self.global_velocity
-        rel_water_vel = glo_water_vel - self.global_velocity
+        rel_wind_vel = glo_wind_vel[:2] - self.global_velocity[:2]
+        rel_water_vel = glo_water_vel[:2] - self.global_velocity[:2]  # slice into 2d vector
 
         rel_net_force, net_torque = self.__compute_net_force_and_torque(
             rel_wind_vel, rel_water_vel, rudder_angle_deg, trim_tab_angle
--- a/src/boat_simulator/boat_simulator/nodes/sim_visualizer/sim_visualizer_node.py
+++ b/src/boat_simulator/boat_simulator/nodes/sim_visualizer/sim_visualizer_node.py
@@ -1,0 +1,101 @@
+import rclpy
+from rclpy.node import Node
+from custom_interfaces.msg import GPS
+from custom_interfaces.msg import DesiredHeading
+
+import matplotlib
+matplotlib.use("Agg")  # Must be before pyplot import
+import matplotlib.pyplot as plt  # noqa
+import matplotlib.markers as mmarkers  # noqa
+
+import imageio  # noqa
+import io  # noqa
+
+
+class SimVisualizer(Node):
+    def __init__(self):
+        super().__init__("sim_visualizer")
+
+        # GPS Subscription
+        self.create_subscription(GPS, "/mock_gps", self.gps_callback, 10)
+
+        # Latest GPS values
+        self.__lat = None
+        self.__lon = None
+
+        # Heading Subscription
+        self.__heading = 0.0
+        self.create_subscription(DesiredHeading, "/desired_heading", self.heading_callback, 10)
+
+        # Matplotlib setup
+        self.__fig, self.__ax = plt.subplots()
+        self.__ax.set_xlabel("Latitude")
+        self.__ax.set_ylabel("Longitude")
+        self.__ax.set_title("Boat Path")
+
+        self.__x_data = []
+        self.__y_data = []
+        self.__frames = []
+        self.__line, = self.__ax.plot([], [], "-b")
+        self.__dot, = self.__ax.plot([], [], marker="^", color="blue", markersize=12)
+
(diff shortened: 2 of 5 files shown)


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -112,12 +116,27 @@ def __init__(self, parent_logger: RcutilsLogger):
...
+
     @staticmethod
     def calculate_desired_heading_and_wp_index(
-        path: ci.Path, waypoint_index: int, boat_lat_lon: ci.HelperLatLon
+        path: ci.Path, prev_lp_wp_index: int, boat_lat_lon: ci.HelperLatLon
     ):
         """Calculates the desired heading using GEODESIC. Updates the waypoint index (i.e. change
-        the waypoint) if the boat is close enough to the current waypoint.
+        the waypoint) if the boat is close enough to the next waypoint.
+        The caller MUST ensure that prev_lp_wp_index <= len(path.waypoints)
```
SPDonaghy: Should we stick to a standard of having a REQUIRES clause for functions where the caller needs to ensure something?

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -128,51 +147,55 @@ def calculate_desired_heading_and_wp_index(
 
         Returns:
             _type_: _description_
+        Throws:
+            Value error if index out of bounds or path is None
```
SPDonaghy: nit: ValueError

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -300,28 +304,43 @@ def update_if_needed(
...
+            return heading_new_path, new_prev_lp_wp_index
+        else:
+            # this assert will never fail!! Added this here to make Pylance happy
+            assert self.state is not None
+            self.state.update_state(gps, ais_ships, filtered_wind_sensor)
+
+        try:
+            # Same as the previous assert, Python can't statically analyze the fact that
+            # self._ompl_path won't be None for some reason. Removing this assert will lead to
+            # pylance/LSP red squiggles
```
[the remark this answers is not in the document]
raghumanimehta: It does! Thanks

src/local_pathfinding/test/test_ompl_path.py
```diff
@@ -282,25 +282,25 @@ def test_create_space(
...
 )
-def test_get_remaining_cost_partial(fresh_ompl_path, wp_index):
+def test_get_remaining_cost_partial(fresh_ompl_path, target_wp_index):
     waypoints = fresh_ompl_path.get_path().waypoints
-    boat_latlon = waypoints[wp_index]
-    next_wp_latlon = waypoints[wp_index + 1] if wp_index + 1 < len(waypoints) else None
+    current_wp_latlon = waypoints[target_wp_index - 1]
+    next_wp_latlon = waypoints[target_wp_index]
 
     def mid_point(start_latlon: HelperLatLon, end_latlon: HelperLatLon):
```
AMaharaj16: Although this is for testing, it isn't immediately clear what this function does. Add a short docstring explaining what this function does.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -300,30 +323,49 @@ def update_if_needed(
...
+        try:
+            heading_old_path, old_target_lp_wp_index = self.calculate_desired_heading_and_wp_index(
+                self._ompl_path.get_path(), target_lp_wp_index, gps.lat_lon  # type: ignore
+            )
+        except (ValueError, IndexError):
+            return heading_new_path, new_target_lp_wp_index
 
-        heading_diff_old_path = cs.calculate_heading_diff(self.state.heading, heading_old_path)
-        heading_diff_new_path = cs.calculate_heading_diff(self.state.heading, heading_new_path)
+        if self.is_path_expired():
```
AMaharaj16: Isn't this case already covered in `must_change_path()` called above?

src/local_pathfinding/local_pathfinding/ompl_path.py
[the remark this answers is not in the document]
raghumanimehta: Ok, will fix.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: Yes. fixed.

PR #840, opened 2026-03-04
Fix Production API Calls

JordanChen123:
### Description
Production site at https://sailbotpolaris.com was experiencing "Failed to fetch" errors for all API requests.
API services were constructing absolute URLs using the server IP address (http://164.92.120.218:3005/api/gps), which caused mixed content errors when the site runs on HTTPS.
image

### Solution

Use relative URLs in production instead of absolute IP-based URLs.
Created lib/apiUrl.ts that detects hostname - returns /api/* for production (resolves to https://sailbotpolaris.com/api/*) and http://localhost:3005/api/* for development.
also updated all 7 service files and DownloadData.js to use the new getApiUrl() helper.

### Verification
Local development tested and working
Will have to verify on prod, once this is merged to main


raghumanimehta:
Hi @yuekaifred @minjunminji, can we have this merged as well if it's ready?

PR #842, opened 2026-03-05
Bubbles5228/763 simvisualizer

### LINKED ISSUE #763: Visualizer for simulator
### Purpose

We’re currently wrapping up the simulator in #404, but we’re missing a way to actually see what’s happening. Right now, it’s hard to tell if the boat is behaving correctly just by looking at raw logs.

Building a visualizer will let us see the boat's position in real-time, which is going to be huge for verifying that our simulation logic actually makes sense and for catching any weird movement glitches that aren't obvious in the data.

### Description

This feature is about taking the raw numbers from the simulator and turning them into a live path on a graph.
- We need to pull the boat state values directly from the 
- The visualizer should plot the boat's position as it moves. Instead of just a single dot, it should leave a trail so we can see the full path the boat has taken over time. A simple 2D plot (like Matplotlib or similar) that updates as the simulation runs, making it easy to spot oscillations or steering issues at a glance.

### Resources

- https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/boat_simulator/boat_simulator/nodes/data_collection/data_collection_node.py
- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1785790595/Boat+Simulator+Design+Specification

bubbles5228:
### Description

- Resolves #763
- simple plotter that tracks the path of the boat using Matplotlib, and saves the result into a GIF into the sim_visualizer folder using imageio
- run using `ros2 launch boat_simulator main_launch.py enable-mock-data:=true enable-sim-visualizer:=true`


src/boat_simulator/tests/unit/nodes/physics_engine/test_model.py
alberto-escobar: slice these:
```
relative_wind_vel = glo_wind_vel[:2] - input_kinematics.linear_velocity[:2]
relative_water_vel = glo_water_vel[:2] - input_kinematics.linear_velocity[:2]
```

alberto-escobar (approved):
Good job!

PR #844, opened 2026-03-05
Andy g6/website split view

AndyG6:
split screen feature in the layout model added!

this change allows the graphs to be displayed 2 at a time, or 3 at a time or *EVEN 4 IN ONE TIME*

# Comprehensive technical changes (claude summary)

Introduced Layout, LayoutItem, and GraphId types in GraphsTypes.ts. A layout item is either a single GraphId or a GraphId[] (split group)
Renamed graphs.order → graphs.layout in the Redux store and session storage
Added GraphsLayoutHelpers.ts with pure layout manipulation helpers: moveGraph, moveGraphToIndex, extractGraph, and splitGraph
Replaced SortableContext/useSortable with useDraggable + a custom pointermove listener for richer drag intent detection (split vs. reorder vs. extract)

# how to recreate change

-You can reorder the graphs via gap drop zones (highlighted with a blue line indicator)
-Split them alongside an existing item (left/right highlight on hover)
-Can also extract a graph out of a split group back into its own row

image


yuekaifred (commented):
this is cool! when youre dragging over to make a split horizontal graph, what do you think about resizing the other div on drag over? currently we just kinda have a ghost that sits on top of the div were dragging over
image

minjunminji:
I'd consider limiting it to only two graphs per row, ppl don't really need four on one row and it's difficult to read the graphs when they're that narrow. Also prevents confusion with moving around the graphs.

raghumanimehta:
Hi @yuekaifred @AndyG6 @minjunminji , can we have this wrapped up please?

minjunminji (approved):

raghumanimehta (approved):

PR #847, opened 2026-03-07
#825: Tracking Wind data over certain amount of time (seconds)

### LINKED ISSUE #825: PATH: Conditions to switch paths -> track wind data over 10 minutes
### Purpose

We will now work on having the rolling average. The wind data is noisy so averageing the wind data of the previous 30 seconds is important. We will go about implement like this:
- In `node_navigate.py`, we get the wind sensor data. This wind data is **apparent wind**. We will keep track of the number of wind data readings we have received and the rolling average upto this point. The updated average can be computed with this $average := (n \times average + new \ data)/(n + 1)$. Please verify this formula. 
- We will have to do this individually for the direction and speed.
- Once we have this implemented we will convert this to `true_wind` wherever needed. You don't need to worry about this for now. 

### Resources

The following files from `local_pathfinding` are important. 
- `node_navigate.py`
- `local_path.py`
- Ask me (@raghumanimehta) if you are stuck or if something is not clear

MilcToast:
### Description

- Resolves #825 

- added `aw_history` and `aw_avg` as fields in the `LocalPathState` class to store values
- created `update_aw_history` function that adds the newest wind reading to wind_history
- created `_calculate_aw_avg` helper function to update `wind_average`'s speed (kmph) and direction (deg) every time a new wind reading is added in wind_history

#### Behavior
- By Default, when initializing a `LocalPathState`, a wind reading gets added to `aw_history`, such that it should never be empty. If `calculate_aw_avg` IS called with an empty `aw_history`, there is a safeguard that returns a wind value with speed and direction: (0kmph, 0deg)
- To account for the circular nature of angles (i.e. the average of -170 degrees and 170 degrees is 180, and not 0), we use the formula for the circular mean:
image  <br>
where the symbols mean:
ā: Mean angle,
atan2: arctan function that determines the angle from `atan2(y, x)`. It also that takes into account which quadrant the angle belongs to, (ex: (neg, neg) will return angle in 3rd quad and not 1st quadrant), and we perform atan2 on the sum of the sines and the sum of the cosine of every single angle.

Conceptually, we are converting all the angles into a unit vector in cartesian coordinates, summing them all up into a resulting vector, which is then reconverted back to an angle.

### Verification

- tests in `test_local_path.py`

### Resources

https://en.wikipedia.org/wiki/Circular_mean

### DIFF
--- a/src/local_pathfinding/local_pathfinding/ompl_path.py
+++ b/src/local_pathfinding/local_pathfinding/ompl_path.py
@@ -397,8 +397,8 @@ class OMPLPath:
             simple_setup,
             self.state.heading,
             self.state.speed,
-            self.state.wind_direction,
-            self.state.wind_speed,
+            self.state.current_aw.dir_deg,
+            self.state.current_aw.speed_kmph,
         )
 
         simple_setup.setOptimizationObjective(objective)


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -362,3 +371,42 @@ def _update(self, ompl_path: OMPLPath):
...
+        exceeds the max length, the oldest reading is removed.
+
+        Args:
+            current_wind (wcs.Wind): Current wind speed (kmph) and direction (deg)
+        """
+
+        self.wind_history.append(current_wind)
+
+        # Recalculate average wind from history once minimum wind readings reached
+        if len(self.wind_history) == WIND_HISTORY_LEN:
```
raghumanimehta: What about the case when we have < `WIND_HISTORY_LEN`? We just return`None` for now. 

There are two problems: 
- There is no type annotation letting the reader know that there is a possibility of `None` being returned. This means a hard crash if the reader doesn't read the entire function. 
- Logically, it makes little sense to say that the callers of this function won't be able to make decisions unless the length is `WIND_HISTORY_LEN`. I might be missing your thought on how that could be handled, so if you have ideas, please let me know. If not, it is advisable to remove this check all together. Handle the `None` and return the average based on how many values have been queued till the moment.

raghumanimehta (changes requested):

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -96,6 +106,47 @@ def update_state(
...
+            Optional[Wind]: Average wind object, or None if aw_history is not full
+        """
+        if len(self.aw_history) < WIND_HISTORY_LEN:
+            return None
+
+        avg_speed, sin_sum, cos_sum = 0.0, 0.0, 0.0
+
+        for wind in self.aw_history:
+            avg_speed += wind.speed_kmph / len(self.aw_history)
+            # Use circular mean to handle wrap-around
```
raghumanimehta: perhaps a good idea to link the formula here. If you are going to mention what formula, you are using, it will save the reader a lot of time if you also provide the link to the formula.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -96,6 +106,47 @@ def update_state(
         self.wind_speed = filtered_wind_sensor.speed.speed
```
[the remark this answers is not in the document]
raghumanimehta: This will also involve making changes to where `LocalPathState` is referenced.

src/local_pathfinding/test/test_local_path.py
raghumanimehta: Please remove these fields as they're no longer part of LocalPathState

PR #859, opened 2026-03-14
Small finishing touches added to explore_bag_data.ipynb file

### LINKED ISSUE #753: Pathfinding data plotting workflow

ethanhu912:
### Description
- Resolves #753 
- Made some small changes to `explore_bag_data.ipynb`
- Created a helper functions cell and moved all repeated functions to that cell
- Removed the code that opened the same database in each cell (so it's only opened once at the start)
- Standardized all calculations to units in kilometers and seconds

### Verification
- Ran each code cell to ensure the same functionality as before
- Verified that the cell returned values in km/h

### DIFF
--- a/.devcontainer/release/release.Dockerfile
+++ b/.devcontainer/release/release.Dockerfile
@@ -82,6 +82,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
         can-utils \
         vim \
         make \
+        clang \
     && apt-get autoremove -y \
     && apt-get clean -y \
     && rm -rf /var/lib/apt/lists/{apt,dpkg,cache,log} /tmp/* /var/tmp/* \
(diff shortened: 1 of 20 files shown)


SPDonaghy:
Hi Ethan! Sorry cant really review this properly as the diff for an ipynb file is practically unreadable. If everything looks good/works for you we can merge this. The notebook can just serve as a good place to test things out but I wont look at the code in them too closely. We can review the code for the actual script (that presumably does the things the cells in the nb do) a bit more closely once that's ready.

PR #860, opened 2026-03-14
Add clang to release image

digcheese:
### Description

- Add Clang to release image


digcheese:
I forgot that Clang is a large install. We should assess if it really is needed to recompile quick fixes in the moment.

PR #864, opened 2026-03-21
Dynamic Bounding Boxes Added

### LINKED ISSUE #787: AISstream bounding box
currently we set a hardcoded, fixed bounding box to collect aisstreams data from. we want this to dynamically follow the boat.

matthewcflam:
Resolves #787. Added AIS subscription bounds that follow the boat GPS instead of staying hardcoded to Vancouver.

### Four new behaviours controlled in env
1. AISSTREAM_DYNAMIC_BBOX_ENABLED=true
2. AISSTREAM_BBOX_RADIUS_KM=30
3. AISSTREAM_BBOX_MOVE_THRESHOLD_KM=5
4. AISSTREAM_GPS_STALE_MS=43200000

### Logic
- Reads the latest boat GPS from MongoDB, validates it (threshold, stale) and falls back safely to the default bounding box when unavailable.
- Added geospatial helpers (distance + center-to-bbox conversion + coordinate normalization). AKA fancy math
- Writes boat movement, staleness, and ships saved to each snapshot in the terminal.

### Refactored subscription flow
- Subscription payload is now built from the current runtime bounding box state.
- Bounding box updates are applied only when the boat has moved beyond the configured threshold.
- Re-subscription happens on the existing periodic cycle, preserving reconnect behaviour.

## It works for Los Angeles!!!
image


minjunminji:
Remove package-lock.json from the PR, it's auto-generated and isn't needed, also its making the PR diff like a thousand lines

src/website/scripts/aisstream-ingest.js
```diff
@@ -140,6 +196,217 @@ const toNumber = (value) =>
...
+  const deltaLat = toRadians(lat2 - lat1);
+  const deltaLon = toRadians(lon2 - lon1);
+  const a =
+    Math.sin(deltaLat / 2) ** 2 +
+    Math.cos(toRadians(lat1)) * Math.cos(toRadians(lat2)) * Math.sin(deltaLon / 2) ** 2;
+  return 2 * EARTH_RADIUS_KM * Math.asin(Math.sqrt(a));
+};
+
+// Build a single AIS bbox from center point + radius in km.
+const buildBoundingBoxAroundCenter = (latitude, longitude, radiusKm) => {
```
minjunminji, line 229: This can produce a box where west longitude is greater than east longitude when the boat is near ±180 longitude. That creates invalid/ambiguous coverage for many bbox consumers and can silently subscribe to the wrong region

Example: center 179.9, radius 30km -> west normalizes near 179.6, east near -179.8 (west > east).

I think here we can split it into two bounding boxes when crossing the antimeridian (one ending at 180, one starting at -180) instead of returning a single wrapped box.

src/website/.env.development
AndyG6, line 1: why is this stuff in the .env.development

fyang151:
> Remove package-lock.json from the PR, it's auto-generated and isn't needed, also its making the PR diff like a thousand lines

note to self update our gitignore for this

yuekaifred:
> > Remove package-lock.json from the PR, it's auto-generated and isn't needed, also its making the PR diff like a thousand lines
> 
> note to self update our gitignore for this

jk lol we actually should i think

PR #871, opened 2026-04-01
Fix #805: first working draft

### LINKED ISSUE #805: PATH/SIM Integrating
### Purpose
We aim to integrate the pathfinding module with our simulator to leverage the simulator's physics engine. This will allow us to evaluate pathfinding performance over long-range voyages, providing a more realistic assessment than current methods. 

Currently, pathfinding utilizes "mock nodes" (such as a constant module for a mock wind sensor), which are effective for unit testing and short-term behavioral verification. However, these mocks do not sufficiently account for environmental dynamics over time. By using the simulator, we can ensure that generated paths remain viable under realistic physical constraints.

### Description
The implementation will be executed in iterative stages to address the complexity of synchronizing the two systems. The primary goals are:
* **Validation:** Verify the current stability and accuracy of the simulator.
* **Integration:** Run the simulator concurrently with the pathfinding module to ensure data flow.
* **Abstraction:** Develop a toggleable system to switch seamlessly between "PATH mocks" (for lightweight testing) and "SIM mocks" (for high-fidelity physics testing).

We will merge code upon the completion of each stage. This project requires proactive research into our existing physics integration and close coordination with the SIM and Control teams.

### Resources
* **Boat Simulator Codebase:** Primary environment for physics integration.
* **visualizer.py:** For debugging and observing path deviations.
* **@alberto-escobar:** Point of contact for architectural guidance.
* **SIM/Ctrl Team:** For expertise on boat dynamics and sensor data emulation.

jb3982:
### Description

- Resolves #805 

This PR adds SIM + PATH integration for end-to-end testing.

Main changes:
- Added `sim` mode to global launch to run the integrated stack.
- Centralized cross-package coordination in global launch (package launch files remain scoped).
- Updated PATH launch for `sim` mode to run only PATH-specific simulation nodes.

- Fixed `desired_heading` instability in PATH:
  - Root cause: stale `global_waypoint_index` relative to updated `global_path.waypoints`
  - Added index validation and re-sync logic
  - Prevented crash path that caused publisher count to drop to 0

- Updated PATH visualizer payload in sim mode (publishes full `LPathData` instead of minimal production payload).
- Enabled mock land injection for PATH testing in sim mode.
- Added/updated NET sim-mode handling (`SYSTEM_MODE::SIM`, transceiver mode branches).

### Verification

- [x] Launched integrated stack using global launch:
  - `ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py mode:=sim`
- [x] Confirmed `desired_heading` continues publishing (publisher count no longer drops to 0 after a few iterations).
- [x] Confirmed `navigate` no longer crashes from global-waypoint index mismatch.
- [x] Confirmed PATH visualizer receives data and renders path output in sim mode.
- [x] Confirmed `sim`mode NET nodes initialize with sim-compatible mode handling.
- [ ] Run extended-duration test (>= 15 min) to monitor waypoint fluctuation and stability over time.
- [ ] Manually compare calculated values against visualizer output to validate consistency.

### Resources

- Issue: #805
Screenshot 2026-03-31 at 5 42 21 AM
Screenshot 2026-03-31 at 5 42 28 AM
Screenshot 2026-03-31 at 5 42 37 AM

### DIFF
--- a/src/boat_simulator/boat_simulator/common/constants.py
+++ b/src/boat_simulator/boat_simulator/common/constants.py
@@ -20,7 +20,7 @@ class Actions:
 
 @dataclass
 class LowLevelControlSubscriptionTopics:
-    GPS: str = "mock_gps"
+    GPS: str = "gps"  # added for #805
 
 
 @dataclass
@@ -31,7 +31,7 @@ class PhysicsEngineSubscriptionTopics:
 
 @dataclass
 class PhysicsEnginePublisherTopics:
-    GPS: str = "mock_gps"
+    GPS: str = "gps"  # added for #805
     KINEMATICS: str = "mock_kinematics"
     WIND_SENSORS: str = "mock_wind_sensors"
 
--- a/src/boat_simulator/boat_simulator/common/constants.py
+++ b/src/boat_simulator/boat_simulator/common/constants.py
@@ -20,7 +20,7 @@ class Actions:
 
 @dataclass
 class LowLevelControlSubscriptionTopics:
-    GPS: str = "gps"  # added for #805
+    GPS: str = "gps"
 
 
 @dataclass
@@ -31,7 +31,7 @@ class PhysicsEngineSubscriptionTopics:
 
 @dataclass
 class PhysicsEnginePublisherTopics:
-    GPS: str = "gps"  # added for #805
+    GPS: str = "gps"
     KINEMATICS: str = "mock_kinematics"
     WIND_SENSORS: str = "mock_wind_sensors"
 
--- a/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
+++ b/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
@@ -56,8 +56,8 @@ class BoatState:
                 kinematic data in the relative reference frame, and the second element representing
                 data in the global reference frame, both using SI units.
         """
-        rel_wind_vel = glo_wind_vel[:2] - self.global_velocity[:2]
-        rel_water_vel = glo_water_vel[:2] - self.global_velocity[:2]  # slice into 2d vector
+        rel_wind_vel = glo_wind_vel - self.global_velocity[:2]
+        rel_water_vel = glo_water_vel - self.global_velocity[:2]
 
         rel_net_force, net_torque = self.__compute_net_force_and_torque(
             rel_wind_vel, rel_water_vel, rudder_angle_deg, trim_tab_angle
--- a/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
+++ b/src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
@@ -56,8 +56,8 @@ class BoatState:
                 kinematic data in the relative reference frame, and the second element representing
                 data in the global reference frame, both using SI units.
         """
-        rel_wind_vel = glo_wind_vel - self.global_velocity[:2]
-        rel_water_vel = glo_water_vel - self.global_velocity[:2]
+        rel_wind_vel = glo_wind_vel[:2] - self.global_velocity[:2]
+        rel_water_vel = glo_water_vel[:2] - self.global_velocity[:2]  # slice into 2d vector
 
         rel_net_force, net_torque = self.__compute_net_force_and_torque(
             rel_wind_vel, rel_water_vel, rudder_angle_deg, trim_tab_angle
--- a/src/custom_interfaces/msg/DesiredHeading.msg
+++ b/src/custom_interfaces/msg/DesiredHeading.msg
@@ -1,2 +1,3 @@
 HelperHeading heading
 uint8 steering
+bool sail  # for issue# 805
--- a/src/global_launch/main_launch.py
+++ b/src/global_launch/main_launch.py
@@ -19,6 +19,7 @@ from launch.substitutions import LaunchConfiguration
 
 PRODUCTION_ROS_PACKAGES = ["controller", "local_pathfinding", "network_systems"]
 DEVELOPMENT_ROS_PACKAGES = ["controller", "boat_simulator", "local_pathfinding", "network_systems"]
+SIM_ROS_PACKAGES = ["controller", "boat_simulator", "local_pathfinding", "network_systems"]
 
 # Global launch arguments and constants.
 ROS_PACKAGES_DIR = os.path.join(
@@ -43,9 +44,9 @@ GLOBAL_LAUNCH_ARGUMENTS = [
     DeclareLaunchArgument(
         name="mode",
         default_value="development",
-        choices=["production", "development"],
-        description="System mode. Decides whether the system is ran with development or production"
-        + " interfaces",
+        choices=["production", "development", "sim"],  # added "sim" for issue#805
+        description="System mode. Decides whether the system is ran with development, production,"
+        + " or sim interfaces",
     ),
     DeclareLaunchArgument(
         name="test_plan",
@@ -160,8 +161,11 @@ def get_running_ros_packages(mode: str) -> List[str]:
             return PRODUCTION_ROS_PACKAGES
         case "development":
             return DEVELOPMENT_ROS_PACKAGES
+        case "sim":
+            return SIM_ROS_PACKAGES
         case _:
-            raise ValueError("Invalid launch mode. Must be one of 'production', 'development'.")
+            raise ValueError(
+                "Invalid launch mode. Must be one of 'production', 'development','sim'.")
 
 
 def get_include_launch_descriptions(ros_package_list: List[str]) -> List[IncludeLaunchDescription]:
--- a/src/global_launch/main_launch.py
+++ b/src/global_launch/main_launch.py
@@ -44,7 +44,7 @@ GLOBAL_LAUNCH_ARGUMENTS = [
     DeclareLaunchArgument(
         name="mode",
         default_value="development",
-        choices=["production", "development", "sim"],  # added "sim" for issue#805
+        choices=["production", "development", "sim"],
         description="System mode. Decides whether the system is ran with development, production,"
         + " or sim interfaces",
     ),
--- a/src/local_pathfinding/launch/main_launch.py
+++ b/src/local_pathfinding/launch/main_launch.py
@@ -4,13 +4,12 @@ import importlib
 import os
 from typing import List, Tuple
 
-from launch_ros.actions import Node
-
 from launch.actions import DeclareLaunchArgument, OpaqueFunction
 from launch.launch_context import LaunchContext
 from launch.launch_description import LaunchDescription
 from launch.some_substitutions_type import SomeSubstitutionsType
 from launch.substitutions import LaunchConfiguration
+from launch_ros.actions import Node
 
 # Local launch arguments and constants
 PACKAGE_NAME = "local_pathfinding"
@@ -67,13 +66,19 @@ def setup_launch(context: LaunchContext) -> List[Node]:
     """
     mode = LaunchConfiguration("mode").perform(context)
 
-    launch_description_entities = []
+    launch_description_entities: List[Node] = []
     launch_description_entities.append(get_navigate_node_description(context))
     if mode == "development":
         launch_description_entities.append(get_mock_global_path_node_description(context))
         launch_description_entities.append(get_mock_wind_sensor_node_description(context))
         launch_description_entities.append(get_mock_ais_node_description(context))
         launch_description_entities.append(get_mock_gps_node_description(context))
+        launch_description_entities.append(get_navigate_observer_node_description(context))
+    elif mode == "sim":
+        # Mock nodes for global_path and ais_ships.
+        # System-level package orchestration is handled by global_launch.
+        launch_description_entities.append(get_mock_global_path_node_description(context))
+        launch_description_entities.append(get_mock_ais_node_description(context))
         launch_description_entities.append(get_navigate_observer_node_description(context))
     return launch_description_entities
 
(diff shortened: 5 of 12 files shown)


src/boat_simulator/boat_simulator/common/constants.py
```diff
@@ -20,7 +20,7 @@ class Actions:
 
 @dataclass
 class LowLevelControlSubscriptionTopics:
-    GPS: str = "mock_gps"
+    GPS: str = "gps"  # added for #805
```
raghumanimehta: Please remove these comments. Why were they added in the first place?

src/boat_simulator/boat_simulator/common/constants.py
```diff
@@ -31,7 +31,7 @@ class PhysicsEngineSubscriptionTopics:
 
 @dataclass
 class PhysicsEnginePublisherTopics:
-    GPS: str = "mock_gps"
+    GPS: str = "gps"  # added for #805
```
raghumanimehta: comment violation.

src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
```diff
@@ -56,8 +56,8 @@ def step(
                 kinematic data in the relative reference frame, and the second element representing
                 data in the global reference frame, both using SI units.
         """
-        rel_wind_vel = glo_wind_vel[:2] - self.global_velocity[:2]
-        rel_water_vel = glo_water_vel[:2] - self.global_velocity[:2]  # slice into 2d vector
+        rel_wind_vel = glo_wind_vel - self.global_velocity[:2]
```
raghumanimehta: What was the bug here? Why was this changed?

src/custom_interfaces/msg/DesiredHeading.msg
```diff
@@ -1,2 +1,3 @@
 HelperHeading heading
 uint8 steering
+bool sail  # for issue# 805
```
raghumanimehta, line 3: comment violation? Why was this added here?

src/global_launch/main_launch.py
```diff
@@ -43,9 +44,9 @@
     DeclareLaunchArgument(
         name="mode",
         default_value="development",
-        choices=["production", "development"],
-        description="System mode. Decides whether the system is ran with development or production"
-        + " interfaces",
+        choices=["production", "development", "sim"],  # added "sim" for issue#805
```
raghumanimehta: Please remove these comments "for issue ..".

raghumanimehta (changes requested):
Please remove all the unnecessary comments. 

I have not looked through the entire PR. Use comments only to explain something non-obvious.

SPDonaghy:
Hey just getting into this now. I like the intent of making a lot of solid improvements, but I think its a few too many ideas for one PR and this probably should have been split into 2-3 separate PRs because it's faster to review/approve them when each one is doing less.

Also, wondering why we need a _sim_ mode if the _development_ mode already ran the system with mocks?

src/global_launch/main_launch.py
SPDonaghy: and what is sim mode for, why not reuse development

src/local_pathfinding/launch/main_launch.py
```diff
@@ -67,14 +66,20 @@ def setup_launch(context: LaunchContext) -> List[Node]:
     """
     mode = LaunchConfiguration("mode").perform(context)
 
-    launch_description_entities = []
+    launch_description_entities: List[Node] = []
```
SPDonaghy: we don't typically use type hints for lvalues like this in the codebase. Did you feel it was necessary?

src/local_pathfinding/launch/main_launch.py
```diff
@@ -67,14 +66,20 @@ def setup_launch(context: LaunchContext) -> List[Node]:
...
     launch_description_entities.append(get_navigate_node_description(context))
     if mode == "development":
         launch_description_entities.append(get_mock_global_path_node_description(context))
         launch_description_entities.append(get_mock_wind_sensor_node_description(context))
         launch_description_entities.append(get_mock_ais_node_description(context))
         launch_description_entities.append(get_mock_gps_node_description(context))
         launch_description_entities.append(get_navigate_observer_node_description(context))
+    elif mode == "sim":
+        # Mock nodes for global_path and ais_ships.
```
SPDonaghy: I think both of these comments can be removed because the first one is evident from the lines of code below and the second one _should_ be known to the reader, or if not known, this doesn't really feel like the appropriate spot to teach the reader that.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -159,8 +154,20 @@ def global_path_callback(self, msg: ci.Path):
             f"Received data from {self.global_path_sub.topic}: {gp.path_to_dict(msg)}"
         )
         self.global_path = msg
-        if self.saved_target_global_waypoint is None:
-            self.saved_target_global_waypoint = self.global_path.waypoints[-1]
+        if not self._has_valid_global_path():
```
SPDonaghy, line 162: I think there's enough going on here with the path and index stuff that is not relevant to integrating pathfinding with simulator that I suggest this all be moved to a separate PR that also includes new unit tests to assert that the bug these changes are attempting to remove is actually fixed.

src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
bubbles5228: I believe the `glo_wind_vel` vector is already 2D; don't really think it's a bug but I also don't think it does anything

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
```diff
@@ -112,9 +112,12 @@ def update_queue(self):
             )
             return
 
-        self.queue.put(vz.VisualizerState(msgs=self.msgs))
-        self.get_logger().info(f"sent new visualizer state with {len(self.msgs)} messages.")
-        self.get_logger().info(f"queue: {self.queue.qsize()}")
+        try:
```
SPDonaghy: Did you experience an exception occur at some point from here? I hadn't heard of this happening before.

src/network_systems/lib/cmn_hdrs/shared_constants.h
```diff
@@ -16,6 +16,7 @@ namespace SYSTEM_MODE
 {
 static const std::string PROD     = "production";
 static const std::string DEV      = "development";
+static const std::string SIM      = "sim";  //  added for issue#805
```
SPDonaghy: rm comment

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -65,6 +65,8 @@ class CanTransceiverIntf : public NetNode
                     RCLCPP_ERROR(this->get_logger(), "%s", err.what());
                     throw err;
                 }
+            } else if (mode == SYSTEM_MODE::SIM) { // added for issue#805
```
SPDonaghy: rm comment

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -86,49 +88,52 @@ class CanTransceiverIntf : public NetNode
...
-
-            can_trns_->registerCanCbs(canCbs);
-
-            sail_cmd_sub_ = this->create_subscription<msg::SailCmd>(
-              ros_topics::SAIL_CMD, QUEUE_SIZE, [this](msg::SailCmd sail_cmd_) { subSailCmdCb(sail_cmd_); });
-            desired_heading_sub_ = this->create_subscription<msg::DesiredHeading>(
-              ros_topics::DESIRED_HEADING, QUEUE_SIZE,
-              [this](msg::DesiredHeading desired_heading_) { subDesiredHeadingCb(desired_heading_); });
-
+            //  edited for issue#805
```
SPDonaghy: rm comment

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -86,49 +88,52 @@ class CanTransceiverIntf : public NetNode
...
+
+                can_trns_->registerCanCbs(canCbs);
+
+                sail_cmd_sub_ = this->create_subscription<msg::SailCmd>(
+                    ros_topics::SAIL_CMD, QUEUE_SIZE, [this](msg::SailCmd sail_cmd_) { subSailCmdCb(sail_cmd_); });
+                desired_heading_sub_ = this->create_subscription<msg::DesiredHeading>(
+                    ros_topics::DESIRED_HEADING, QUEUE_SIZE,
+                    [this](msg::DesiredHeading desired_heading_) { subDesiredHeadingCb(desired_heading_); });
+            }
+            // till here, if statement
```
SPDonaghy: sorry, not sure what this means

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -146,6 +151,11 @@ class CanTransceiverIntf : public NetNode
                     publishBoatSimInput(boat_sim_input_msg_);
                     // Add any other necessary looping callbacks
                 });
+            } else if (mode == SYSTEM_MODE::SIM) { //  added forissue#805
```
SPDonaghy: rm comment

src/boat_simulator/boat_simulator/common/constants.py
jb3982: These comments were temporary markers that I added to track modified lines while working across multiple large files (made it easier for me to look for the edits I made). They weren’t intended to be part of the final PR. I’ll clean them up in the next commit.

src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
jb3982: This change was made while getting the simulator running and applying the required adjustments to ensure SIM could run correctly.

src/custom_interfaces/msg/DesiredHeading.msg
jb3982: This field was added to resolve a mismatch in the `DesiredHeading` topic that was preventing the system from running correctly (PATH was pausing due to inconsistent fields). So, a temporary `sail` field with a default True value was introduced to allow the system to run as a stub. The inline comment was only meant as a temporary note to indicate that this change was made for issue #805. I’ll remove that.

src/global_launch/main_launch.py
jb3982: I’ll remove the “for issue #805” and other similar comments in the next commit.
And I’ve addressed the reasoning for “sim” mode in the general comment's reply below.

jb3982:
> Hey just getting into this now. I like the intent of making a lot of solid improvements, but I think its a few too many ideas for one PR and this probably should have been split into 2-3 separate PRs because it's faster to review/approve them when each one is doing less.
> 
> Also, wondering why we need a _sim_ mode if the _development_ mode already ran the system with mocks?

First of all thanks for the feedback. I agree this PR is larger than ideal.
I initially planned to split out the SIM/NET-related changes, but since I had a working integrated version and the deadline was close, I pushed it as a single PR. I’ll keep PRs more focused going forward.

Regarding `sim` mode:
My goal was to get the whole system running (PATH, SIM, CTRL, NET) through a single launch command. Earlier this required manually starting each module in separate terminals, which made iteration and restarts a hassel.
In this setup, PATH runs a subset of nodes (mock global path, mock AIS and navigate observer) while receiving data from SIM instead of the usual `development` mode inputs.
Initially this logic was placed inside PATH’s main_launch.py, but that meant we were launching the whole system from a sub module, which looked like a bad design choice. So, I moved it to the global launch.
The sim mode was introduced to keep this behavior separate from `development`, since `development` mode assumes a different set of node configuration.

src/local_pathfinding/launch/main_launch.py
jb3982: This wasn’t necessary, I added it while resolving a type-related error during my earlier changes, but it’s no longer needed after re-structuring the launch logic. And I missed reverting it before submitting the PR. I’ll remove it.

src/local_pathfinding/launch/main_launch.py
jb3982: These comments were temporary notes I added while working through the changes to keep track of my progress. I’ll remove them.

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: This change was tied to the issue where only the `desired_heading` topic was not being published. The root cause was that the size of the `global_path` waypoints list was not stable across updates.
Since the path is regenerated based on GPS updates and interpolation, the number of waypoints can vary between callbacks (due to distance approximation and recomputation). We were implicitly assuming stable indexing, which led to invalid indices and caused `node_navigate` to stop progressing, preventing `desired_heading` from being published.
These changes were made so that waypoint indexing and target selection remain consistent even when the waypoint list size changes, allowing `desired_heading` to be published reliably.
I agree this is a broader issue within the SIM–PATH integration and should be backed by unit tests. I’ve verified the fix locally with consistent publishing, and I’ll run longer tests to further validate the behaviour.

src/local_pathfinding/local_pathfinding/node_navigate_observer.py
jb3982: I did encounter error around this path while debugging the issue with `desired_heading` not being published, so I added the `try-except` temporarily to guard against failures from `self.queue.put(...)`, if any, during testing.
Since the root issue was elsewhere, this is no longer needed and I’ll have it removed.

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
jb3982: That was a temporary marker I added while restructuring this block. I’ll have this and other redundant comments removed in the next commit.

src/boat_simulator/boat_simulator/nodes/physics_engine/model.py
bubbles5228: The current code works for me -- was something not working?

SPDonaghy:
Thanks for the response @jb3982 let me know when this is ready for another pass. Make sure to resolve any conversations if you have addressed the feedback in them.

Can you please separate the logic for global path waypoint tracking into a separate pr? Thanks

src/boat_simulator/boat_simulator/common/constants.py
raghumanimehta: Git is used prceisely for tracking diffs.

PR #875, opened 2026-04-04
Potential Fix for Send Check Signal Fail

Apps247:
- Changed `clearSerialBuffer` call position in `send` to match that of `debugSend`

- Increased waiting time in select spots

- Need to test with hardware to see if issue is resolved


raghumanimehta:
> * Changed `clearSerialBuffer` call position in `send` to match that of `debugSend`
> * Increased waiting time in select spots
> * Need to test with hardware to see if issue is resolved

What hardware do you need? Also @Apps247 please link the issue this PR resolves.

lross03:
No longer needed

PR #876, opened 2026-04-04
Added Standalone Remote Transceiver Server

### LINKED ISSUE #851: See if remote transceiver can be removed from ROS
### Purpose

Deployment on the droplet would be simpler if it could be run separately

### Description

See if remote transceiver could me modified to run without ROS

Apps247:
Resolves #851

The code for the server is in `src/network_systems/projects/remote_server_standalone`. This code does not interact with the other files in the project.

It comes with its own setup scripts, build script, and run script

To run, move this folder to an independent location on a device (eg: Digital Ocean Droplet), run setup, run build, then run the server.

We could move this to a separate repository


lross03 (approved):
good

Issue #829, opened 2026-02-28
PATH: Conditions to switch paths - Segment Deviation Box


AMaharaj16:
During PATH/ELEC meeting, the suggestion that if the boat strays some distance from its current segment in the path, pathfinding should be rerun. To satisfy this criteria, add a box around each segment that spans the whole segment with a width roughly 10-12% (should be held as a constant) of the length of the segment. Create a function in `local_path` that returns true if the length of the line perpendicular to the segment to the boat is larger than this 10-12% threshold and add a robust testing suite. For now, call the function `exceeded_segment_deviation()`.

PR #878, opened 2026-05-12
Raghumanimehta/stopping-the-boat-conditions-to-switch-paths

### LINKED ISSUE #866: PATH: Conditions to Switch Paths
### Purpose

Now that we have the wind tracking over 30 seconds feature implemented. We have to add this check to `must_change` method. You will have to look out for cases where `is_significant_change` returns `None`. 

Once you have done that, come up with Tests that check if the path in fact has changed. This part is quite hard. So before you start doing this, have a discussion with @AMaharaj16, @SPDonaghy, or @raghumanimehta (me). I would suggest using `Mocks` for this as they allow us to statically analyze behaviour. 

Manually verify that the deque that was added as part of `is_significant_change` is tracking the data properly. I know @MilcToast has added tests, but it is a good idea to see it running in production to ensure that it is actually accumulating data properly. 

### Resources

- https://github.com/UBCSailbot/sailbot_workspace/blob/0b35a48f20a33d48694aea9b4a3588d8769de518/src/local_pathfinding/local_pathfinding/local_path.py#L290
- https://github.com/UBCSailbot/sailbot_workspace/blob/0b35a48f20a33d48694aea9b4a3588d8769de518/src/local_pathfinding/local_pathfinding/local_path.py#L382

- https://docs.python.org/3/library/unittest.mock.html

raghumanimehta:
### Description
Resolves #866 and #765. 
This PR fully implements and tests the conditions to switch paths described here: https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1890025723/Conditions+to+Update+Local+Path

I have also revamped the error handling of Pathfinding in `node_navigate.py` and `local_path.py`. There were uncaught errors being thrown that would crash the system; all the exceptions thrown in these files are now being handled gracefully, when the errors are unrecoverable, I set the `sail` field of `desired_heading` to `False`. 

I realize one of the changes mentioned in the above confluence link is not yet implemented, that is for another PR to do. The issue for that is here: https://github.com/UBCSailbot/sailbot_workspace/issues/829. 

### Note: 
This PR is not that big, about 500 LOC is tests. Please verify the tests properly as well. The reason I am resolving multiple things in this PR is beacuse I felt the systems are coupled enough that handling them in one PR is effecient both for me and the reviewer. 

### Verification

- [x] Unit tests 
- [x] compiles and runs fine


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -305,12 +315,16 @@ def must_change_path(self, received_new_global_waypoint: bool) -> tuple[bool, st
...
+        ):
+            return True, "Significant wind change"
         if self._target_lp_wp_index < 1:
             return True, f"Target waypoint index too low: {self._target_lp_wp_index}"
         if self._target_lp_wp_index >= len(self.path.waypoints):
             return (
                 True,
-                f"Target waypoint index out of bounds: {self._target_lp_wp_index} >= {len(self.path.waypoints)}", # noqa
+                f"Target waypoint index out of bounds: {self._target_lp_wp_index} >= {len(self.path.waypoints)}",  # noqa
             )
```
FireBoyAJ24, line 406: Could we use Enums to capture the meanings of different possible causes of changing paths?

src/local_pathfinding/local_pathfinding/local_path.py
FireBoyAJ24: Could we reduce the number of parameters passed into this function? Could we use the LocalPathState class for GPS, AIS Ship, global path and other attributes?

FireBoyAJ24 (changes requested):
Just some possible refactoring suggestions.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: Yes, we can. However, I feel that woudl be an overkill. The only for the reason being returned is to print that on the console. That's the only place we use reason. I would agree on changing this immediately to Enums if we had to use the Enum further in the execution.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: We can. We will have to create the LocalPathState object in `node_navigate.py` and pass that object to the update_if_needed function. For the current design to work, we will have to make `Sailbot` own `LocalPathState`, we make updates to the state in `node_navigate.py` itself and then pass it to the update_if_needed function. My concern regarding this would be time. I am not sure how worthwhile this change is.

raghumanimehta: Never mind this, in this day and age of LLMs, such a refactor is a matter of minutes. I hope it doesn't break anything else though.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -305,142 +339,110 @@ def must_change_path(self, received_new_global_waypoint: bool) -> tuple[bool, st
             return True, "Path intersects collision zone"
         if self.is_path_expired():
             return True, "Path has expired (TTL exceeded)"
+        if new_aw is not None and self.is_significant_wind_change(
+            new_aw, self.state.current_aw
```
AMaharaj16: `self.state.current_aw` is the wind object that came in the prior reading from `filtered_wind_sensor`. This object is assigned in `update_state()` in `LocalPathState` which we call if we **are not** changing paths in `update_if_needed()`.

The issue with this is it will require two consecutive wind readings to be different by some threshold to switch paths (bad), rather than the new wind object differing from the wind object created at path generation by some threshold (good)

We should probably create a new field associated with the wind value at path generation.

AMaharaj16: Another thing here is you are passing apparent wind into `is_significant_wind_change()` but the function is implemented and documented to take in true wind. Good chance to use `boat_to_global_coordinate()` in `wind_coord_systems.py`.

src/local_pathfinding/test/test_local_path.py
```diff
@@ -61,6 +61,171 @@ def create_test_local_path_for_in_collision_zone(
...
+                    HelperLatLon(latitude=0.0, longitude=0.1),
+                    HelperLatLon(latitude=0.0, longitude=0.0),
+                ]
+            ),
+            1,
+            HelperLatLon(latitude=0.0, longitude=0.09999),
+            -90.0,
+            2,
+        ),
+    ],
```
FireBoyAJ24, line 355: An additional test case could be adding 8 waypoints and then choosing the target waypoint is 2 or 3 and it may switch paths and go to 1.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -305,142 +339,110 @@ def must_change_path(self, received_new_global_waypoint: bool) -> tuple[bool, st
...
                 - Updated waypoint index
-            The method decides whether to return the heading for new path or old path
+            The method decides whether to return the heading for a newly generated path or the
+            existing path.
+
+        Raises:
+            PathNotFoundError: If a required path is unavailable or a replacement path cannot be
+                generated.
+            ValueError: If the LocalPathState is passed invalid values.
+            IndexError: If the target waypoint is reached and no next local waypoint exists.
```
AMaharaj16, line 449: Should look into how we want to handle the divide by 0 case for `is_significant_wind_change()` function call:
`speed_change_ratio = abs(prev_tw_speed_kmph - current_tw_speed_kmph) / prev_tw_speed_kmph` (Line 314)

Although it very likely won't happen, we should try to handle it.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -305,142 +339,110 @@ def must_change_path(self, received_new_global_waypoint: bool) -> tuple[bool, st
...
+                    land_multi_polygon=inputs.land_multi_polygon,
+                )
+                if new_ompl_path.solved:
+                    break
+                else:
+                    tries += 1
+
+            if not new_ompl_path.solved:
+                raise PathNotFoundError("Old Path must change and new path couldn't be solved" +
+                                        f"within {MAX_OMPL_PATH_GEN_TRIES}")
```
AMaharaj16: nit: add space before within

src/local_pathfinding/test/test_local_path.py
```diff
@@ -61,6 +61,171 @@ def create_test_local_path_for_in_collision_zone(
...
+                waypoints=[
+                    HelperLatLon(latitude=0.0, longitude=0.1),
+                    HelperLatLon(latitude=0.0, longitude=0.0),
+                ]
+            ),
+            1,
+            HelperLatLon(latitude=0.0, longitude=0.00001),
+            IndexError,
+        ),
+    ],
```
FireBoyAJ24, line 420: We should add a ValueError

FireBoyAJ24 (changes requested):

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: There is actually a bigger bug around this part that I missed. I will fix this.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: We wish to compare with the rolling average, not the value that was computed at the time of path generation. 
The documentation mentions that the `is_significant_wind` works with any wind type.

src/local_pathfinding/test/test_local_path.py
raghumanimehta: added 2 more cases

src/local_pathfinding/test/test_local_path.py
raghumanimehta: The function `calculate_desired_heading_and_waypoint_index` does not throw `ValueError`.

FireBoyAJ24 (approved):

src/local_pathfinding/local_pathfinding/node_navigate.py
FireBoyAJ24: There is a unit mismatch here

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -249,198 +287,177 @@ def in_collision_zone(self):
...
+        previous_dir_deg = previous_wind_data.dir_deg
 
-        current_tw_speed_kmph = new_tw_data.speed_kmph
-        current_tw_dir_deg = new_tw_data.dir_deg
+        current_speed_kmph = new_wind_data.speed_kmph
+        current_dir_deg = new_wind_data.dir_deg
 
-        speed_change_ratio = abs(prev_tw_speed_kmph - current_tw_speed_kmph) / prev_tw_speed_kmph
-        dir_change = abs(cs.bound_to_180(current_tw_dir_deg - prev_tw_dir_deg))
+        speed_change_ratio = abs(previous_speed_kmph - current_speed_kmph) / previous_speed_kmph
```
FireBoyAJ24: A division by zero could occur if the previous_speed_kmph is 0,

FireBoyAJ24 (changes requested):

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -249,198 +287,177 @@ def in_collision_zone(self):
...
-            return True, "Path intersects collision zone"
+            return MustChangeReason(True, "Path intersects collision zone")
         if self.is_path_expired():
-            return True, "Path has expired (TTL exceeded)"
+            return MustChangeReason(True, "Path has expired (TTL exceeded)")
+        if (
+            (new_aw is not None)
+            and (self.state.aw_avg is not None)  # must ensure that LocalPathState.aw_avg
+            # is not None; this happens when len(LocalPathState.aw_history < WIND_HISTORY_LEN)
+            and self.is_significant_wind_change(new_aw, self.state.aw_avg)
```
AMaharaj16: This seems to be comparing the current rolling average with the new apparent wind vector to see if it warrants a path switch. Aren't we supposed to be comparing the current rolling average (as tw) with the true wind vector at path generation time? The path we are currently on is based on one wind vector, if we drift from that wind vector too much (rolling average), we should switch the path since the current path is no longer optimal given the new wind direction.

raghumanimehta: `new_aw` is not a vector; it's a `Wind` object. One latest wind value only. 
The comparison function works fine as long as both the winds are in the same unit format. Comparison between apparent wind and apparent wind average is fine. I have addressed the confusion in the documentation update of `is_significant_change`.

AMaharaj16: I think I didn't make my point clear. We aren't supposed to call `is_significant_wind_change()` on the new wind object and the current rolling average, which is the current implementation you have here. Instead, it should be comparing the **wind object at path generation** and the current rolling average. That's how we'd know if the current path is no longer valid: if the wind has changed from what wind we generated the path from. This is even highlighted in the docstring for `is_significant_wind_change()`:

`new_tw_data (Wind): Current wind speed/direction that may require a change in path` -> Should be rolling average really.
`previous_tw_data (Wind): Wind speed/direction when the current path was generated` -> Current path generated.

Does that make sense or am I missing something here?

raghumanimehta: That logically doesn't make sense to track a rolling average to compare an updated rolling average with `Wind` object at **path generation time**. The whole point of a rolling average is to avoid noisy data ruining our computations. 

Consider this: 
- at path generation time, the wind immediately spiked 
- all the following winds are relatively smoother and not as high as the initial value. 
In this scenario, we will never know which is the culprit. Therefore we track a rolling average to get a trend and not rely on single data point values. The new value we compare is what we are tracking against, and to compensate that there could be another spike or dip, we have thresholds in `is_significant_wind_change`.

AMaharaj16: I get the point about rolling average, that's why we added it. The rolling average can be seen as basically the current state of the wind without any noisy data interfering. In that case, why would we compare the rolling average with the new wind object? That would make it so that if we see even one new wind object that is an outlier, we will switch paths, which we **don't want**. Instead, we should compare with the wind object at path generation. 

In the scenario you mentioned, if the wind spikes at path generation time, then our path is built based on that wind object. Since it's an outlier wind reading, we probably don't want to use that path, and we won't since the rolling average would quickly drift away from it as you suggested. This scenario would work out just fine with this implementation.

FireBoyAJ24: Another consideration is that the Wind Sensor data is filtered in the Network system with a rolling average. So the chance that there is an outlier is further mitigated from the source.

FireBoyAJ24: I agree with Aayush over this issue. We should only compare the rolling average of the wind to the wind at the time of path generation. Comparing the rolling-average wind to the current wind object means that a gradual change in wind direction from 30 to 180 degrees can occur without triggering a significant wind change.

We should meet up and discuss this.

raghumanimehta: One data point as baseline is a bad idea. 
If you are going to go for one data point as base line, it should be rolling average at t (assuming it exists, there are cases where it doesn't) and rolling average at t + t' (assuming this exists as well). I agree to this. 
I don't agree to using one data point `Wind` object at path generation time as baseline. That would be way more brittle.

FireBoyAJ24: Actually, the wind at the path generation would be the rolling average data point. When t = 0, the rolling average will start out with 1 or 2 data points combined. This is indeed brittle, and the path generated from it would be brittle. As the rolling average of the wind object improves and shows a sizable difference at t = 5 s, then the path_gen_wind_object = rolling_average_at_t_equal_5s.

This will make it less brittle over time and ensure that if the path generated by the wind object (containing the prior rolling average) is significantly different, it will change accordingly, rather than based on the prior rolling average.

raghumanimehta: https://github.com/UBCSailbot/sailbot_workspace/pull/878/changes/BASE..17748c4687a2e9ac382a7fc88d3b365070a53df2#diff-817f393727b15a82f5ce8b13d386725a2c1c6ce81f6db5b6165c568e3c75d37aL106
`self.current_aw` is one data point. We shouldn't use this is what I am saying.

FireBoyAJ24: Indeed. We are suggesting changing the implementation.

What we can do for now is to create an issue for this and merge this PR. After this on-water testing, we can use a better implementation for the next on-water test.

raghumanimehta: There is another bug with error handling, some of the errors are not handled properly and the container for the deployment is already built so this PR is not going on board anyway for this on-water test.

raghumanimehta:
@AMaharaj16 @FireBoyAJ24 Ready for another review. Made a lot of changes. 
- now properly handles the errors that are thrown. 
- also updated the wind handling stuff

The code is a bit spaghetti because of the new changes.

src/local_pathfinding/local_pathfinding/local_path.py
FireBoyAJ24: So if prev_speed_kmph is 0.0, would we always change paths? Could this cause the desired heading to change multiple times, as the paths would frequently change?

Should we have a case where the number of paths changes frequently (update_if_needed is true every 1 Hz) and causes the desired heading to change significantly, then we hold the previous desired heading until the next update_if_needed is similar to it. This suggestion is not great, since we cannot truly determine whether the desired extreme heading is correct without context about the environment.

In a nutshell, do we want to treat the zero error as an error or as a needed switch path (which can cause frequent switch path triggers)?

Line handling the significant wind change result: https://github.com/UBCSailbot/sailbot_workspace/pull/878/changes#diff-817f393727b15a82f5ce8b13d386725a2c1c6ce81f6db5b6165c568e3c75d37aR396

FireBoyAJ24 (changes requested):

src/local_pathfinding/local_pathfinding/node_navigate.py
FireBoyAJ24: Still needs to be clarified

raghumanimehta: should be resolved now

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: I don't think we should have a ratio in the first place. 
Wind speed increase from 0.01 to 0.05 is not significant. 
Wind speed increase from 1 to 5 is way more significant. 
The ratio loses this information. I have asked Emma and Michael, what the threshold be for this.

raghumanimehta: https://ubcsailbot.slack.com/archives/C0ADGH53B0T/p1779945885478369
Should be fixed now.

FireBoyAJ24 (dismissed):

src/local_pathfinding/local_pathfinding/local_path.py
FireBoyAJ24: > I get the point about rolling average, that's why we added it. The rolling average can be seen as basically the current state of the wind without any noisy data interfering. In that case, why would we compare the rolling average with the new wind object? That would make it so that if we see even one new wind object that is an outlier, we will switch paths, which we **don't want**. Instead, we should compare with the wind object at path generation.
> 
> In the scenario you mentioned, if the wind spikes at path generation time, then our path is built based on that wind object. Since it's an outlier wind reading, we probably don't want to use that path, and we won't since the rolling average would quickly drift away from it as you suggested. This scenario would work out just fine with this implementation.

We should make an issue on this @AMaharaj16.

FireBoyAJ24:
> I think I didn't make my point clear. We aren't supposed to call `is_significant_wind_change()` on the new wind object and the current rolling average, which is the current implementation you have here. Instead, it should be comparing the **wind object at path generation** and the current rolling average. That's how we'd know if the current path is no longer valid: if the wind has changed from what wind we generated the path from. This is even highlighted in the docstring for `is_significant_wind_change()`:
> 
> `new_tw_data (Wind): Current wind speed/direction that may require a change in path` -> Should be rolling average really. `previous_tw_data (Wind): Wind speed/direction when the current path was generated` -> Current path generated.
> 
> Does that make sense or am I missing something here?

I have approved the work for now, as it looks good. We will need to change the wind implementation as discussed above.

raghumanimehta:
I have already changed the implementation of wind. What else needs to be changed?

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: This should be fixed. I have implemented an entirely rethought wind system. What do you think is missing?

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -243,204 +309,239 @@ def in_collision_zone(self):
...
-        ais_ships: ci.AISShips,
-        global_path: ci.Path,
+        inputs: LocalPathInputs,
         target_lp_wp_index: int,
         received_new_global_waypoint: bool,
-        target_global_waypoint: ci.HelperLatLon,
-        filtered_wind_sensor: ci.WindSensor,
-        planner: str,
-        land_multi_polygon: MultiPolygon = None,
     ) -> tuple[Optional[float], Optional[int]]:
```
AMaharaj16: Changes this output signature according to line 442.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -33,22 +130,24 @@ class LocalPathState:
...
-        global_path (List[Tuple[float, float]]): Path to the destination that Sailbot is
-                                                 navigating along.
-        target_global_waypoint (ci.HelperLatLon): The global waypoint that we are heading towards.
+        global_path (ci.Path): Path to the destination that Sailbot is navigating along.
+        reference_latlon (ci.HelperLatLon): The global waypoint that Sailbot is heading toward.
             The global waypoint is the same as the reference latlon.
-        wind_speed (float): Wind speed.
-        wind_direction (int): Wind direction.
+        aw_speed_kmph (float): Apparent wind speed from the filtered wind sensor, in km/h.
+        aw_dir_boat_coord_deg (int): Apparent wind direction from the filtered wind sensor, in boat
```
AMaharaj16: shouldn't these two above fields be removed as they're replaced by `current_aw`?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -290,18 +297,24 @@ def get_desired_heading(self) -> float:
...
+                    target_global_waypoint=self.saved_target_global_waypoint,
+                    filtered_wind_sensor=self.filtered_wind_sensor,
+                    planner=self.planner,
+                    land_multi_polygon=self.land_multi_polygon,
+                ),
+                target_lp_wp_index=self.target_lp_wp_index,
+                received_new_global_waypoint=received_new_global_waypoint,
+            )
+            return desired_heading, True
+        except PathNotFoundError:
```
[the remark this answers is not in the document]
AMaharaj16: @raghumanimehta, pls address this. seems like an actual concern.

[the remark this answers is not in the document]
AMaharaj16: oh looks like you added some error handling in `update_if_needed()`, this is fully resolved then i assume?

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: Yes.

AMaharaj16 (approved):
looks good! good work @raghumanimehta

PR #881, opened 2026-05-16
#862: Handling Errors Raised in obstacles.py

### LINKED ISSUE #862: PATH: `update_collision_zone` throws`ValueError`. Make changes to handle it gracefully.
Checkout `obstacles.py`. `update_collision_zone` throws `ValueError` which is not handled anywhere in the codebase.

MilcToast:
### Description

- Resolves #862 

#### Changes Made
- For the uninitialized collision zone error (line 62 in `obstacles.py`):
  - This error path appears unreachable. update_collision_zone is called in each child constructor, so the collision zone should always be initialized before is_valid is called.
    - `boat.update_collision_zone` always initializes it
    - `land.update_collision_zone` already raises its own error on failure before this check can be reached. (if this error in this function is not raised, then the collision zone will not be `None`)
- For the uninitialized state space error (line 153 in `obstacles.py`):
  - In `ompl_path.py`, `_init_simple_setup` ensures a non-empty state_space, and `init_obstacles` converts it to a non-`None` `state_space_latlon` before passing it into Land's constructor. The constructor then calls `update_collision_zone`, at which point state_space_latlon should not be `None`.
- For the AIS ship id mismatch error (line 233 in `obstacles.py`):
  -  I believe this error can be removed since the mismatch case appears unreachable. The only call site is in `ompl_path.py`, where `existing_boat` is retrieved using `ship.id`, so the subsequent check is effectively comparing the boat against itself.
  - Removed test cases that asserted these errors were raised, as the errors have been removed.

### Verification

- Ran full simulation; no crashes or unexpected behavior observed after removal of error paths
- Confirmed all obstacle types (boat, land) initialize collision zones before is_valid (in non test codes)
- Manually traced ompl_path.py → init_obstacles → Land initialization path


MilcToast:
We don’t necessarily need to remove these raised errors. Another option would be to keep the checks and handle them differently instead. For example,we could skip obstacle initialization if the condition is detected, or have is_valid return False when encountering an uninitialized collision zone rather than raising the error. 

I went with removal here because I believe that these cases are unreachable, but happy to adjust if we want to preserve explicit safety guards.

raghumanimehta (commented):
I took a brief look and it looks good to me but I will reason this thoroughly tomorrow before merging this in.

FireBoyAJ24 (commented):
> We don’t necessarily need to remove these raised errors. Another option would be to keep the checks and handle them differently instead. For example,we could skip obstacle initialization if the condition is detected, or have is_valid return False when encountering an uninitialized collision zone rather than raising the error.
> 
> I went with removal here because I believe that these cases are unreachable, but happy to adjust if we want to preserve explicit safety guards.

I like the idea of adding a ROS warning log here, as you suggested. If the obstacle is not initialized, it would be nice to skip the validity check and path generation, as you stated above.

raghumanimehta:
> > We don’t necessarily need to remove these raised errors. Another option would be to keep the checks and handle them differently instead. For example,we could skip obstacle initialization if the condition is detected, or have is_valid return False when encountering an uninitialized collision zone rather than raising the error.
> > I went with removal here because I believe that these cases are unreachable, but happy to adjust if we want to preserve explicit safety guards.
> 
> I like the idea of adding a ROS warning log here, as you suggested. If the obstacle is not initialized, it would be nice to skip the validity check and path generation, as you stated above.

@MilcToast I agree with @FireBoyAJ24 here. I would suggest adding logs; I will reason about the code and will let you know if we should do anything besides just logs. You should add the logs asap, ensure the logs are detailed and distinguishable.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -53,13 +57,14 @@ def is_valid(self, point: cs.XY) -> bool:
...
-            RuntimeError: If the collision zone has not yet been initialized.
+            bool: If the collision zone has not yet been initialized defaults to False.
+                True if the point is not within the obstacle's collision zone, false otherwise.
         """
         if self.collision_zone is None:
-            raise RuntimeError("Collision zone has not been initialized")
+            _LOGGER.warning(
+                f"collision_zone not initialized for obstacle {type(self)}; treating as invalid"
+            )
+            return False
```
MilcToast: I'm really not sure whether to treat a point in the state for an obstacle with no collision zone to be valid or invalid. If we default to `False` for safety, we end up invalidating the whole map and making it impossible to find a path. If we default to `True`, we end up risking colliding into an obstacle.

MilcToast: Maybe we could have a circular buffer around the uninitialized obstacles?

raghumanimehta: this makes sense to me. i dont think there is a need for circular buffer. @FireBoyAJ24 @MilcToast @rnar17 thoughts??

FireBoyAJ24: The circular buffer makes sense. Having a reasonable buffer with a reasonable, arbitrary number is fine as long as it is neither too extreme nor too minute.

raghumanimehta: The problem with assigning a point a buffer is that we might end up with all the points being tiny collision zones. Also, I don't believe this branch will be triggered often; the collision zones are almost always not None. I didn't notice it much.

MilcToast: The issue is moreso that if you default the return to False (aka the boat is colliding) when a collision zone is uninitalized, a single obstacle without a collision zone results in all possible positions/states of the boat being considered as invalid. (The uninitialized obstacle claims to be everywhere at the same time)

Do you think it would be better to treat the state as valid (so True) when there's no collision zone then since the branch won't be triggered often?

raghumanimehta: I think returning True makes more sense. If the collision zone is `Land`, then it could be so big that a circular threshold won't cut it. If the collision zone is `AIS` then we won't have projection. A circular threshold is only slightly better than returning `True`. 

Ideally, this class should be reworked such that this kind of stuff doesn't happen, but we can't do that. For now returning True makes the most sense to me. I am willing to risk it because the circular threshold is arbitrary and I don't want to worry about another threshold. 

@FireBoyAJ24 @AMaharaj16

AMaharaj16: I agree with leaving it to return true. Regarding the circular collision zone, the two mains reasons we could add it is for robustness and observability. However, even if we implement this, it isn't really solving the root problem. I don't think it's really necessary to add the circular collision zone.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -53,13 +57,14 @@ def is_valid(self, point: cs.XY) -> bool:
             point (ci.HelperLatLon): Point representing the state point to be checked.
 
         Returns:
-            bool: True if the point is not within the obstacle's collision zone, false otherwise.
-
-        Raises:
-            RuntimeError: If the collision zone has not yet been initialized.
+            bool: If the collision zone has not yet been initialized defaults to False.
```
AMaharaj16: Defaults to `True` now.

AMaharaj16 (approved):
looks good!

raghumanimehta (approved):

PR #889, opened 2026-05-21
Added ROS2 Log Folder and Documentation in Global Launch

### LINKED ISSUE #886: Record ROS logs into a file
### Purpose

Currently, ROS logs are stored in a log directory and split into multiple nodes' .txt files. The issue is that it is hard to tell which file contains the logs we want. In addition, we may not even find the appropriate ROS logs for whatever reason.

### Description

To make debugging easier, we can use a simple `tee` command to record ROS log output and use it during on-water and simulator testing. This is in the hopes of making our lives of debugging easier.
Example of recording the ROS logs in a file command:

```
ros2 launch my_package my_launch.launch.py 2>&1 | tee launch_output.log
```

To accomplish this, we will need to add a parameter to allow us to activate debug mode when launching the ROS nodes and add the `tee` after each ROS launch command.

FireBoyAJ24:
### Description
- Resolves #886
- Documents the log recording workflow for on-water testing, including how to
  capture ROS2 logs to a file using `2>&1 | tee` and the `log_level` launch argument
- Adds a `.gitignore` glob (`combined_log*.txt`) to exclude all log file variants
  from being committed

### Verification
- [x] Verified `log_level:=debug` is a valid launch argument in `main_launch.py`
- [x] Confirmed the full launch command captures both stdout and stderr to file
- [x] Confirmed `combined_log*.txt` glob covers `combined_log.txt`, `combined_log_clean.txt`, numbered variants, etc.


src/global_launch/README.md
```diff
@@ -4,6 +4,10 @@
 - Running `ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py` in the terminal will begin the entire system.
 - You can utilize different launch arguments to perform different tasks when
   launching.
+  <!-- markdownlint-disable MD013 -->
+    - For example, we can look at all the debug statements for all the nodes when we use `loge_level:=debug` as our launch argument.
```
AMaharaj16: Did you mean `log_level:=debug`?

raghumanimehta:
```suggestion
    - For example, we can look at all the debug statements for all the nodes when we use `log_level:=debug` as our launch argument.
```

src/global_launch/README.md
```diff
@@ -14,3 +18,12 @@
...
 - Usually, you wouldn't want to push rosbag recordings to main, as they take up a large amount of storage. Thus, the session_recordings folder is ignored when pushing. If you create a new folder, make sure you add it to `gitignore` unless you do want to push to main.
 <!-- markdownlint-enable MD013 -->
+
+## Recording the logs for future analysis
+
+<!-- markdownlint-disable MD013 -->
+- When we want to store our logs in a separate file, then we can use the following add command,
+`2>&1 | tee src/global_launch/voyage_log/combined_log.txt`. The logs will be stored in the `combined_log*.txt` file. Where * can be any suffix like `combined_log_1.txt`, `combined_log_clean.txt`, etc.
+- The `esc[32m` is present in the `[DEBUG]` statement, as that is used for colouring that line green.
+- A on-water testing launch would look something like this, `ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py record:=true mode:="production" 2>&1 | tee src/global_launch/voyage_log/combined_log.txt`
```
raghumanimehta: @digcheese Can we add this button as well, please?

raghumanimehta (changes requested):

src/global_launch/README.md
FireBoyAJ24: Fixed that now. Thanks

src/global_launch/README.md
FireBoyAJ24: Curious, what kind of button?

jeffrey-xue: We have buttons on the gui that copy terminal commands to your clipboard. Is it reasonable to just add this command for log storing to the button we already have?

raghumanimehta: Depends, can we just click on it once and not worry about typing the command when we are on the chase boat? If yes, sure.

raghumanimehta: @FireBoyAJ24, it is painful to type commands on the boat as per Michael (I have no reason to doubt that). Since, he's going to be on the chase boat, and the computer is not the best and things will be shaky, we want the experience to be as good as possible for him (or anyone else). 

Polaris debug tools have a gui, that allows us to quicky execute common commands.

jeffrey-xue: @raghumanimehta @FireBoyAJ24 Regarding #900 and specially the task about creating combined-log files with a time stamp, I think it makes the most sense to modify the command here. 

That modification would be just changing the path to `src/global_launch/voyage_log/combined_log_$(date %D-%T).txt`

May I or one of you make a commit to this pr for consistency with issue #900 ?

FireBoyAJ24: In my latest PR for mock ais. I have reduced the size of the global launch rod launch. I can investigate reducing the size further by using a bash script instead to make it easier

raghumanimehta (dismissed):

src/global_launch/README.md
FireBoyAJ24: @digcheese, could you make the changes as you're familiar with what needs to be consistent with your previous PRs?

AMaharaj16: Why don't we have `voyage_log/` in the repo? Without it tee has nowhere to write and the command fails. Am I missing something?

raghumanimehta: If you create `voyage_log`but don't have any files in it and push it then git doesn't track it for some reason. So you need to add at least one file for git to track it.

FireBoyAJ24: 🫡

PR #896, opened 2026-05-23
#829: Segment Deviation Box for Conditions to Switch Path

### LINKED ISSUE #829: PATH: Conditions to switch paths - Segment Deviation Box

MilcToast:
### Description

- Resolves #829

### Changes Made

- Added `LocalPath.exceeded_segment_deviation`
  - Computes geodesic segment length between consecutive waypoints
  - Converts coordinates into a local XY frame for geometric calculations
  - Uses orthogonal projection to determine the closest point on the segment
  - Clamps projections to ensure distance is measured against the finite segment
  - Added `GPS_POSITION_ERROR_KM` constant as an additive **floor to the deviation threshold**, also preventing false positives on short segments where `SEGMENT_DEVIATION_THRESHOLD * segment_length_km` alone would be negligibly small
  - Compares perpendicular distance against:
    - `SEGMENT_DEVIATION_THRESHOLD * segment_length_km + GPS_POSITION_ERROR_KM`
  - Raises `IndexError` when provided `target_lp_wp_index` is out of bounds
  
- Added unit tests covering:
  - Threshold edge cases
  - Projection clamping before and after the segment
  - Diagonal path handling
  - Identical waypoint (zero-length segment) handling
  - Near-identical waypoint (microscopic threshold) handling
  - Very short segment behaviour where `GPS_POSITION_ERROR_KM` dominates the threshold
  - Boats positioned directly on the segment 
  - Index out of bounds error handling

### Verification

- Ran all new unit tests
- Verified threshold edge cases behave as expected
- Verified projection clamping for boats positioned before and after the segment
- Verified diagonal path handling
- Verified degenerate waypoint (no segment) handling safely returns `False`
- Verified near-identical waypoints do not produce false positives due to microscopic threshold
- Verified IndexError is raised for out-of-bounds waypoint indices


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,64 @@ def is_significant_wind_change(
...
+        segment.
+
+        Args:
+            path (ci.Path): Array of waypoints
+            target_lp_wp_index (int): 0-based array index of the local waypoint Polaris is
+                currently heading toward. This should start at index 1 because index 0 is the
+                OMPL start waypoint near the boat.
+            boat_lat_lon (ci.HelperLatLon): boat coordinates
+        """
+        prev_wp = path.waypoints[target_lp_wp_index - 1]
```
raghumanimehta, line 399: Should have index checks here? For example, if somehow we pass something like 0, that would be bad. This shouldn't happen but a check would be nice. would love to hear other thoughts @AMaharaj16 @FireBoyAJ24

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,64 @@ def is_significant_wind_change(
...
+            prev_wp.longitude,
+            prev_wp.latitude,
+            target_wp.longitude,
+            target_wp.latitude,
+        )
+        segment_length_km = cs.meters_to_km(segment_length_m)
+        max_deviation_km = segment_length_km * SEGMENT_DEVIATION_THRESHOLD
+
+        # Build a local XY frame with prev_wp as origin.
+        prev_xy = cs.XY(0.0, 0.0)
```
raghumanimehta: nit: specify that `xy` is in `km` by appending `km` to the end of the variable name.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,64 @@ def is_significant_wind_change(
...
+        boat_xy = cs.latlon_to_xy(prev_wp, boat_lat_lon)
+
+        dx = target_xy.x - prev_xy.x
+        dy = target_xy.y - prev_xy.y
+        segment_len_sq = dx * dx + dy * dy
+
+        if segment_len_sq == 0.0:
+            return False  # Waypoints are the same, no deviation possible
+
+        # Orthogonal projection of A to Boat onto the line defined by prev and target to find the
```
raghumanimehta: what is "A"?

raghumanimehta (changes requested):
Some superficial comments; haven't reviewed the math yet. But good job! Really happy with the pace.

src/local_pathfinding/local_pathfinding/local_path.py
MilcToast: Whoops, it was a typo from when I was doing the math on paper, I've removed it now.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: @codex Thoughts? No code changes.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,66 @@ def is_significant_wind_change(
...
+        target_wp = path.waypoints[target_lp_wp_index]
+
+        _, _, segment_length_m = cs.GEODESIC.inv(
+            prev_wp.longitude,
+            prev_wp.latitude,
+            target_wp.longitude,
+            target_wp.latitude,
+        )
+        segment_length_km = cs.meters_to_km(segment_length_m)
+        max_deviation_km = segment_length_km * SEGMENT_DEVIATION_THRESHOLD
```
raghumanimehta: Please account for GPS noise here (can add a constant here perhaps as a minimum)

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -11,12 +11,13 @@
...
 import local_pathfinding.coord_systems as cs
 import local_pathfinding.obstacles as ob
-from local_pathfinding.wind_coord_systems import Wind
 from local_pathfinding.ompl_path import OMPLPath
+from local_pathfinding.wind_coord_systems import Wind
 
 WIND_SPEED_CHANGE_THRESH_PROP = 0.3
 WIND_DIRECTION_CHANGE_THRESH_DEG = 10
 WIND_HISTORY_LEN = 30
+SEGMENT_DEVIATION_THRESHOLD = 0.1
```
raghumanimehta: @FireBoyAJ24 Does this value match what you experienced on the water?

src/local_pathfinding/test/test_local_path.py
```diff
@@ -509,6 +507,231 @@ def test_aw_avg_updates_with_new_readings():
...
+            1,
+            _boat_position_for_segment_deviation(
+                EXCEEDED_SEGMENT_DEVIATION_BASIC_PATH,
+                1,
+                1 + lp.SEGMENT_DEVIATION_THRESHOLD * 2,
+                0.5,
+            ),
+            True,
+        ),
+        (
```
raghumanimehta, line 1033: What is this test case? I like you added comments for other ones. Please add for this one as well.

src/local_pathfinding/test/test_local_path.py
```diff
@@ -509,6 +507,231 @@ def test_aw_avg_updates_with_new_readings():
...
+            _boat_position_for_segment_deviation(
+                EXCEEDED_SEGMENT_DEVIATION_SHORT_PATH,
+                1,
+                -lp.SEGMENT_DEVIATION_THRESHOLD * 0.5,
+                0.5,
+            ),
+            False,
+        ),
+        (
+            EXCEEDED_SEGMENT_DEVIATION_SHORT_PATH,
```
raghumanimehta, line 1060: no comment here as well

src/local_pathfinding/test/test_local_path.py
```diff
@@ -509,6 +507,231 @@ def test_aw_avg_updates_with_new_readings():
...
+                EXCEEDED_SEGMENT_DEVIATION_BASIC_PATH,
+                1,
+                0.3,
+                1.0,
+            ),
+            False,
+        ),
+    ],
+)
+def test_exceeded_segment_deviation(path, target_index, boat_position, expected):
```
raghumanimehta, line 1167: Can we add some more extreme cases? 

Some suggestions from AI, i like them all: 
- Invalid index case: target_lp_wp_index = 0 should not silently use waypoints[-1]. Test that it raises IndexError, matching calculate_desired_heading_and_wp_index()’s contract.

- Out-of-range index case: target_lp_wp_index == len(path.waypoints) should raise cleanly too. It currently raises naturally, but an explicit guard/test would make the API contract clear.

- Very short segment case: e.g. a 1 m segment with the boat 0.5-1 m away. This is the issue’s “short segments may false-positive” concern. If the PR adds a minimum deviation threshold, this test should assert the boat does not exceed until it passes that minimum.

- Tiny but nonzero segment case: waypoints almost identical, but not exactly identical. The current segment_len_sq == 0.0 only handles exact zero; a near-zero segment can produce a microscopic threshold.

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: You should throw an `IndexError` if we get `target_lp_wp_index == 0`. Add a descriptive log statement as well usign `warn` level.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,66 @@ def is_significant_wind_change(
...
+        dx_km = target_xy_km.x - prev_xy_km.x
+        dy_km = target_xy_km.y - prev_xy_km.y
+        segment_len_sq = dx_km * dx_km + dy_km * dy_km
+
+        if segment_len_sq == 0.0:
+            return False  # Waypoints are the same, no deviation possible
+
+        # Orthogonal projection of the boat onto the line defined by prev and target
+        # to find the closest point on the line to the boat
+        t = ((boat_xy_km.x - prev_xy_km.x) *
```
raghumanimehta: `t` is not a good variable name.

raghumanimehta: Should be more descriptive.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,66 @@ def is_significant_wind_change(
...
+        max_deviation_km = segment_length_km * SEGMENT_DEVIATION_THRESHOLD
+
+        # Build a local XY frame with prev_wp as origin.
+        prev_xy_km = cs.XY(0.0, 0.0)
+        target_xy_km = cs.latlon_to_xy(prev_wp, target_wp)
+        boat_xy_km = cs.latlon_to_xy(prev_wp, boat_lat_lon)
+
+        dx_km = target_xy_km.x - prev_xy_km.x
+        dy_km = target_xy_km.y - prev_xy_km.y
+        segment_len_sq = dx_km * dx_km + dy_km * dy_km
```
raghumanimehta:
```suggestion
        segment_len_sq = dx_km**2 + dy_km**2
```

raghumanimehta (changes requested):
Minor changes required. Thanks!

src/local_pathfinding/local_pathfinding/local_path.py
raghumanimehta: https://ubcsailbot.slack.com/archives/C0739N8SHV1/p1779998763721379

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -287,6 +288,66 @@ def is_significant_wind_change(
...
+        # Orthogonal projection of the boat onto the line defined by prev and target
+        # to find the closest point on the line to the boat
+        t = ((boat_xy_km.x - prev_xy_km.x) *
+             dx_km + (boat_xy_km.y - prev_xy_km.y) * dy_km) / segment_len_sq
+
+        # Ensures that the closest point is within the line segment defined by prev and target
+        t = max(0.0, min(1.0, t))
+
+        closest_x_km = prev_xy_km.x + t * dx_km
+        closest_y_km = prev_xy_km.y + t * dy_km
```
FireBoyAJ24: `prev_xy_km.x` and `prev_xy_km.y` is always set to 0.0 because of the above line, `prev_xy_km = cs.XY(0.0, 0.0)`. This makes closest_x_km and closest_y_km sound redundant and confuses me. As on line 346, `distance_to_segment_km = math.hypot(boat_xy_km.x - closest_x_km,
                                            boat_xy_km.y - closest_y_km)` the boat_xy_km.x should have been the x component vector of `prev_lp - boat_xy` and the same for the y component. Since the previous local waypoint is set to the origin, this issue would not surface until the origin was not the previous local waypoint.

There is a lot of simplification and combining on this line `t = ((boat_xy_km.x - prev_xy_km.x) * dx_km + (boat_xy_km.y - prev_xy_km.y) * dy_km) / segment_len_sq`. `t` is the scalar projection value for the vector between `prev_lp - boat_xy` onto `prev_lp - target_lp`.

We can instead utilize numpy, as the above calculations are derived from the vector projection formula, as shown below,

Image

Resource: https://personal.math.ubc.ca/~tbjw/ila/projections.html

src/local_pathfinding/local_pathfinding/local_path.py
FireBoyAJ24: Regarding this value, it is effectively the percentage of the distance between two waypoints. The fun aspect of this is that the distance between the previous and target waypoints can vary, so the max segment deviation threshold distance can change accordingly.

This is **not at all a problem**. However, if the distance between local waypoints was ~200 meters, then the max deviation would be 20 meters. If the boat is in irons for 20-30 minutes, then it can drift by 50 meters or more. I can't say for when it is not in irons, as the boat hasn't yet performed in the optimal/normal conditions. I recommend increasing the threshold to 30%-50%, so we would definitely know the boat is mainly drifting.

FireBoyAJ24 (changes requested):

FireBoyAJ24 (dismissed):

raghumanimehta (dismissed):

Issue #900, opened 2026-05-28
DevOps: Update the `Dockerfile`


raghumanimehta:
### Summary

Update the Dockerfiles and setup scripts so the software setup is reliable, reproducible, and easier to debug.

### What we want

- The logs folder should be created (if it doesn't exist).
- The logs file should be created with the name (if it doesn't exist).
- The path for the file is: `$ROS_WORKSPACE/src/global_launch/voyage_logs/combined_log_{timestamp}.txt`
- Separate(ish) task: We should have a way to enter the container without running all the ROS nodes.

### Requirements

1. Create logs folder and log file
   - The setup/run script should create the required logs folder.
   - The script should create the required logs file using the expected timestamp format.
   - The script should verify that the logs path exists before writing to it.

2. Enter container without launching ROS nodes
   - Add a documented way to enter the container without automatically starting all ROS nodes.
   - This is needed for debugging, inspecting files, running manual commands, and extracting data without running the full system.

3. Bonus thing: 
   -  Figure out how to extract data without launching every node in the container. This is related to point 2. However, it would be great if we have a script that would copy the files to outside the container  / outside both container + ssh sesion. 

### Acceptance Criteria

- [x] Running the setup script creates the required logs folder.
- [x] Running the setup script creates the required logs file.
- [ ] The script verifies that the logs path exists before writing logs.
- [x] There is a documented command or script option to enter the container without launching all ROS nodes.
- [x] The Docker setup works from a clean checkout.

### Assignees

DevOps: @JeffreyXue, @RaghumaniMehta, @AkshanjayKompelli

jeffrey-xue:
I think it's redundant for `setup.sh` to verify the creation of the combined_log directory. If the creation of the directory fails, then setup.sh will terminate early. Which for our prod. uses, will then terminate the docker build process for the release-image. 

Verifying whether or not the path exists before writing is a bigger feature that I think would be better integrated as a polaris-gui feature (i.e. the button to start the software would run a script beforehand that does a variety of system checks).  I think in any case, the path being created properly but then disappearing on the day-of should be an extreme edge case.

PR #904, opened 2026-05-29
Create global_launch/combined_logs directory for production logging

digcheese:
### Description

- Addresses requirement 1 of #900 by creating the necessary directory when `setup.sh` gets ran. i.e. during the release-image build process.  

### Verification

- Built the release image locally and verified that the build does not fail and the directory exists.
- Tested that `tee` can create files which, in combination with an appropriate modification to #889, will add timestamps to each log file.


scripts/setup.sh
```diff
@@ -34,4 +34,7 @@ sudo apt-get update
 rosdep update --rosdistro "$ROS_DISTRO"
 rosdep install --from-paths src --ignore-src --rosdistro "$ROS_DISTRO" $DEP_FLAGS
 
+# Create logging folder for all ROS logs.
+sudo mkdir $ROS_WORKSPACE/src/global_launch/voyage_log
```
raghumanimehta: You should also have the command to have the log file created like log_file_{timestam}.txt here. You can hide the directory creation behind an if

raghumanimehta (changes requested):
minor change

scripts/setup.sh
jeffrey-xue: I don't think we should create the log file here with the timestamp. It would make writing to the log_file more cumbersome because`setup.sh` is only ran during the build process and not when we start all the ROS nodes. i.e. the timestamp wouldn't be accurate to our use case.

Also `tee` only errors because the directory didn't exist. `tee` can create the log file itself.

raghumanimehta: tee won't capture the timestamp itself though, I would prefer to have a separate one line script that does tee and ros launch with production (name it somethign like `owt_launch.sh`

FireBoyAJ24: We can simply run `ros2 launch my_pkg my_launch.py 2>&1 | tee voyage_log/combined_ros_log_$(date +%Y%m%d_%H%M%S).txt`. This will create the file and print stdout to it.

FireBoyAJ24 (approved):

PR #910, opened 2026-06-01
Added Mock AIS for on Water Testing

FireBoyAJ24:
### Description

Add mock ais in production since there are no ships near the boat during on-water testing. Using a mock AIS node for pathfinding allows us to test Polaris with different ghost boats nearby during on-water testing. In addition, I made the `global_launch` into a proper ROS2 package, as that was causing the launch commands to be too long (ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py ,,, to ros2 launch global_launch ...)

### Verification

- [x] Run `ros2 launch global_launch main_launch.py record:=true mode:=production` where `on_water_mock_ais` is False in `global_launch/config/globals.yaml`. There should be no mock AIS node running from pathfinding.
- [x] Run `ros2 launch global_launch main_launch.py record:=true mode:=production` where `on_water_mock_ais` is True in `global_launch/config/globals.yaml`. There should be a mock AIS node running from pathfinding.
- [x]  Run `ros2 launch global_launch main_launch.py record:=true mode:=development` where `on_water_mock_ais` is True in `global_launch/config/globals.yaml` and `on_water_test_plan` is set to any other on-water test plan you want in `src/local_pathfinding/test_plans`. It is a success as long as the test plan exists in the correct directory and produces the correct result by checking the ais_ships message.


.github/workflows/docs.yml
```diff
@@ -88,13 +89,13 @@ jobs:
       - name: Build docs
         if: github.event_name == 'push' || github.event.action != 'closed'
         shell: bash
-        run: docker compose -f .devcontainer/docker-compose.yml exec -T docs /bin/sh -c "mkdocs build"
+        run: docker compose -f .devcontainer/docker-compose.yml -f .devcontainer/docs/docker-compose.docs.yml exec -T docs /bin/sh -c "mkdocs build"
```
raghumanimehta, line 92: What's this change for?

.github/workflows/docs.yml
```diff
@@ -104,7 +105,7 @@ jobs:
         if: github.event_name == 'pull_request' && github.event.action != 'closed'
         shell: bash
         run: |
-          docker compose -f .devcontainer/docker-compose.yml exec -T docs /bin/sh -c "\
+          docker compose -f .devcontainer/docker-compose.yml -f .devcontainer/docs/docker-compose.docs.yml exec -T docs /bin/sh -c "\
```
raghumanimehta, line 108: Can you explain this as well?

docs/current/sailbot_workspace/reference/launch_files.md
```diff
@@ -170,7 +170,7 @@ launch command.
     of arguments:
 
     ```sh
-    ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py -s
+    ros2 launch global_launch main_launch.py -s
```
raghumanimehta, line 173: Can we have both the one that is removed and this one? Both are correct right?

docs/current/sailbot_workspace/reference/launch_files.md
```diff
@@ -211,7 +211,7 @@ launch command.
 
     <!-- markdownlint-disable MD013 -->
     ```sh
-    ros2 launch $ROS_WORKSPACE/src/global_launch/main_launch.py enable_sim_multithreading:=true
+    ros2 launch global_launch main_launch.py enable_sim_multithreading:=true
```
raghumanimehta, line 214: ditto as above

src/global_launch/setup.py
```diff
@@ -0,0 +1,22 @@
+from os.path import join
```
raghumanimehta, line 1: Can you explain what this file does? i don't know what this is doing?

src/local_pathfinding/test_plans/on_water_mock_ais.yaml
```diff
@@ -0,0 +1,6 @@
+ais:
+  - {id: 1, lat: 49.01261318191024, lon: -135.65593922412154, heading: 90.0, speed: 20.0, rot: 30, width: 1.0, height: 1.0}
```
raghumanimehta: I trust this has been verified as close to the place we will be doing owt

src/network_systems/projects/remote_server_standalone/README.md
```diff
@@ -1,6 +1,7 @@
 # Remote Server Standalone
 
-This folder contains a standalone version of the remote_transceiver server and sailbot_db, decoupled from ROS. It is ready to be copied into a new repository and built independently.
+This folder contains a standalone version of the remote_transceiver server and sailbot_db,
+decoupled from ROS. It is ready to be copied into a new repository and built independently.
```
raghumanimehta, line 4: weird change here?

src/network_systems/projects/remote_server_standalone/README.md
```diff
@@ -11,8 +12,14 @@ This folder contains a standalone version of the remote_transceiver server and s
 
 ## Build
 
-Use the provided CMakeLists.txt to build the project. All ROS dependencies have been removed.
+Use the provided CMakeLists.txt to build the project. All ROS dependencies
+have been removed.
 
-Note that a clean build (removing all dependencies etc.) takes a long time and should be avoided. There's a chance the script won't fetch all packages required and manual installations and setup may be required, especially for the MongoDB CXX driver.
+Note that a clean build (removing all dependencies etc.) takes a long time
```
raghumanimehta, line 18: ditto

raghumanimehta (changes requested):
Looks good, just want to clarify things before merging.

.github/workflows/docs.yml
FireBoyAJ24: These changes were needed as I was getting a lot of docs workflow errors

docs/current/sailbot_workspace/reference/launch_files.md
FireBoyAJ24: You can run both

FireBoyAJ24: Do you want me to add it back with this as an alternative in the launch file?

docs/current/sailbot_workspace/reference/launch_files.md
FireBoyAJ24: When launching the boat, just typing global_launch instead of `$ROS_WORKSPACE` shenanigans would save us some pain.

src/global_launch/setup.py
FireBoyAJ24: It is a super important file for all ROS packages. It is the place where the ROS Middleware accesses to identify where the key data files (launch files, source code, miscellaneous code used by the source code) are when building the package.

https://docs.ros.org/en/foxy/How-To-Guides/Developing-a-ROS-2-Package.html

src/local_pathfinding/test_plans/on_water_mock_ais.yaml
FireBoyAJ24: 😊. Good catch. That waypoint is in the middle of the Pacific Ocean. I will change that.

src/network_systems/projects/remote_server_standalone/README.md
FireBoyAJ24: Due to the whole docs workflow failing the test for unexpected reasons

src/network_systems/projects/remote_server_standalone/README.md
FireBoyAJ24: Same as said above

docs/current/sailbot_workspace/reference/launch_files.md
raghumanimehta: Yes. Given the original uses env path, it is safer as it provides a safer alternative. i am not sure how the new one directly runs ( i am wondering if it's path independent or dependent) but if it is, we might not be able to quickly diagnose the issue.

src/local_pathfinding/test_plans/on_water_mock_ais.yaml
raghumanimehta: 🤣

docs/current/sailbot_workspace/reference/launch_files.md
FireBoyAJ24: As long as the global path is colcon built. Then it should run like other ros packages we have.

FireBoyAJ24: Aka, it should be path independent

PR #911, opened 2026-06-01
882 path adding scripts on pathfinding data to automate pathfinding test plan

### LINKED ISSUE #882: PATH: Adding scripts on pathfinding data to automate pathfinding test plan
### Purpose
Currently, we have a data collection system to track any messages posted to the local_pathfinding topic during any on-water tests. We also have some scripts written to convert these into DataFrames and then to perform some calculations.

We want to continue adding to these scripts to better reflect the test plan written on Confluence based on the various tests that we will be going through.

Currently, we should look into if it is possible to track how many tacks/gybes our boat performs according to tests 2 & 3, as well as comparing our desired heading to the true wind direction to determine whether our path goes on "broad reached" or "close hauled" or in irons. Finally, look into whether it is possible to track the reason behind our path switching and the remaining waypoints of the original path.

### Changes
We will be implementing these scripts in explore_bag_data.ipynb on top of our currently existing scripts.

### Resources
https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/2398453909/Test+Plan+-+Pathfinding

ethanhu912:
### Description
- Resolves #882 
- Made some changes to `explore_bag_data.ipynb`
    - Added two scripts. One that tracks the number of tacks & gybes that the boat has taken in its path. This is done by calculating the angle between the desired heading and the true wind direction. When the angle changes sign, we know that a tack or gybe has occurred.
    - Added a script to display when a path change has occurred, the reason behind it, as well as the remaining waypoints that the boat did not pass through.
    - Adjusted the ordering and organization of the file.
- Added functionality for the path change reason in `local_path.py` and `node_navigate.py`
    - When a path is changed, the reason returned by must_change_path function is passed to update_if_needed and stored on LocalPath. 
    - The remaining waypoints from the previous path when the new path is generated is also stored on LocalPath. 
    - When the LPathData message is built, these two are included in the message and published.
    - If the path is reused (not changed), the last replan reason is set to an empty string and remaining waypoints set to 0.
    - Added tests to ensure:
        - That the correct reason is returned for all cases in which boat is in a condition where it must change path, i.e., when the path passes through an obstacle.
        - That the correct amount of remaining waypoints is returned and published.
        - That an empty message gets published when the path gets reused (no change).
        - That a message containing the reason gets sent when the path is changed.
        - That node_navigate.py actually posts these messages to the LocalPath topic.

### Verification
- Passes all tests
- Ran all scripts with a valid bag recording to ensure that scripts ran correctly with the new replan_reason field.
    - Used these scripts to also verify that replan_reason is correctly functioning


raghumanimehta (changes requested):
Image

I can't really run any of the cells without errors. Do you have to make sure somethings must happen before running? if yes, edit the first cell and add all the requirements there. If there is a docker container requirements like `installing xyz package` add it there as well (if you are comfortable with docker then you can add it to the dockerfiles for easy development but this is not necessary, I can ask someone later to take care of this).

ethanhu912:
Thanks Raghumani.

Are you using the notebook virtual environment? You can see in the top right side of your screen. If not, you can install it using `install_notebook_env.sh` and the README.md in the notebooks folder.

You will also need to run each setup cell (section titled 1. Setup) and helper function cell (titled 2. Helper Functions) chronologically. Then, you should be able to run anything in the file. I've added a note in the setup section outlining this.

FireBoyAJ24:
I will review this right now. Sorry for the delay!!!!

FireBoyAJ24:
@ethanhu912, @raghumanimehta and @AMaharaj16, should we add the `notebooks/install_notebook_env.sh` to `scripts/setup.sh` as the virtualenv is isolated from the ROS2 and system packages? I recommend doing this as it would make it easier for everyone on the pathfinding team to use the script here. The `scripts/setup.sh` always runs after Docker's devcontainer is built. Therefore, we don't have to run the script manually.

raghumanimehta:
> @ethanhu912, @raghumanimehta and @AMaharaj16, should we add the `notebooks/install_notebook_env.sh` to `scripts/setup.sh` as the virtualenv is isolated from the ROS2 and system packages? I recommend doing this as it would make it easier for everyone on the pathfinding team to use the script here. The `scripts/setup.sh` always runs after Docker's devcontainer is built. Therefore, we don't have to run the script manually.

Yes. @digcheese Can you take a look at this?

FireBoyAJ24 (changes requested):
This is good work! Using the OWT 5, I got the following plots:

Image

Image

Image

Image

There were no tacks, gybes and ambiguous crossings.

The only thing I want to add is an overall path map of the boat travelling from start to finish in either lat/lon or X/Y coordinates.

raghumanimehta (dismissed):

PR #913, opened 2026-06-04
#908: Connecting segment deviation to update_if_needed 

### LINKED ISSUE #908: PATH: Connect Segment deviation box to `update_if_needed`
### Purpose

In the function `update_if_needed` (here: https://github.com/UBCSailbot/sailbot_workspace/blob/01ffb0861a7d0608d326dacd152ee70cc10663b6/src/local_pathfinding/local_pathfinding/local_path.py#L480), we want to integrate segment_deviation_box calculations with `must_change_path` so that `update_if_needed` becomes aware if the boat has deviated from path or not. 

Make sure, you handle `IndexError` (and other errors thrown by segment deviation box, if any). Handle it by logging it and throwing a `PathNotFoundError`.

### Acceptance criteria 
- Add tests to the tests of `update_if_needed` in `test_local_path.py` that test the path is changed when there is a deviation big enough and the path is not changed when the deviation is not big enough. 
- Run the system and verify it runs properly. Adjust thresholds if switching is way too extreme.

MilcToast:
### Description

- Resolves #908 

### Changes Made
- Updated `must_change_path` to accept `boat_lat_lon`
- Integrated `exceeded_segment_deviation` check after the index out of bounds (too low/high) checks, therefore no issues should be raised from `exceeded_segment_deviation` 
- If `boat_lat_lon` is not provided (defaults to None), the deviation check is skipped and no path change is triggered from this condition
- `exceeded_segment_deviation` now logs when deviation is exceeded, including the current distance and max threshold in km

### Verification

- Added test verifying that the path is regenerated when deviation exceeds thresholds (`exceeded_segment_deviation` returns `True`)
- Added test verifying that the path remains unchanged when deviation is within acceptable limits (`exceeded_segment_deviation` returns `False`)
- Verified that the path is regenerated when the boat exceeds the deviation threshold, confirmed via logs by temporarily lowering the threshold to trigger the check

### DIFF
--- a/src/local_pathfinding/local_pathfinding/local_path.py
+++ b/src/local_pathfinding/local_pathfinding/local_path.py
@@ -526,6 +526,8 @@ class LocalPath:
             IndexError: If the target waypoint is reached and no next local waypoint exists.
         """
         self._target_lp_wp_index = target_lp_wp_index
+        boat_lat_lon = None
+
         new_aw = Wind(
             inputs.filtered_wind_sensor.speed.speed,
             inputs.filtered_wind_sensor.direction,
@@ -548,11 +550,11 @@ class LocalPath:
                 self._logger.warn(e)
                 boat_lat_lon = self.state.position  # type: ignore
 
-            must_change_reason = self.must_change_path(
-                received_new_global_waypoint,
-                boat_lat_lon,
-                new_aw,
-            )
+        must_change_reason = self.must_change_path(
+            received_new_global_waypoint,
+            boat_lat_lon,
+            new_aw,
+        )
 
         if must_change_reason.should_change_path:
             tries = 0
--- a/src/local_pathfinding/local_pathfinding/local_path.py
+++ b/src/local_pathfinding/local_pathfinding/local_path.py
@@ -433,7 +433,10 @@ class LocalPath:
         return distance_to_segment_km > max_deviation_km
 
     def must_change_path(
-        self, received_new_global_waypoint: bool, new_aw: Optional[Wind] = None
+        self,
+        received_new_global_waypoint: bool,
+        boat_lat_lon: Optional[ci.HelperLatLon] = None,
+        new_aw: Optional[Wind] = None,
     ) -> MustChangeReason:
         """Check if the path must be changed.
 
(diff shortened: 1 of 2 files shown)


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -430,10 +430,21 @@ def exceeded_segment_deviation(
         rejection_vector = boat_vector - projection_vector
         distance_to_segment_km = np.linalg.norm(rejection_vector)
 
-        return distance_to_segment_km > max_deviation_km
+        segment_exceeded = distance_to_segment_km > max_deviation_km
```
raghumanimehta, line 430: good change.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -427,10 +427,21 @@ def exceeded_segment_deviation(
...
+        return segment_exceeded
 
     def must_change_path(
-        self, received_new_global_waypoint: bool, new_aw: Optional[Wind] = None
+        self,
+        received_new_global_waypoint: bool,
+        boat_lat_lon: Optional[ci.HelperLatLon] = None,
+        new_aw: Optional[Wind] = None,
     ) -> MustChangeReason:
         """Check if the path must be changed.
```
AMaharaj16, line 447: could you add an args list in the docstring here?

src/local_pathfinding/local_pathfinding/local_path.py
AMaharaj16: typo here at the end of the line with the brackets

AMaharaj16: note for other reviewers: this case is actually not possible as both index cases are handled before `exceeded_segment_deviation()` is called in `must_change_path()`. still worth keeping in case we add this function in anywhere else though (for whatever reason).

AMaharaj16 (changes requested):
looks great, good work Jeremy! let me know when you've addressed my comments. just nits lol.

MilcToast:
> looks great, good work Jeremy! let me know when you've addressed my comments. just nits lol.

Thanks! I just pushed the fixes and addressed all your comments. Take another look whenever you can, and let me know if there is anything else you would like me to touch up!

raghumanimehta (approved):

AMaharaj16 (approved):
nice work!

PR #916, opened 2026-06-04
Desired heading now publishes sail = False instead of not instantiating.

raghumanimehta:
Title. 
Verification:

- [x] unit test
- [x] should later add test plan but don't have time to do that before next week.

### DIFF
--- a/src/local_pathfinding/local_pathfinding/node_navigate.py
+++ b/src/local_pathfinding/local_pathfinding/node_navigate.py
@@ -199,26 +199,38 @@ class Sailbot(Node):
         self.desired_heading = msg
 
         self.get_logger().info(
-            f"Publishing to {self.desired_heading_pub.topic}: {msg.heading.heading}, sail == {msg.sail}"  # noqa
+            f"Publishing to {self.desired_heading_pub.topic}: {msg.heading.heading}, "
+            f"sail == {msg.sail}"  # noqa
         )
         self.desired_heading_pub.publish(msg)
 
-        if msg.sail:
-            self.get_logger().debug(f"Publishing local path data to {self.lpath_data_pub.topic}")
-            self.publish_local_path_data()
+        self.get_logger().debug(f"Publishing local path data to {self.lpath_data_pub.topic}")
+        self.publish_local_path_data(msg.sail)
 
-    def publish_local_path_data(self):
+    def publish_local_path_data(self, sail: bool):
         """
         Collect all navigation data and publish it in one message.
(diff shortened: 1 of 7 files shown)


src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -177,6 +177,14 @@ def desired_heading_callback(self):
 
         if not self._all_subs_active():
             self._log_inactive_subs_warning()
+            msg = ci.DesiredHeading()
+            msg.heading.heading = 0.0
+            msg.sail = False
```
FireBoyAJ24, line 182: What does sail false and true mean. We should add this to the message in `custom_interfaces` as a comment

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -177,6 +177,14 @@ def desired_heading_callback(self):
 
         if not self._all_subs_active():
             self._log_inactive_subs_warning()
+            msg = ci.DesiredHeading()
+            msg.heading.heading = 0.0
```
FireBoyAJ24, line 181: Should we also change the steering int value, as that is what the network system uses to update the CAN bus over whether to set the rudder to 0 degrees instead of `sail`

FireBoyAJ24 (changes requested):

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: It's documented in one of the readmes on my last PR (don't remember if it's merged in or not). `Sail = True` means that the boat should sail. `Sail = False` means that no good path exists for a variety of reasons. 

`Sail = False` because staying put in a bad situation makes more sense than potentially moving to a worse situation.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: That is the plan. @lross03 is aware about this and they will handle this in NET.

src/local_pathfinding/local_pathfinding/node_navigate.py
FireBoyAJ24: It should also be documented in the custom interfaces. I make a quick commit over this

FireBoyAJ24 (dismissed):

FireBoyAJ24 (approved):

PR #917, opened 2026-06-06
Jb3982/895 gps data loss handling

### LINKED ISSUE #895: Stop the boat if GPS is unavailable
This issue is for the implementation that came from [this thread](https://ubcsailbot.slack.com/archives/C0ADGH53B0T/p1775773183926459).

We want to ensure that if we stop receiving data from certain topics, we should just stop the boats sail and wait. 

The objective of this issue is that if PATH hasn't received GPS data for >2 minutes, we need to publish `sail = false` with the heading bit set to 1 (stop turning).

jb3982:
### Description

- Resolves #895 

Adds GPS data-loss handling to PATH's `node_navigate.py`. PATH now tracks when GPS data was last received, starting from node startup and resetting the timer whenever `gps_callback()` receives a GPS message.

If GPS data has not been received for more than 2 minutes, PATH publishes a `DesiredHeading` stop command with `heading = 0.0`, `steering = 1`, and `sail = false`, then stops normal desired-heading calculation for that cycle.

### Verification

- [x] Confirmed focused local pathfinding navigate tests pass: `9 passed`
- [x] Added tests to `test_node_navigate.py`.


src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -133,6 +134,7 @@ def __init__(self):
         self.desired_heading = None
 
         # attributes
+        self.gps_timeout_start_time = self.get_clock().now()
```
raghumanimehta: why use `get_clock().now()` instead of `datetime.now()`?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -175,16 +178,22 @@ def filtered_wind_sensor_callback(self, msg: ci.WindSensor):
     def desired_heading_callback(self):
         """Get and publish the desired heading."""
 
-        if not self._all_subs_active():
-            self._log_inactive_subs_warning()
+        if self._gps_has_timed_out():
             msg = ci.DesiredHeading()
-            msg.heading.heading = 0.0
+            msg.steering = 1
```
raghumanimehta: `DesriredHeading` doesn't have field steering. What made you think it does?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -390,33 +385,30 @@ def update_params(self):
...
-            and self.gps is not None
-            and self.global_path is not None
-            and self.filtered_wind_sensor is not None
-        )
+        return self.ais_ships and self.gps and self.global_path and self.filtered_wind_sensor
+
+    def _gps_has_timed_out(self) -> bool:
+        """Checks if we haven't received a GPS message for more than 2 minutes."""
+
+        elapsed_sec = ((self.get_clock().now() - self.gps_timeout_start_time).nanoseconds) / (1e9)
```
raghumanimehta: Does `get_clock.now()` return nanoseconds? This seems as a unit mismatch. Verify this.  Also what time unit is `self.gps_timeout_start_time`? The unit should be easily legible from the variable name.

src/local_pathfinding/test/test_node_navigate.py
```diff
@@ -219,3 +195,46 @@ def test_get_desired_heading_disables_sail_when_path_not_found():
     sailbot.get_logger.return_value.warning.assert_called_once_with(
         "Unable to generate a local path; disabling sail"
     )
+
+
+class MockTime:
+    def __init__(self, nanoseconds):
+        self.nanoseconds = nanoseconds
+
+    def __sub__(self, other):
```
raghumanimehta, line 229: what does `__sub__` do ? and why was this used?

raghumanimehta (changes requested):
I have requested some clarifications and tests are not enough.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: Correction. It exists. However, why are you setting it to 1?

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: This was given in ROS2 node documentation, so I used it instead of Python’s `datetime.now()`. Also, I thought that would keep the timeout consistent with ROS time.

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: It returns a `rclpy.time.Time` object.
The `.nanoseconds` then gets the elapsed time in nanoseconds. Dividing by `1e9` converts that to seconds, so it matches `GPS_TIMEOUT_SEC` units.

`self.gps_timeout_start_time` is a ROS `Time` timestamp. I can rename it to something like `gps_timeout_ros_time` to make the unit clearer.

src/local_pathfinding/test/test_node_navigate.py
jb3982: It is used indirectly when the test calls `sailbot.desired_heading_callback()`. That method calls `_gps_has_timed_out()`, which then runs:

`self.get_clock().now() - self.gps_timeout_start_time`

In the test, both these values are `MockTime` objects and without `__sub__`, the test fails with `TypeError` because Python fails to subtract two `MockTime` objects.

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: I set it to 1 because the issue description says to publish `sail = false` with the heading bit set to 1 to stop turning. I inferred that `steering` was the only bit-like field in `DesiredHeading`, and I confirmed that interpretation with Aayush.

I wasn’t fully sure whether PATH was responsible for this field, since normal desired heading publishing does not set `steering`. This was also discussed in the last PATH sync, and the current direction was to leave it here for now while we confirm the `steering` confusion during PR review.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: can you send me the documentation? our code eslewhere uses `datetime.now()`, we should update that if this is better.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: PATH doesn't touch steering, setting `sail = False` is sufficient.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: Ok, there should be type hints for this where `get_clock.now()` is init.

src/local_pathfinding/test/test_node_navigate.py
raghumanimehta: oh, it's sub for subtract? please name it subtract. Also where are the test cases in the tests?

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: The Humble `rclpy` docs were pretty sparse, so I used the Python docs in Rolling and confirmed the same in Humble source:

- `Node.get_clock()` docs: https://docs.ros.org/en/rolling/p/rclpy/api/node.html#rclpy.node.Node.get_clock
- `Time` / `.nanoseconds` docs: https://docs.ros.org/en/rolling/p/rclpy/api/time.html
- Humble `Clock.now()` source: https://github.com/ros2/rclpy/blob/humble/rclpy/rclpy/clock.py
- Humble `Node.get_clock()` source: https://github.com/ros2/rclpy/blob/humble/rclpy/rclpy/node.py
- Humble `Time.nanoseconds` source: https://github.com/ros2/rclpy/blob/humble/rclpy/rclpy/time.py

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -175,6 +178,20 @@ def filtered_wind_sensor_callback(self, msg: ci.WindSensor):
...
+            msg = ci.DesiredHeading()
+            msg.steering = 1
+            msg.sail = False
+
+            self.desired_heading = msg
+
+            self.get_logger().warning(
+                f"GPS data has not been received for more than {GPS_TIMEOUT_SEC:.0f} seconds. "
+                f"Publishing to {self.desired_heading_pub.topic}: {msg.heading.heading}"
+            )
```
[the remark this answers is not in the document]
raghumanimehta: @FireBoyAJ24 @AMaharaj16 thoughts?

src/local_pathfinding/test/test_node_navigate.py
jb3982: I tried replacing `__sub__` with `subtract`, but then the test fails with:

`TypeError: unsupported operand type(s) for -: 'MockTime' and 'MockTime'`
From what I found, `__sub__` is the special method name Python looks for when code uses the `-` operator.
Python docs I found this in: https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types

As for tests, I added a direct timeout test and edited the existing callback test. I think these cover the meaningful cases for this change:
- Timeout math: boundary cases around GPS_TIMEOUT_SEC
- No GPS received from the start
- GPS callback resetting the timeout timestamp

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: I updated this to `gps_timeout_start_ros_time` and added the `Time` type hint where `self.get_clock().now()` initializes the timestamp:

`self.gps_timeout_start_ros_time: Time = self.get_clock().now()`

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -133,6 +135,7 @@ def __init__(self):
         self.desired_heading = None
 
         # attributes
+        self.gps_timeout_start_ros_time: Time = self.get_clock().now()
```
raghumanimehta, line 139: If this is init here, why do you init it in the callback function?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -175,6 +178,20 @@ def filtered_wind_sensor_callback(self, msg: ci.WindSensor):
     def desired_heading_callback(self):
         """Get and publish the desired heading."""
 
+        if self._gps_has_timed_out():
+            msg = ci.DesiredHeading()
+            msg.steering = 1
+            msg.sail = False
+
```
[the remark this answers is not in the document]
raghumanimehta: @jb3982 i agree with copilot here.

src/local_pathfinding/local_pathfinding/node_navigate.py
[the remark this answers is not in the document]
raghumanimehta: @jb3982 ignore this.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -390,33 +385,31 @@ def update_params(self):
...
     def _all_subs_active(self) -> bool:
-        return (
-            self.ais_ships is not None
-            and self.gps is not None
-            and self.global_path is not None
-            and self.filtered_wind_sensor is not None
-        )
+        return self.ais_ships and self.gps and self.global_path and self.filtered_wind_sensor
+
+    def _gps_has_timed_out(self) -> bool:
```
raghumanimehta: better name would be: 

```suggestion
    def _has_gps_timed_out(self) -> bool:
```

raghumanimehta: don't forget to update elsewhere as well

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -390,33 +385,31 @@ def update_params(self):
...
+        """Checks if we haven't received a GPS message for more than 2 minutes."""
+
+        elapsed_sec = ((self.get_clock().now() -
+                        self.gps_timeout_start_ros_time).nanoseconds) / (1e9)
```
raghumanimehta: magic number i.e. `1e9`. Should probably have constant as 
`SECONDS_TO_NS` or something like this.

src/local_pathfinding/test/test_node_navigate.py
```diff
@@ -219,3 +195,64 @@ def test_get_desired_heading_disables_sail_when_path_not_found():
...
+    sailbot.gps_timeout_start_ros_time = MockTime(0)
+    sailbot.get_clock = mock.Mock(
+        return_value=mock.Mock(now=mock.Mock(return_value=MockTime(int(elapsed_sec * 1e9))))
+    )
+
+    assert sailbot._gps_has_timed_out() is expected
+
+
+def test_desired_heading_callback_publishes_stop_when_no_gps_received_after_timeout():
+    sailbot = nn.Sailbot.__new__(nn.Sailbot)
```
raghumanimehta, line 252: why not use `__init__`?

raghumanimehta (changes requested):
minor fixes

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -370,6 +387,12 @@ def update_params(self):
     def _all_subs_active(self) -> bool:
         return self.ais_ships and self.gps and self.global_path and self.filtered_wind_sensor
 
+    def _gps_has_timed_out(self) -> bool:
+        """Checks if we haven't received a GPS message for more than 2 minutes."""
+
+        elapsed_sec = ((self.get_clock().now() - self.gps_timeout_start_time).nanoseconds) / (1e9)
+        return (elapsed_sec > GPS_TIMEOUT_SEC)
```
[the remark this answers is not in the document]
raghumanimehta: @jb3982 ignore this.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: Can we also have the subtraction in the next line so that is easier tor read?

src/local_pathfinding/local_pathfinding/node_navigate.py
jb3982: The re-assignmentr in `gps_callback()` updates the timestamp whenever a GPS message is received.

The initial assignment in `__init__` sets the startup baseline so we can still detect the case where no GPS message is ever received from the start.

src/local_pathfinding/test/test_node_navigate.py
jb3982: Actually, I used the existing code from `test_get_desired_heading_disables_sail_when_path_not_found`.

My understanding was that this created a `Sailbot` instance without running the full ROS node initialization in `__init__`. That allowed us to manually set only the attributes needed for this case.

raghumanimehta (approved):

PR #918, opened 2026-06-06
Raghumanimehta/915 goal objective

### LINKED ISSUE #915: PATH: Prevent OMPL local paths from switching back away from the goal
## Prevent OMPL local paths from switching back away from the goal

We are seeing generated local paths that make a “switch back” away from the intended global waypoint before continuing toward it. In the plotted example, the path briefly moves backward in `x` even though the goal is at `(0, 0)` in OMPL local coordinates.

Image

### Suggested fix

Add a new optimization objective in `src/local_pathfinding/local_pathfinding/ompl_objectives.py` that penalizes path segments that do not make progress toward the goal.

Preferred approach: compute progress using projection onto the start-to-goal vector.

- Build a goal-progress direction vector from the OMPL start state to the goal state.
- For each segment `s1 -> s2`, compute `segment_vector dot goal_direction`.
- If the projection is positive, return cost `0`.
- If the projection is zero/negative, return a very large cost or `inf`.
- Include a small tolerance so tiny numerical jitter near perpendicular does not get punished.

This is better than checking `x`/`y` independently because tacking or obstacle avoidance may require sideways motion while still making down-course progress.

### Alternate/simple approach

Add an objective that checks coordinate-wise progress toward the intended goal:

- If either `x` or `y` moves toward the intended goal coordinate, return cost `0`.
- If neither coordinate moves toward the goal, return very high or infinite cost.

This is simpler, but may still allow some switchbacks if one coordinate improves while the overall segment moves away from the goal.

### Testing

Add unit tests for the new objective covering:

- Segment directly toward goal has cost `0`.
- Segment perpendicular or away from goal is penalized.
- Segment with tiny near-zero projection respects tolerance.
- A tacking-like segment with positive goal projection is allowed.
- Existing wind/time objective behavior remains unchanged.

11:34 AM

raghumanimehta:
### Description

- Resolves #915 (hopefully) 
- To solve switchbacks, added a new objective in `ompl_path.py`. This objective is quite straightforward, it does a dot product of vector from any state `s` to state `s+1` and `s` to the goal state `g`. What this gives us is a positive constant if the vector `s` to `s+1` lies in the same general direction as vector from `s` to `g`. This is heurisitc. 
- The previous point is not enough, I also added ValidityChecker (https://ompl.kavrakilab.org/classompl_1_1base_1_1MotionValidator.html); this is a hard constraint version of the previous point. If the check fails, ompl will not consider the path that passes through this.  

### Verification

- [x] Manual verification. Would appreciate extensive testing on your part as well. 
- [x] Tests


src/local_pathfinding/local_pathfinding/ompl_objectives.py
```diff
@@ -269,6 +306,10 @@ def get_sailing_objective(
         objective=TimeObjective(space_information, tw_dir_rad, tw_speed_kmph),
         weight=TIME_OBJECTIVE_WEIGHT,
     )
+    multiObjective.addObjective(
+        objective=GoalDirectionObjective(space_information, goal_position_in_xy),
+        weight=1.0,  # should always be 1.0
```
FireBoyAJ24, line 311: Why should the weight of the GoalDirectionObjective always be 1.0? Is it because its cost should be the largest in comparison to the `TimeObjective` and `WindObjective`?

src/local_pathfinding/test/test_ompl_validity.py
```diff
@@ -0,0 +1,97 @@
...
+        (
+            # Movement exactly 90 degrees to the goal direction should be allowed
+            90.0,
+            True,
+        ),
+        (
+            # Movement 91 degrees to the goal direction is slightly backward
+            91.0,
+            False,
+        ),
```
FireBoyAJ24, line 50: I assume `-91.0, False` or `270, True` is redundant as `s2 = cs.XY(math.cos(angle_rad), math.sin(angle_rad))` would simply be the mirror image of the positive vector and will be handled similarly. Should we add it or will it be redundant?

s2,
FireBoyAJ24, line 1: What is this?

FireBoyAJ24 (dismissed):

src/local_pathfinding/local_pathfinding/ompl_objectives.py
raghumanimehta: Because it is always 0 or inf so it acts like a yes or no.

s2,
raghumanimehta: Seems like a weird abstract that was added. Removed.

src/local_pathfinding/test/test_ompl_validity.py
raghumanimehta: I think it's fine to not add this, the function is just a large wrapper for a dot product. But i don't mind if you think it should be added.

PR #919, opened 2026-06-08
User/bubbles5228/path mock gps noise

### LINKED ISSUE #903: PATH: Mock GPS noise
### Description

This feature adds more realistic GPS behaviour to the mock sensor system by introducing two independent effects:

1. **GPS noise**
   - Adds Gaussian noise to GPS readings.
   - Adds occasional outliers to simulate bad GPS jumps.
   - Can be turned on/off independently before running.
   - If possible, should also be toggleable during runtime through the visualizer.

2. **Ocean drift**
   - Adds a small current-based offset to the GPS position.
   - This should be separate from GPS noise.
   - As a second step, the drift offset can include small random variation so the current is not perfectly constant.
   - Can be turned on/off independently before running.
   - If possible, should also be toggleable during runtime through the visualizer.

These features should stay simple. The goal is not to build a full physical ocean model, but to make the mock GPS less idealized so Local Pathfinding and related systems can be tested under more realistic conditions.

Each feature should be independently configurable:

- GPS noise on/off
- GPS outliers on/off, or controlled as part of GPS noise
- Ocean drift on/off
- Ocean drift randomization on/off

The visualizer should expose these toggles if feasible, so the behaviour can be enabled or disabled while the system is running. Otherwise, they should at least be configurable before startup.

Testing should be done through manual verification by running the mock sensor system and checking that:

- GPS values remain unchanged when all effects are disabled.
- Gaussian noise causes small random variation in GPS readings.
- Outliers occasionally create larger GPS jumps when enabled.
- Ocean drift creates a consistent position offset over time.
- Drift randomization causes the offset to vary slightly instead of staying perfectly constant.
- Each feature can be turned on and off independently without affecting the others.

bubbles5228:
### Description
- resolves #903
- adds functionality for gps noise and ocean drift, along with drift direction, speed, noise, and acceleration
- everything is toggleable

### Verification
- the visualizer was run and behavior seemed normal; logs also looked normal
- `ros2 launch local_pathfinding main_launch.py mode:=development`


FireBoyAJ24:
Can you (@bubbles5228) please document the different launch params in the local_pathfinding `README.md`? In addition, example ros2 launch commands?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -87,15 +110,87 @@ def __init__(self) -> None:
...
+
+        self._drift_speed_kmph += self._drift_accel_kmph2 * dt_hours
+
+        drift_speed = self._drift_speed_kmph
+        drift_dir = self._drift_dir_deg
+
+        if self._use_drift_randomization:
+            drift_speed += random.gauss(0.0, drift_speed * 0.05)
+            drift_dir += random.gauss(0.0, 2.0)
+
```
FireBoyAJ24: Would be nice to make these magic numbers into constants and place at the top of this file

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -87,15 +110,87 @@ def __init__(self) -> None:
         # Parameter Event Handler (Parameters can change over the life of the simulation)
         self.add_on_set_parameters_callback(self._on_set_parameters)
 
+    def add_gps_noise(
+        self, lat_lon_msg: ci.HelperLatLon, noise_sigma_meters: float = 0.5
```
FireBoyAJ24: We should add the `noise_sigma_meters = 0.5` as a constant at the top of the file

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -87,15 +110,87 @@ def __init__(self) -> None:
...
+            published_location = self._current_location
+
+        if self._use_drift:
+            published_location = self.add_ocean_drift(published_location)
+
+            self.get_logger().info(
+                f"Drift Offset: {self._drift_offset_km}\n"
+                f"Drift Speed: {self._drift_speed_kmph}\n"
+                f"Drift Direction {self._drift_dir_deg}"
+            )
```
FireBoyAJ24: It would be best if we add the ocean drift to the true gps position before we add noise to the gps location. Right now the ocean drift is added to a jittering lat lon instead.

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -125,11 +220,27 @@ def _on_set_parameters(self, params: List[Parameter]) -> SetParametersResult:
...
+            elif p.name == "use_ocean_drift":
+                self._use_drift = bool(p.value)
+                if not self._use_drift:
+                    self._drift_offset_km = cs.XY(
+                        0.0, 0.0
+                    )  # resets the drift offset after it's turned off
+            elif p.name == "use_drift_randomization":
+                self._use_drift_randomization = bool(p.value)
+            elif p.name == "ocean_drift_speed_kmph":
+                self._drift_speed_kmph = float(p.value)
```
FireBoyAJ24, line 234: Should we reset the drift speed to 0.0 when we toggle the `use_ocean_drift` off (initially on)? Otherwise the drift speed will remain at its previous value and keep on accumulating once the `use_ocean_drift` is turned on.

FireBoyAJ24 (changes requested):
Overall, the work is nice! Just questions over the drift speed accumulating and documentation. ~~Could we also have a pop-up when we toggle or change the GPS noise/ocean drift, similar to when the wind params change? We can log it in ROS as well.~~ It's there but just takes time to show.

src/local_pathfinding/launch/main_launch.py
```diff
@@ -15,7 +15,41 @@
...
+    DeclareLaunchArgument(
+        name="use_drift_randomization",
+        default_value="true",
+        choices=["true", "false"],
+        description="Enable small random variation to the ocean drift current each tick.",
+    ),
+    DeclareLaunchArgument(
+        name="ocean_drift_speed_kmph",
+        default_value="0.5",
+        description="Base speed of the ocean current in km/h.",
```
raghumanimehta: this is speed over ground right?

src/local_pathfinding/launch/main_launch.py
```diff
@@ -15,7 +15,41 @@
...
+    ),
+    DeclareLaunchArgument(
+        name="ocean_drift_dir_deg",
+        default_value="45.0",
+        description="Direction the current flows toward in degrees (0=north, 90=east).",
```
raghumanimehta: should clarify that it is (-180,.0 180.0]. Please verify if it is (-180,.0 180.0] or [-180.0,  180.0), I don't remember. Use the one that is used for wind as well. Check `wind_coordinate_system.py` or `WindReadme.md`

src/local_pathfinding/launch/main_launch.py
```diff
@@ -247,9 +281,32 @@ def get_mock_gps_node_description(context: LaunchContext) -> Node:
     """Gets the launch description for the mock gps node"""
     node_name = "mock_gps"
     test_plan = LaunchConfiguration("test_plan").perform(context)
+
+    use_gps_noise = LaunchConfiguration("use_gps_noise").perform(context)
+    use_gps_noise_bool = use_gps_noise.lower() in ["true", "1", "yes"]
```
raghumanimehta, line 289: so you can specify either `yes` or `1` or `true` to turn it on? 

This should be documented. Create a new markdown in `mock_nodes` folder so that we don't have to refer to this everytime. 

Ideally, document the usage of mock nodes completely, even the parts you have not worked on. 
feel free to use AI for this but please verify the output.

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -17,12 +18,15 @@
 from local_pathfinding.ompl_objectives import TimeObjective
 
 SECONDS_PER_HOUR = 3600
+GPS_NOISE_SIGMA_METERS = 0.5
```
raghumanimehta, line 21: These are not controllable from terminal or GUI, right? 

How did you get these values?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
```diff
@@ -87,15 +113,84 @@ def __init__(self) -> None:
...
+
+        Returns:
+            ci.HelperLatLon: Object containing the new lat/lon coordinates with ocean drift.
+        """
+
+        dt_hours = self.pub_period_sec / SECONDS_PER_HOUR
+
+        self._drift_speed_kmph += self._drift_accel_kmph2 * dt_hours
+
+        drift_speed = self._drift_speed_kmph
```
raghumanimehta: missing units to variable name

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
raghumanimehta: Yes, we should reset

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -37,8 +37,8 @@
 DEFAULT_PLOT_RANGE = [-100.0, 100.0]
 BOX_BUFFER_SIZE_KM = 1.0
 
-WIND_BOX_X_DOMAIN = (0.76, 0.99)
-WIND_BOX_Y_DOMAIN = (0.00, 0.22)
+WIND_BOX_X_DOMAIN = (0.80, 0.98)
```
raghumanimehta, line 41: this was done to accomodate,  new UI?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -1037,7 +1037,11 @@ def write_wind_params(tw_dir_deg: float, tw_speed_kmph: float) -> None:
     with open(WIND_PARAMS_YAML, "w") as f:
         yaml.safe_dump(data, f, sort_keys=False)
 
-    apply_wind_params()
+    def apply_wind_params():
```
raghumanimehta: Why do we have definitions of `apply_wind_params`?

src/local_pathfinding/local_pathfinding/visualizer.py
```diff
@@ -1099,7 +1128,7 @@ def dash_app(q: Queue):
                 id="control-panel",
                 style={
                     "position": "absolute",
-                    "bottom": "120px",  # Distance from the very bottom of the screen
+                    "bottom": "175px",  # Distance from the very bottom of the screen
```
raghumanimehta: not your doing, but remove the comment here.

src/local_pathfinding/README.md
```diff
@@ -30,6 +30,12 @@ Launch arguments are added to the run command in the format `<name>:=<value>`.
 | `log_level` | Logging level (default: `{INFO}`) | A [severity level](https://docs.ros.org/en/humble/Concepts/About-Logging.html#severity-level) (case insensitive) `{DEBUG, INFO, WARN, ERROR, and FATAL}` |
 |`mode` | Mode (default: `development`) | `{development, production}`|
 |`test_plan` | Test Plan is not required when `mode=production` | any test definition yaml file in `local_pathfinding/test_plans`. (eg. `test_plan:=basic.yaml`) |
+|`use_gps_noise`|Enable GPS noise (default:`true`)|`{true, false}`|
```
raghumanimehta, line 34: good that you documented this here. 
Earlier in the review, I mentioned to create a readme in mock_nodes/ but instead add a heading Mock Nodes here and then add the full description of how to use the mock nodes. The description you added here, can be used as well.

raghumanimehta (changes requested):
Good work! I have requested some changes though

src/local_pathfinding/local_pathfinding/visualizer.py
bubbles5228: yes

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
bubbles5228: no they are not controllable
ocean drift was mostly just a value that seemed to give me reasonable drift, gps was based on gps testing found on confluence which i took and did something with it (dont remember the specifics) but i think it should be realistic given the type of gps we have

src/local_pathfinding/launch/main_launch.py
bubbles5228: yes

src/local_pathfinding/launch/main_launch.py
raghumanimehta: Can you add what it is in the description?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_gps.py
raghumanimehta: still missing the units. By units i mean, `drift_speed_kmph` instead of `drift_speed`.

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: I think you forgot to address this. If not, what was changed to address this?

src/local_pathfinding/local_pathfinding/visualizer.py
raghumanimehta: The comment's not removed yet.

raghumanimehta (changes requested):
There are still some things to iron out.

PR #923, opened 2026-06-09
Raghumanimehta/922 wind motion validator

### LINKED ISSUE #922: PATH: Motion validity for Wind
### Purpose

We added a state validity check for not having switch backs, implemented in #918 . We should have a similar state validity check for wind as well. 
Difference between an `objective` and `state validity` is that objectives are used for optimizaitons whereas state validity is used to designate hard no-go zones. 
What would be wind no-go zones? The same as what we have in objectives with one major change. Instead of having a complex function modelling the cost of the wind, this should be simply:
```
if no-go-zone:
    return False
return True   # verify if this is the bool orientation for OMPL api
```

### Resources

- https://ompl.kavrakilab.org/classompl_1_1base_1_1DiscreteMotionValidator.html
- https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/local_pathfinding/local_pathfinding/ompl_validity.py

raghumanimehta:
### Description

- Resolves #922 
- Added a common motion validator that checks for goal progression and no go zone for wind. 
- The paths, in my expereince, are much better now
- I have also refactored some code to ensure that we reuse the core logic as much as possible
- I have also added some debug statements for significant wind change path because i realized debugging that was a pain and it was worthwhile to have better debugs there. 
- I have (Gemini :) ) added visual cones for visualizer, the cones show wind no go zone (irons and reverse irons (forgot what it's called)). 
- btw you can't have multiple validators like multiple objectives. trust me, i checked. 

### Verification

- [x] Manual verification
- [x] Unit tests. I realize some cases might have been missed, so feel free to recommend more.


src/local_pathfinding/local_pathfinding/mock_nodes/wind_params.yaml
```diff
@@ -1,8 +1,8 @@
...
-    tw_dir_deg: 80
-    tw_speed_kmph: 10.0
+    tw_dir_deg: 0
+    tw_speed_kmph: 0.0
 /mock_gps:
   ros__parameters:
-    tw_dir_deg: 80
-    tw_speed_kmph: 10.0
+    tw_dir_deg: 0
+    tw_speed_kmph: 0.0
```
FireBoyAJ24, line 8: We should add this file to gitignore as these changes don't need to be tracked.

src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -57,10 +57,11 @@ class WindTracker:
             rolling average was populated, using only one apparent wind point.
     """
 
-    def __init__(self):
+    def __init__(self, logger: Optional[RcutilsLogger] = None):
         self.aw_history: deque = deque(maxlen=WIND_HISTORY_LEN)
         self.aw_avg: Optional[Wind] = None
         self.using_one_aw_point: bool = True
+        self._logger = logger or RcutilsLogger(name="wind_tracker")
```
FireBoyAJ24: We can import a logger like this,
```
from rclpy.logging import get_logger

# Inside your node or module:
logger = get_logger('node_name')
```

FireBoyAJ24 (dismissed):

src/local_pathfinding/local_pathfinding/mock_nodes/wind_params.yaml
raghumanimehta: It's already added, Idk why that is not working .

PR #927, opened 2026-06-10
Fix #920: Tracking true wind in WindTracker instead of apparent wind

### LINKED ISSUE #920: PATH: Fix unnecessary path changes caused by apparent wind
### Purpose

Currently the boat is changing paths very frequently after reaching a waypoint or changing direction, because `significant_wind_change` is being triggered by apparent wind direction shifts that are a result of the boat's own heading change, rather than a genuine change in environmental wind conditions.

### Changes

- In `local_path.py`:
  - Have the wind history deque track true wind rather than apprent wind
  - Have the stored wind at path_generated_wind also be converted to true wind.
  - Validate with log data (triggers due to boat rotation eliminated, while true wind changes remain)

### Resources

To reproduce this:
- Observe the boat as it reaches a waypoint or executes a heading change
- Note that `significant_wind_change` fires shortly after, triggering a path recalculation
- The wind conditions themselves have not meaningfully changed, only the boat's heading has

MilcToast:
### Description

- Resolves #920 

### Changes Made
- In the `WindTracker` class, `tw_history` and `tw_avg` now track true wind; apparent-to-true conversion via get_true_wind happens in `update_state` and `update_if_needed` before any wind is stored
- `path_generated_wind` now stores the true wind value
- `must_change_path` updated to pass `tw_avg` and `path_generated_wind` (both true wind) into `is_significant_wind_change`

Note: minor true wind direction fluctuations (~6-9°) observed in logs, occasionally hitting the 10° threshold. likely due to noise from small variations in boat speed/heading affecting apparent-to-true wind conversion. 

### Verification

- Validate with log data: path changes triggered purely by heading rotation should no longer fire `is_significant_wind_change`
- Added a test `test_update_if_needed_reuses_path_when_boat_changes_heading` that verifies a heading change does not trigger a path regeneration when true wind is constant
- Verified with simulation that a genuine true wind direction/speed change correctly triggers a path regeneration as expected


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -40,56 +40,59 @@ class MustChangeReason:
...
-        self.aw_avg: Optional[Wind] = None
-        self.using_one_aw_point: bool = True
+        self.tw_history: deque = deque(maxlen=WIND_HISTORY_LEN)
+        self.tw_avg: Optional[Wind] = None
+        self.using_one_tw_point: bool = True
 
-    def update_aw_history(self, current_aw: Wind):
-        """Updates apparent wind history and recalculates the average wind. The wind values are
-        all apparent wind.
+    def update_tw_history(self, current_wind: Wind):
```
raghumanimehta: the parameter current_wind can be improved to new_tw.

raghumanimehta (changes requested):
@MilcToast 
Your work looks good. I have one small suggestion and a request. Suggestion is within the review. 

**In update_if_needed, verify that the windTracker is regularly updated with new wind values.**
It seems that the updates are not happening properly; don't make changes. Share what you find out with me first.

raghumanimehta:
> @MilcToast Your work looks good. I have one small suggestion and a request. Suggestion is within the review.
> 
> **In update_if_needed, verify that the windTracker is regularly updated with new wind values.** It seems that the updates are not happening properly; don't make changes. Share what you find out with me first.

Did you look into this @MilcToast ?

MilcToast:
> > @MilcToast Your work looks good. I have one small suggestion and a request. Suggestion is within the review.
> > **In update_if_needed, verify that the windTracker is regularly updated with new wind values.** It seems that the updates are not happening properly; don't make changes. Share what you find out with me first.
> 
> Did you look into this @MilcToast ?

No, not yet. I'll look into it today.

MilcToast:
> **In update_if_needed, verify that the windTracker is regularly updated with new wind values.** It seems that the updates are not happening properly; don't make changes. Share what you find out with me first.

The wind tracker is being updated regularly and correctly. After updating the wind parameters in the visualizer, new true wind readings come in. Otherwise the same wind readings (with minimal noise differences) come in. Both of these cases every ~2 seconds and the rolling average updates as expected, so no issues there. 

Something else I noticed though: the true wind speed logged by the windTracker is consistently higher than what I set. For example, setting tw_speed_kmph to 8 results in ~13 kph being logged, and setting it to 15 results in ~24 kph. Not too sure why.

MilcToast:
Just something else to note for the wind tracker: right at launch, wind_tracker.using_one_tw_point evaluates to True, meaning the very first global path is generated using just a single true wind reading.

Because of this, if there's a significant wind shift within the first 10 minutes of deployment, the boat won't automatically trigger a path change to adapt (unless a differnet event forces a path change). It might be worth adding a check or shortening the recalculation threshold during the initialization window so the boat doesn't get locked onto a suboptimal or risky path right after launch.

raghumanimehta:
> Just something else to note for the wind tracker: right at launch, wind_tracker.using_one_tw_point evaluates to True, meaning the very first global path is generated using just a single true wind reading.
> 
> Because of this, if there's a significant wind shift within the first 10 minutes of deployment, the boat won't automatically trigger a path change to adapt (unless a differnet event forces a path change). It might be worth adding a check or shortening the recalculation threshold during the initialization window so the boat doesn't get locked onto a suboptimal or risky path right after launch.

That's the intended desing. It should be fine. Suboptimal path is fine in terms of wind as long as it is not risky, which it shouldn't be... We can think of addin a shorter deadline for the first iteration though. @AMaharaj16 .

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
```diff
@@ -117,7 +116,7 @@ def get_true_wind(
     return ZERO_VECTOR_CONSTANT, 0.0
 
 
-def get_apparent_wind(
+def tw_gc_to_aw_gc(
     tw_dir_deg_gc: float,
```
raghumanimehta, line 113: Change this to use Wind Object.

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
```diff
@@ -91,12 +90,12 @@ def get_true_wind(
         The function computes vector components in an east/north frame, subtracting the boat
         motion from the apparent wind to obtain the true wind.
     """
-    aw_dir_rad_bc = math.radians(aw_dir_deg_bc)
+    aw_dir_rad_gc = math.radians(aw_dir_deg_gc)
 
     # boat wind is in the direction of the boat heading (reverse of boat heading)
     bw_dir_rad_gc = math.radians(cs.bound_to_180(boat_heading_deg_gc + 180))
```
raghumanimehta, line 93: why are you adding 180 here then bounding it to 180 and then converting to radians? 
Could just be math.radians(boat_heading_deg_gc). Also for boat, use heading instead of dir.

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
```diff
@@ -91,12 +90,12 @@ def get_true_wind(
...
-    aw_east_kmph = aw_speed_kmph * math.sin(aw_dir_rad_bc)
-    aw_north_kmph = aw_speed_kmph * math.cos(aw_dir_rad_bc)
+    aw_east_kmph = aw_speed_kmph * math.sin(aw_dir_rad_gc)
```
raghumanimehta, line 94: Did you verify these computations? Changing the coordinate can break these.

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
raghumanimehta, line 143: same here, why are you adding 180?

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -395,9 +396,22 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+            current_tw.speed_kmph,
+            self.state.heading,
+            self.state.speed
+        )
+
+        current_aw_dir_deg_bc = wcs.global_to_boat_coordinate(
+            self.state.heading,
+            current_aw_dir_deg_gc,
+        )
+        current_aw = wcs.Wind(current_aw_speed_kmph, current_aw_dir_deg_bc)
```
raghumanimehta: Did you verify that the test cases ran fine without changing anything else? if yes, then good. If no, then how did you verify that the calculations are correct.

src/local_pathfinding/local_pathfinding/ompl_objectives.py
```diff
@@ -303,7 +303,7 @@ def get_sailing_objective(
         boat_heading_deg_gc, aw_direction_deg_bc
     )
 
-    tw_dir_rad, tw_speed_kmph = wcs.get_true_wind(
+    tw_dir_rad, tw_speed_kmph = wcs.aw_gc_to_tw_gc(
```
raghumanimehta: I think we should add the coordinate to everywhere in the codebase we mention wind. It can get very confusing, otherwise. Please make this change.

raghumanimehta (changes requested):
There are still some problems that need to be figured out.

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
MilcToast: 180 is added to the boat's heading to calculate the boat wind's direction, since it is in the opposite direction of where the boat is heading towards. You then need to bound it back to 180 in case it leaves the range of (-180, 180].

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
MilcToast: The math in this function was previously written as if all the angles were with respect to the same coordinate frame, meaning one of them was mislabeled. 
In the calculation, the boat wind and apparent wind's vectors are separated into their north and east component vectors and then subtracted from each other. This separation is done with a simple sin/cos for both, meaning that whoever wrote this previously expected both wind directions to be in the same coordinate frame.
The rest just combines the east and north components of the true wind's vector to get its magnitude and direction in the same coordinate frame as the shared frame of apparent and boat wind.

Therefore, a change in math wasn't necessary (since we want everything to be in the same coordinate frame now)

src/local_pathfinding/local_pathfinding/ompl_path.py
MilcToast: I can't run the tests right now because of an OMPL environment blocker (it keeps throwing an AttributeError for DubinsStateSpace whenever any test tries to run).

The code just retrieves the LocalPathState's current true wind (which was calculated from wind sensor data using the aw_gc_to_tw_gc function) and converts it back. I don't think there should be any issues with this unless there is a mathematical mistake inside those two newly named functions themselves.

MilcToast: Alternatively, we could just keep the current apparent wind field that I removed inside LocalPathState and avoid the double conversion entirely.

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
MilcToast: Same reason as above

src/local_pathfinding/local_pathfinding/wind_coord_systems.py
raghumanimehta: Going to do this in a different PR

raghumanimehta (dismissed):

MilcToast:
Other Note: The old function name `get_true_wind` is still being used in `explore_bag_data.ipynb`. I tried to go in and update it, but it's a little buggy for me, so it won't open properly in the notebook editor. Leaving this file as-is for now if someone else wants to take a pass at fixing the syntax and updating the call!

raghumanimehta:
> he old function name `get_true_wind` is still being used in `explore_bag_data.ipynb`. I tried to go in and update it, but it's a little buggy for me, so it won't open properly in the notebook editor. L

@ethanhu912 Can you please take a look at this?

AMaharaj16 (dismissed):
Looks good besides documentation in `/src/local_pathfinding/README.md` needs to still be updated in accordance with these changes.

raghumanimehta (approved):

PR #935, opened 2026-06-15
Pathfinding logging

AMaharaj16:
## Summary

OMPL crashed during OWT5 and we couldn't tell why from the logs. This adds targeted logging around the local pathfinding loop so any next failure is actually diagnosable.

We didn't get a definitive root cause from OWT5, but a strong suspect is the old per-state file logging in is_state_valid: it wrote to a hardcoded `/workspaces/...` devcontainer path on every invalid state. If that path isn't writable wherever the node actually runs, open() throws from inside OMPL's solve loop, and since update_if_needed only catches ValueError, it would propagate and kill the navigate node. The write is back to exactly what it was on main, but the call is now wrapped so a file error can't crash the solver.

## Changes

  ompl_path.py:
  - Wrapped solve() so an exception from one of our callbacks or a dependency (shapely, pyproj, numpy, OMPL) gets logged with the inputs that were
  in flight, then re-raised, instead of disappearing.
  - Added a guard for a non-finite start position before it poisons the OMPL bounds, and a warning when the wind/heading/speed feeding the objective
  is non-finite.
  - is_state_valid is now an instance method so it has access to the logger, and the log_invalid_state call is wrapped in a try/except.
  - Clearer message when land.pkl fails to load (resolved path plus error type).

  local_path.py:
  - Retry-loop warnings now say which attempt failed and why.
  - On every path switch we now log how long we were on the previous path and the reason for switching.

  node_navigate.py:
  - The pathfinding timer callback now catches any unexpected error, logs the cause and inputs, and disables sail for that tick instead of letting the navigate node die. This is the one behavior change, so it's worth a close look in review.

  test_ompl_path.py:
  - Calls is_state_valid on an instance now that it takes self.


src/local_pathfinding/local_pathfinding/local_path.py
```diff
@@ -661,6 +671,11 @@ def update_if_needed(
                     + f" within {MAX_OMPL_PATH_GEN_TRIES}"
                 )
 
+            if self.state is not None:
+                time_on_prev_sec = self._now_sec() - self.state.path_generated_time_sec
```
raghumanimehta, line 714: I think there is a function for this.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -184,6 +186,28 @@ def filtered_wind_sensor_callback(self, msg: ci.WindSensor):
 
     # publisher callbacks
     def desired_heading_callback(self):
+        """Timer callback. Runs one pathfinding tick; if anything unexpected raises, log the
```
raghumanimehta: I don't like this change. This can add bugs on to an already well tested system. Basic logs here would be better.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -88,7 +88,22 @@ def __init__(
         # this needs state
         self._simple_setup = self._init_simple_setup(land_multi_polygon)
 
-        self.solved = self._simple_setup.solve(time=MAX_SOLVER_RUN_TIME_SEC)
+        try:
```
raghumanimehta: ditto for this one. I think this already caught in local_path.py

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -343,6 +362,16 @@ def update_objectives(self):
     def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
         # Create buffered rectangles around sailbot's position and the goal state
         start_position_in_xy = cs.latlon_to_xy(self.state.reference_latlon, self.state.position)
+        if not (math.isfinite(start_position_in_xy.x) and math.isfinite(start_position_in_xy.y)):
```
raghumanimehta: Don't need this imo.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -379,7 +408,7 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
         )
 
         simple_setup = og.SimpleSetup(space)
-        simple_setup.setStateValidityChecker(base.StateValidityCheckerFn(OMPLPath.is_state_valid))
+        simple_setup.setStateValidityChecker(base.StateValidityCheckerFn(self.is_state_valid))
```
raghumanimehta, line 426: Can you explain this change?

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -399,6 +428,22 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
         else:
             current_aw = self.state.path_generated_wind
 
+        # NaN wind/heading/speed makes OMPL costs undefined and can crash/stall the solver with
+        # no catchable exception, so surface it here (the solve-wrapper can't catch it). These
+        # values feed both the wind-aware motion validator and the objective below.
+        if not (
+            math.isfinite(current_aw.dir_deg)
```
raghumanimehta: Why are you going so defensive for everycheck? We don't want to add bloat.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -429,10 +474,14 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
 
         return simple_setup
 
-    def is_state_valid(state: Union[base.State, base.SE2StateInternal]) -> bool:
+    def is_state_valid(self, state: Union[base.State, base.SE2StateInternal]) -> bool:
```
raghumanimehta, line 470: Why was this function chagned?

raghumanimehta (changes requested):
Why are there so many logical changes?
Can we just have logging changes?

src/local_pathfinding/local_pathfinding/ompl_path.py
AMaharaj16: needs to be an instance method so i can log through `self._logger` (see new line 475). the logic is all the same besides logging of the errors raised by `log_invalid_state`. i just simplified it a bit and the comments made no sense to me.

src/local_pathfinding/local_pathfinding/local_path.py
AMaharaj16: where? i can't seem to find it anywhere.

src/local_pathfinding/local_pathfinding/ompl_path.py
AMaharaj16: `is_state_valid` is now an instance method with the `self` field to allow for logging so i changed it accordingly here (the only place it is called). before, it was a static class method with no `self` field.

src/local_pathfinding/local_pathfinding/ompl_path.py
AMaharaj16: should i remove this change then?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -212,25 +214,45 @@ def desired_heading_callback(self):
             self.desired_heading_pub.publish(msg)
             return  # should not continue, return and try again next loop
 
-        self.update_params()
+        try:
```
raghumanimehta, line 236: what can go wrong here that you have added this?

src/local_pathfinding/local_pathfinding/ompl_path.py
raghumanimehta: If it's caught in the calling function, yes. Also, please don't abuse `try catch`

src/local_pathfinding/local_pathfinding/ompl_path.py
raghumanimehta: how does this help with logging?

src/local_pathfinding/local_pathfinding/ompl_path.py
raghumanimehta: The comment is not non-sensical. `log_invalid_state` writes to a file, if we don't uncomment this part out before launch then it will keep writing it and we will run out of disk space.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -444,21 +463,16 @@ def is_state_valid(state: Union[base.State, base.SE2StateInternal]) -> bool:
...
-                    else:  # happens in prod
-                        log_invalid_state(state=cs.XY(state.getX(), state.getY()), obstacle=o)
-
+                    point = cs.XY(state.getX(), state.getY())
+
+                if not o.is_valid(point):
+                    # Per-state file logging; guarded so a write error can't crash the solver.
+                    try:
+                        log_invalid_state(state=point, obstacle=o)
+                    except Exception as e:
```
raghumanimehta: What are you trying to catch here? 

If you are using genearl `Exception`, that usually means that you don't know what you are defending against. This is bad practice.

raghumanimehta (changes requested):

FireBoyAJ24 (changes requested):
Could we have an info log for the current global and local target waypoint and the target index? This can help us see what the pathfinding is planning towards.

src/local_pathfinding/local_pathfinding/ompl_path.py
AMaharaj16: this specific line doesn't help but it's an extension of a change i made to `is_state_valid` to allow for logging.

src/local_pathfinding/local_pathfinding/node_navigate.py
AMaharaj16: these function calls have lots of depth to them, they can throw lots of errors. this is a final catch to handle anything else that could escape these functions.

PR #936, opened 2026-06-16
Adding On water Land mass and adding improving the Land mass Visualizer

FireBoyAJ24:
### Description
This PR updates the land-mass generation in `local_pathfinding` so that on-water
Testing at Jericho Beach is more accurate with a metre-resolution land coverage, while
keeping the existing production (OSM) land flow intact for launch.

- **Selectable land source.** `pickle_land_data.py` now takes a `--source` argument:
  - `production` — the full OSM Pacific coastline (kilometre resolution), used
    before launch.
  - `on_water` — Vancouver "local-area-boundary" neighbourhood polygons (metre resolution) used as mock land for on-water test plans, since OSM land is too coarse for close-to-shore testing.
- **Coast-aligned edge extrusion.** Added `extrude_edge_to_point()`, which locally pushes the nearest land mass back out to a real feature (the Jericho Pier). The on_water boundary sits ~100 m off the West Point Grey shore, which is too far inland for correct OMPL path generation near the pier; the extrusion adds a coast-parallel section out to the pier so the boat can be deployed past that point. Enabled via `--extrude`.
- **Verification plot.** With `--extrude`, `plot_extrusion()` renders a figure (`west_point_grey_pier_parallel.png`) comparing the extruded land against the pristine original coastline and the on-water test track. matplotlib/pyyaml are optional and the plot is skipped (not failed) if they're missing.
- **Path-anchored I/O.** Shapefile/pickle paths are now resolved relative to the module, so the script runs from any working directory, and the union/repair logic was hardened (`unary_union` + `buffer(0)` + `make_valid`).
- **New Jericho on-water test plan** (`jericho_on_water_test.yaml`) replicating a recorded GPS track off Jericho/Locarno Beach.
- **New Launch test plan** (`launch.yaml) created based on the launch plans for Polaris
- **`TestPlan` fix:** `_land` is now `None` when a test plan omits `land`, instead of constructing an empty `MultiPolygon`.
- Regenerated `land.pkl`, added the on_water shapefiles (+ pristine backup), and documented the generation commands in the land `README.md`.
- Added lat/lon to the boats and added global waypoints (with lat/lon hover) to make it easier to understand what is going on.
- Added OpenStreetMap overlay to make it easier to see how the land polygons look compared to a map
- Removed the interpolation and generate path in `mock_global_path.py` and kept it fixed to what the global paths are in the test plan. This is so that it makes it easier to verify that the visualizer behaves similarly to the test plan. 
- Fixed a bug where the test plan did not get properly loaded for `mock_global_path.py`

### Verification
- [ ] `python3 src/local_pathfinding/land/pickle_land_data.py --source production` regenerates `land.pkl` without error.
    - [ ] `ros2 launch local_pathfinding ... test_plan:="launch.yaml"` verify that the test plan loads properly
- [ ] `python3 src/local_pathfinding/land/pickle_land_data.py --source on_water --extrude` produces the extruded `land.pkl` and writes `west_point_grey_pier_parallel.png`.
    - [ ] `ros2 launch local_pathfinding ... test_plan:="jericho_on_water_test.yaml"` verify that the test plan loads properly
- [ ] Inspected the extrusion plot: extruded land reaches the Jericho Pier target, stays parallel to the shoreline, and the on-water track sits in open water. Based on the generated matplotlib image

image
Above is the On water testing visualizer overlay

image
Above is the launch visualizer overlay

### Resources
- Land source (Vancouver local-area-boundary): https://opendata.vancouver.ca/explore/dataset/local-area-boundary/information/?disjunctive.name
- OSM land data / ODbL: https://www.openstreetmap.org/copyright


src/global_launch/config/globals.yaml
```diff
@@ -15,7 +15,7 @@ mock_global_path:
       gps_threshold: 1.5
 mock_ais:
    ros__parameters:
-      on_water_mock_ais: True
+      on_water_mock_ais: False
```
raghumanimehta, line 18: I would really suggest having a separate `yaml` file specifically for on-water testing. Reason: if we change more stuff, we won't have to revert all the changes back. This will make our life easier. During production use `production.yaml`, during on-water-test use `on-water-test-launch.yaml`.

src/local_pathfinding/land/shp/local-area-boundary.cpg
```diff
@@ -0,0 +1 @@
+ISO-8859-1
```
raghumanimehta, line 1: What is this?

src/local_pathfinding/land/pickle_land_data.py
```diff
@@ -1,13 +1,48 @@
...
 MERCATOR = pyproj.CRS("EPSG:3857")
+# UTM zone 10N — a metre-based CRS over southern BC, used for accurate metre offsets.
+UTM_10N = pyproj.CRS("EPSG:26910")
+
+# Paths are anchored to this file so the script works regardless of the current directory.
+LAND_DIR = os.path.dirname(os.path.abspath(__file__))
+SHP_DIR = os.path.join(LAND_DIR, "shp")
+PKL_PATH = os.path.join(LAND_DIR, "pkl", "land.pkl")
+
+# Pristine (pre-edit) on-water boundaries and the test plan used as overlays in the
```
raghumanimehta, line 30: I don't get this part.

src/local_pathfinding/land/pickle_land_data.py
```diff
@@ -31,23 +66,211 @@ def dump_pkl(object: Any, file_path: str):
...
+    polygon: Polygon,
+    target_lat: float,
+    target_lon: float,
+    section_width_m: float = 300.0,
+    reach_margin_m: float = 150.0,
+    working_crs: pyproj.CRS = UTM_10N,
+) -> Polygon:
+    """Locally extrude a polygon's nearest edge outward to reach a target point.
+
+    Used to push a coastline land mass back out to a real feature (e.g. a Jericho Pier) over a
```
raghumanimehta, line 79: This description could be updated. Also this is a visual detail that is not making much sense to me. It would be helpfult to include a diagram somewhere.

raghumanimehta (changes requested):
This is a really big PR. I am not too worried about anything but I do have some comments. 
While your descriptions are good, i fail to see the intent behind functions like the extrude one. I do know the intent is mentioned "somehwere" but having it in the docs for the function would be nice. 

That being said, i didn't review it in much detail. I am tired and it's looooong.

src/local_pathfinding/land/shp/local-area-boundary.cpg
FireBoyAJ24: It just tells us the encoding of the attribute table dbf file. `.cpg` and `.shx` are present solely to state the format of the other files.

src/local_pathfinding/land/pickle_land_data.py
FireBoyAJ24: The pristine original on-water boundaries were ~100 meters from Jericho beach. I fixed things up now.

FireBoyAJ24: I have updated those comments from pristine to original so it is easier to understand

src/local_pathfinding/land/pickle_land_data.py
FireBoyAJ24: The diagram will be in the README, as we can't add a diagram in this file. I will add the description over the design decisions in the README and, in the future, update the Confluence page.

src/global_launch/config/globals.yaml
FireBoyAJ24: I will create a separate PR on this to make it clear.

FireBoyAJ24:
I am closing this PR and will split into 2 major PRs (one for the on water land mass generation and the other for the visualization). There will also be another smaller PR dealing with the below suggestion by @raghumanimehta: 

> I would really suggest having a separate yaml file specifically for on-water testing. Reason: if we change more stuff, we won't have to revert all the changes back. This will make our life easier. During production use production.yaml, during on-water-test use on-water-test-launch.yaml.

PR #948, opened 2026-06-20
Raghumanimehta/some viz upgrades

raghumanimehta:
### Description

- Resolves #_issue_

### Verification

- [ ] 

### Resources

-


src/custom_interfaces/msg/HelperLatLon.msg
```diff
@@ -2,3 +2,4 @@
 float32 latitude
 # Unit: decimal degrees
 float32 longitude
+HelperHeading heading
```
FireBoyAJ24, line 5: Should we inform @lross03 that `HelperLatLon` has changed their field? Network system's `local_transceiver.cpp` uses this message type in dealing with global paths

src/local_pathfinding/test/test_coord_systems.py
```diff
@@ -84,6 +84,31 @@ def test_true_bearing_to_plotly_cartesian(true_bearing: float, plotly_cartesian:
...
+        (180.0, -90.0),
+        (90.0, 0.0),
+        (45.0, 45.0),
+        (-45.0, 135.0),
+        (135.0, -45.0),
+        (-135.0, -135.0),
+        (1.0, 89.0),
+        (-1.0, 91.0),
+        (-179.0, -91.0),
+        (179.0, -89.0),
```
FireBoyAJ24, line 103: Need to add angles greater than 360, I will add it

FireBoyAJ24 (dismissed):

src/local_pathfinding/test/test_coord_systems.py
raghumanimehta: is this done?

PR #959, opened 2026-06-22
#949: Setting Goal Yaw

### LINKED ISSUE #949: PATH: set goal yaw
### Purpose

I added yaw to OMPL in #948 . This gives it eyes. I have only added the yaw at the start which is based off the heading (in radians converted from true bearing coordinate system to OMPL's cartesian aka regular cartesian coordinates). We should also add yaw at the goal position so that OMPL knows what it is optimizing for. 

We need to worry about the following cases: 
- Next global waypoint is the last global waypoint: this once can be anything 
- Next global waypoint is any other global waypoint: this should be the direction the boat would have to face from the next global waypoint to face next to next global wayptoint. 

### Resources

- `ompl_path.py` file and 
- # 948

ethanhu912:
### Description
- Resolves #949 
- Added yaw at goal position
- Added two functions:
    - One to compute the heading from the next global waypoint to the next-next waypoint
        - If the next global waypoint is the last, sets it to a default heading of east
        - Passes this heading to setYaw.
    - One to offset the goal heading if it is in irons
        - Offsets it towards the closest edge of the no-go zone
- Added tests to ensure that goal heading is in the correct direction if it is not in irons, and that goal heading gets offset to the correct side of no-go zone if in irons

### Verification
- Passes all tests

### Resources
- #948 
- `ompl_path.py`


src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -487,6 +488,84 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+        reference = self.state.reference_latlon
+
+        # Global waypoints are stored in reverse order: index 0 is the final destination.
+        if reference == waypoints[0]:
+            return self._offset_if_in_irons(0.0)
+
+        for i in range(1, len(waypoints)):
+            if waypoints[i] == reference:
+                # reference_latlon sits at (0, 0) in the OMPL frame
+                next_to_next_xy = cs.latlon_to_xy(reference, waypoints[i - 1])
```
raghumanimehta:
```suggestion
                next_to_next_gw_xy = cs.latlon_to_xy(reference, waypoints[i - 1])
```

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -487,6 +488,84 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+                bearing_deg = cs.get_path_segment_true_bearing(cs.XY(0, 0), next_to_next_xy)
```
raghumanimehta: nit: You should directly refer reference_xy to improve readibility rather than `cs.XY(0, 0)`.
not nit: shouldn't the order of argument be reversed here?

raghumanimehta: Please pushback if i am wrong here.

raghumanimehta: Nevermind, the confusion is resolved. I would ask for better naming of variables and a small comment clarifying that reference is the next_gw_xy and that it is calculating a bearing from this to next to next.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -487,6 +488,84 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+                target_yaw_rad = math.radians(cs.true_bearing_to_OMPL_cartesian(bearing_deg))
+                return self._offset_if_in_irons(target_yaw_rad)
+
+        self._logger.warning(
+            "reference_latlon not found in global_path waypoints; using default goal yaw 0.0"
+        )
+        return self._offset_if_in_irons(0.0)
+
+    def _offset_if_in_irons(self, ompl_yaw_rad: float) -> float:
```
raghumanimehta: rename to include `rad` in the name as you did for the others.

raghumanimehta: rename `ompl_yaw_rad` to `target_yaw_rad`. Seems clearer to me. Thoughts?

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -487,6 +488,84 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+
+        The planning-wind snapshot (path_generated_wind when available, otherwise the
+        current true wind) is used so the goal yaw, motion validator, and objective all
+        share the same wind baseline.
+
+        Args:
+            ompl_yaw_rad: Proposed goal yaw in OMPL Cartesian radians.
+
+        Returns:
+            float: A goal yaw in OMPL Cartesian radians outside the wind no-go zone.
```
raghumanimehta, line 550: Should specify what the OMPL Cartesian coordinates are for readibility.

src/local_pathfinding/local_pathfinding/ompl_path.py
```diff
@@ -487,6 +488,84 @@ def _init_simple_setup(self, land_multi_polygon) -> og.SimpleSetup:
...
+
+        # OMPL Cartesian yaw → boat heading in true-bearing radians, bounded to (-pi, pi].
+        boat_heading_rad_gc = cs.bound_to_180(
+            cs.cartesian_to_true_bearing(ompl_yaw_rad, rad=True), rad=True
+        )
+        twa = wcs.get_true_wind_angle(boat_heading_rad_gc, tw_dir_rad_gc)
+
+        if abs(twa) < NO_GO_ZONE:
+            # Upwind irons: snap to the closer ±NO_GO_ZONE edge.
+            new_twa = NO_GO_ZONE if twa >= 0 else -NO_GO_ZONE
```
raghumanimehta: We shouldn't be setting it eactly to `NO_GO_ZONE`. Consider adding a small margin to make sure OMPL doesn't crash because of validitiy not passing.

src/local_pathfinding/test/test_ompl_path.py
```diff
@@ -503,3 +505,134 @@ def test_get_remaining_cost_projection_logic(fresh_ompl_path):
...
         f"cost from start ({cost_from_start:.3f}) and cost from end ({cost_from_end:.3f})"
     )
+
+
+def _yaw_test_self(waypoints, target, tw_dir_deg, tw_speed_kmph=10.0):
+    """Minimal stand-in for an OMPLPath instance, exposing only the attributes the
+    goal-yaw helpers read. Avoids running the OMPL solver in every test.
+    """
+    wind = Wind(tw_speed_kmph, tw_dir_deg)
+    state = SimpleNamespace(
```
raghumanimehta, line 521: nit: consider using `Mock` instead of `SimpleNamespace`.

src/local_pathfinding/test/test_ompl_path.py
```diff
@@ -503,3 +505,134 @@ def test_get_remaining_cost_projection_logic(fresh_ompl_path):
...
+        global_path=Path(waypoints=waypoints),
+        reference_latlon=target,
+        path_generated_wind=wind,
+        current_tw=wind,
+    )
+    fake = SimpleNamespace(state=state, _logger=RcutilsLogger())
+    fake._offset_if_in_irons = types.MethodType(ompl_path.OMPLPath._offset_if_in_irons, fake)
+    return fake
+
+
```
raghumanimehta, line 533: In the test file, why not use `pytest mark parametrize`?

raghumanimehta: Never mind this, I got codex to care of the problems in tests. Focus on ohter comments.

raghumanimehta (changes requested):
Well done @ethanhu912 ! This was good work. I had some problems with readibility and that is why i am requesting changes but this is a high quality contribution. 

Please update the tests when you add the buffer near NO_GO_ZONE in `_offset_if_in_irons`.

src/local_pathfinding/local_pathfinding/ompl_path.py
ethanhu912: Agreed. Will change right now

raghumanimehta (dismissed):

Issue #937, opened 2026-06-17
PATH: Improve Global Path Handelling


FireBoyAJ24:
### Purpose

From past discussions, we have seen a lot of confusion over how the global path is sent, received, handled and consumed. There has been past work on the receiving and handling of global paths; however, with the addition of `node_mock_global_path` and the recent work on `node_global_path`, things have gotten mixed up. The documentation is still out of date, and there are concerns based on this #936 PR, that there are numerous bugs in how the global path is handling and how the order of the waypoints could vary, as I haven't enforced it properly in the test plans.

### Changes

In short, we first need to finalize the design specification for the global path and document the software's quirks. Then we need to identify discrepancies between the design intent and the local pathfinding codebase. Finally, we need to make the necessary fixes.

Based on my work in #963, I have already begun removing interpolation and generating paths, as we can verify those values before sending them to the boat via satellite or prior to deployment. We should move towards merging the global path nodes as stated above so that the design remains the same between development and production. While also reducing duplicate code and nodes.

We will need to review and fix any issues with the global path handling in `node_navigate` and add additional test cases to handle various scenarios the pathfinding software faces.

### Resources

- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1763868673/Design+Spec+-+Global+Pathfinding
- `src/local_pathfinding` directory

PR #962, opened 2026-06-24
Raghumanimehta/937 updated global path handling

### LINKED ISSUE #902: PATH: Global Path instantiation
### Description

According to https://github.com/UBCSailbot/sailbot_workspace/blob/main/src/global_launch/config/README.md#local-pathfinding-parameters , Global Pathfinding expects a path to csv file that has the Global coordinates. However, it seems there is no way / where to pass this parameter and there is no node that actually reads it and does anything meaningful. This could be the reason that local pathfinding doesn't launch in production; the local pathfinding in development uses test plan + mock_global_path.py . Investigate this and fix it. 

#### Expected Behavior

- There is a place to pass this `path to csv` and the said node (or whatever) reads the file and outputs the global path properly. 
- Ensure it reaches local pathfinding. 

#### Acceptance criteria: 
- Document what the issue was and what was fixed. Update the docs in the repository on how to update the global path.

raghumanimehta:
### Description

- Resolves #902 #937 
- Nuked node_global_path.py
- Simplified global path process (checkout the comment above GlobalPath class) 
- added tests to make sure path storage works properly
- added a backup path for worst case scenario where we are not getting paths or we can't write to file
### Verification

- [x] Tests 
- [ ] Manual verification. After issue 967 is resolved.


src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -8,18 +11,68 @@
...
+Sailbot leaves gp unset and the existing inactive-input path publishes sail=False.
+
+This assumes persisted main and backup paths use the same reverse-order convention as NET paths.
+For backup paths, choosing the closest waypoint is intentionally a pragmatic resume heuristic: it
+does not try to infer whether the boat has already passed that waypoint along the route segment.
+Once selected, the path still advances toward the destination by decrementing the index.
+
+Once a path is active, navigation advances by decrementing the index. When the index drops below
+zero, the global path is exhausted and Sailbot disables sail instead of cycling between waypoints.
+"""
```
FireBoyAJ24, line 61: We should discuss this with the rest of the team over shutting down path planning compared to cycling between waypoints. Should Polaris go to the last waypoint after it moves x distance from the last waypoint?

src/local_pathfinding/global_paths/path_builder/path_builder.py
```diff
@@ -22,17 +22,98 @@
...
-)
+import local_pathfinding.coord_systems as cs
 
 app = Flask(__name__)
 
 DEFAULT_DIR = "/workspaces/sailbot_workspace/src/local_pathfinding/global_paths/"
 
 
+def write_to_file(file_path: str, global_path: Path, tmstmp: bool = False):
+    """Writes a global path to a csv file."""
```
FireBoyAJ24, line 33: What is the purpose of this file now?

FireBoyAJ24: It seems to not be used anywhere and is only discussed in the README.md of this folder.

src/global_launch/config/README.md
```diff
@@ -37,28 +37,21 @@ ROS parameters common across all ROS nodes in the network.
...
 - _Datatype_: `string`
 - _Acceptable Values_: Any valid filepath to a properly formatted csv file.
 
 **`global_path_interval_spacing_km`**
 
-- _Description_: The upper bound on the spacing between consecutive points in
-  the global path in km. The path read from the csv file is interpolated so
-  that no two consecutive waypoints are farther apart than this value. This is
-  a global parameter shared by all nodes.
+- _Description_: Global waypoint spacing parameter shared by local pathfinding
```
FireBoyAJ24: `global_path_interval_spacing_km` ros param is not used in `node_navigate` so it is kind of confusing why we are keeping this here.

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_global_path.py
```diff
@@ -63,14 +62,22 @@ def __init__(self):
...
     # Timer callbacks
     def global_path_callback(self, gps: ci.GPS):
         # Publish the test plan's global path exactly as defined. This node does not interpolate
         # between waypoints or generate a path from the boat's position — it relies solely on the
         # waypoints provided in the test plan. The path is still published on every GPS message so
         # that node_navigate receives the waypoints even if this node launched first.
         msg = self.global_path
-        self.get_logger().debug(f"Publishing mock global path: {gp.path_to_dict(msg)}")
+        self.get_logger().debug(f"Publishing mock global path: {self._path_to_dict(msg)}")
         self.global_path_pub.publish(msg)
```
FireBoyAJ24, line 79: The global path publisher will also publish whenever it receives GPS messages (~1-2 Hz). This can cause the node_navigate to keep writing to the CSV file. It doesn't seem like it will affect the global path target waypoint, but it could possibly cause the file to become corrupt if there is a sudden power cut. The chance here is higher compared to when the global path is sent once and stored once. While getting updated as the global path changes based on receiving the message from the satellite.

FireBoyAJ24 (commented):
My only implementation concern is why we are keeping ``global_path_interval_spacing_km` and the frequent sending of the global path when we recieve gps message frequently

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: Who do you want to ask?

src/local_pathfinding/global_paths/path_builder/path_builder.py
raghumanimehta: we write the path to the csv file to ensure that if the boat power cycles then we should have the last path that we were following. storing it only in memory means that we can lose all the data if the boat power cycles.

src/global_launch/config/README.md
raghumanimehta: good catch. should be removed.

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_global_path.py
raghumanimehta: This node will be reworked as per #967

raghumanimehta:
> My only implementation concern is why we are keeping ``global_path_interval_spacing_km` and the frequent sending of the global path when we recieve gps message frequently

I think you are confusing the mock global path node and the actual global path node. The former is in PATH for testing, and i agree its design is not representative of how NET will send global path. The latter only sends data when NET ask it to, not in a loop.

src/local_pathfinding/local_pathfinding/node_navigate.py
FireBoyAJ24: @MichaelGreenough and @captain-ubcsailbot

FireBoyAJ24:
> > My only implementation concern is why we are keeping ``global_path_interval_spacing_km` and the frequent sending of the global path when we recieve gps message frequently
> 
> I think you are confusing the mock global path node and the actual global path node. The former is in PATH for testing, and i agree its design is not representative of how NET will send global path. The latter only sends data when NET ask it to, not in a loop.

I would recommend having the mock global path node to mirror as close to a behaviour to NET so that we can catch any more bugs like QoS mismatch, frequent publishing, etc. This should be of note in #967.

FireBoyAJ24 (approved):

FireBoyAJ24 (approved):

PR #969, opened 2026-06-25
removed doPost() global_path branch and tests

### LINKED ISSUE #961: NET: Remove redundant POST request handler in `remote_tranceiver.cpp`
We don't send global_paths via POST to NET anymore. 
Instead, we will directly use ground control to send global path waypoints that is properly spaced and verified. 

### Old architecture:
PATH used a csv to store the global paths. It would make a POST request to NET which would then send the path to Irridium, which would then send the data to the boat using local_tranceiver which would publish the data on the ros topic `global_path` which would then be read by the PATH. 

### New Architecture:
We manually send data via Irridium to local_tranceiver which is read by PATH. PATH will also write the new path on disk (csv) to ensure that paths are not lost. If we just keep in memory and the boat loses power, we lose the paths as well. Instead we store the path to disk to ensure some path is always available even if we can't connect to satellite for long periods of time.

raghumanimehta:
### Description

- Resolves #961 
- Removed the `GLOBAL_PATH` branch of `doPost()` function
- Updated the tests accordingly (mostly deletions). 
- Now doing `POST` with target = `\global_path` returns Bad request.


lross03 (approved):
lgtm

PR #972, opened 2026-06-26
Updated CAN Frames

### LINKED ISSUE #809: Update CAN frames as of 02/2026
### Purpose

There have been some changes to CAN frames on the ELEC side, and these changes need to be reflected in software.

### Changes

Read through the CAN frames document and verify that all existing frames are correct, and any newly created frames are added to software. Currently, I know we dont handle error values from ELEC, so for those, please ensure can_transceiver logs the error, but does not broadcast these values to the ROS system.

Frames are found in can_frame_parser.cpp

https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1827176527/CAN+Frames

### LINKED ISSUE #964: NET: set heading bit to 1 to inform ELEC to stop trying to turn the boat
### Purpose

Whenever PATH determines that it is unsafe to sail, it sets `DesiredHeading`'s `sail` field to `False`. 
The reason for this could be multiple: 
- insufficient sensor input for PATH
- impending collision such that no path is solvable. 
- File I/O crashes in PATH 
There could be more reasons as we continue to develop the system. 

### Changes

- Not sure since I am not aware of the NET codebase, but should be wherever you reading desired heading and convey that info to ELEC via CAN.

### Resources

- https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1827176527/CAN+Frames#Frames 
- From the above page, particularly this part is of interest to us: 

Image
This was under Can frame id 0x001

Apps247:
Updated CAN frames based on Confluence Page dated Oct 29, 2025. Resolves #809 

It seemed like some of the previous code was buggy, and the `toLinuxCan()` vs the decoding functions performed slightly different calculations. I made them match, but I'm unsure if it's fully correct.

Also added the logic to set the steering enable bit to 1 (which actually disables steering) if PATH sets `sail` to false. Resolves #964


src/network_systems/projects/can_transceiver/src/can_frame_parser.cpp
```diff
@@ -749,17 +749,35 @@ DesiredHeading::DesiredHeading(const CanFrame & cf) : DesiredHeading(static_cast
...
 : BaseFrame(id, CAN_BYTE_DLEN_),
-  heading_(utils::boundTo360(ros_desired_heading.heading.heading)),
+  heading_(
+    (ros_desired_heading.steering & 0b10000000) != 0  //NOLINT(readability-magic-numbers)
+      ? ros_desired_heading.heading.heading
+      : utils::boundTo360(ros_desired_heading.heading.heading)),
   steering_(ros_desired_heading.steering)
 {
+    // Set steering enable bit based on the sail flag in the ROS message
+    uint8_t steering_disabled_bit_mask = 0b01000000;  //NOLINT(readability-magic-numbers)
```
raghumanimehta, line 777: I just reviewed this part.

raghumanimehta (dismissed):
I approve this for the part that concerns pathfinding. I haven't looked at the rest of the code so don't merge befor @lross03 and / or @Jng468's approval.

lross03:
Please ensure the filtering for debug frames is correct per CAN frame documentation. For example, the pressure sensors (now removed entirely) occupy the frame IDs of the heart beat frames. I filtered them out as a hotfix for the OWT, but we need to ensure all the software frame IDs match the documentation and are filtered correctly.

https://github.com/UBCSailbot/sailbot_workspace/blob/1a23f24bccfbcc30183d724ac33d9f1d4a4e0a00/src/network_systems/projects/can_transceiver/src/can_transceiver.cpp#L30-L37

lross03 (changes requested):
Above comment

lross03 (approved):
lgtm

PR #975, opened 2026-06-27
Prevented multiple messages from sending on single call

### LINKED ISSUE #934: Debug send should not repeatedly send in a loop.
### Purpose

Sometimes debug send will get stuck in a loop and send its message 20 times which eats into our credits. For all future credit-costing messages, we should either retry fewer times or return always.

Apps247:
Resolves #934

Multiple sends were happening in the same call because something in the loop failed and caused a retry *after* the send command completed successfully.

Changed the failures after a successful send command to immediately return instead of retrying.

Also changed the retry limit from 20 to 5.

### DIFF
--- a/src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
+++ b/src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
@@ -171,7 +171,7 @@ bool LocalTransceiver::send()
     std::string write_bin_cmd_str = AT::write_bin::CMD + std::to_string(data.size());  //according to specs
     AT::Line    at_write_cmd(write_bin_cmd_str);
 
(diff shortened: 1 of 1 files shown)


raghumanimehta (commented):
Question: Does returning false lead to more frequent failures?

lross03 (approved):
lgtm

lross03:
> Question: Does returning false lead to more frequent failures?

no. we also cant determine if the message sent but a separate failure happened without some kind of handshake with the receiving end.

Apps247:
If it becomes a problem, then we can go back to retrying for actual send() when we launch. We can reduce the limit from 20 to 5 so credit usage doesn't blow up

PR #978, opened 2026-06-29
Validates the global path before local transceiver publishes the path#832

### LINKED ISSUE #832: Validate global path object before local transceiver publishes
### Purpose

Currently, the receive() function for local transceiver parses a message received via the satellite modem into a protobuf global path object, then puts the waypoints into a HelperLatLon message and publishes.

Before we publish this message, we should check that the contents are valid waypoints according to what PATH expects.

### Changes

- Validate the global path waypoints before we publish them to ROS

arkumar57:
### Description

- Resolves #832

- Wrote a function named `validateGlobalPathWayPoints` in `local_transceiver.cpp` which returned boolean. The function would take in the path to be published as input and would return `True` if the latitude and longitude values for the global path fell within the valid range as defined. Otherwise, it would return `false`.
- The result of this function was used in the `receive` function in `local_transceiver.cpp`. The `receive` either returns the path to published or `-1` if it was `True` or `False` respectively.

### Verification

- Wrote a test case named `parseInMsgInvalid` in the `test_local_transceiver.cpp` to validate the functionality of the code.

### Resources

- https://github.com/UBCSailbot/sailbot_workspace/issues/832


src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
```diff
@@ -597,7 +568,29 @@ custom_interfaces::msg::Path LocalTransceiver::receive()
...
+
+bool LocalTransceiver::validateGlobalPathWayPoints(const custom_interfaces::msg::Path & path_to_publish)
+{
+    if (path_to_publish.waypoints.empty()) {
+        return false;
+    }
+
+    for (const auto & waypoint : path_to_publish.waypoints) {  //NOLINT
+        if (
+          (waypoint.latitude < -90.0 || waypoint.latitude > 90.0) ||      //NOLINT(readability-magic-numbers)
```
raghumanimehta: Probably shouldn't use NOLINT here... For readibility, it would be better to let the reader know that we are talking about max and min lattitude.

src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
```diff
@@ -550,23 +512,17 @@ custom_interfaces::msg::Path LocalTransceiver::receive()
...
-                    log_debug_("DEBUG: MT status 0 continue");
-                }
-                continue;
-            } else if (rsp.MT_status_ == 1) {  //NOLINT
-                break;
-            } else if (rsp.MT_status_ == 2) {
-                continue;
-            }
-        } else {
+        if (rsp.MO_status_ != 0) {
```
raghumanimehta, line 558: Good change

raghumanimehta (dismissed):
Looks good to me, but wait for @lross03 's review. 

In the meanwhile, i would appreciate if you get rid of the NOLINTS for magic numbers. The for loop ones are fine.

raghumanimehta:
my changes broke the Ci... i will fix it tomorrow.

src/network_systems/projects/local_transceiver/test/test_local_transceiver.cpp
```diff
@@ -638,69 +639,96 @@ TEST_F(TestLocalTransceiver, checkCache)
...
-// }
+    if (global_path.waypoints_size() > 0) {
+        ASSERT_EQ(global_path.waypoints_size(), sample_data.waypoints_size())
+          << "Mismatch in number of waypoints received.";
+        ASSERT_EQ(global_path.waypoints(0).latitude(), holder);
+        ASSERT_EQ(global_path.waypoints(0).longitude(), holder);
+    } else {
+        std::cout << "No waypoints received." << std::endl;
+    }
+}
```
lross03: Please dont uncomment these tests. we dont want to use the virtual iridium tests for the foreseeable future

src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
```diff
@@ -621,8 +618,11 @@ custom_interfaces::msg::Path LocalTransceiver::parseInMsg(const std::string & ms
 
     for (auto waypoint : path.waypoints()) {
         custom_interfaces::msg::HelperLatLon helperLatLon;
-        helperLatLon.set__longitude(waypoint.longitude());
-        helperLatLon.set__latitude(waypoint.latitude());
+        const float latitude  = waypoint.latitude();
+        const float longitude = waypoint.longitude();
+
+        helperLatLon.set__latitude(latitude < -90.0f || latitude > 90.0f ? -1.0f : latitude);
+        helperLatLon.set__longitude(longitude < -180.0f || longitude > 180.0f ? -1.0f : longitude);
```
lross03: With Raghumani's change where local transceiver wont publish a message if the waypoint array is empty, we should also instead just return a path object with empty waypoint array. This will get filtered out anyways. Here we can add a error log statement indicating the longitude/latitude is invalid.

src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
```diff
@@ -552,10 +552,7 @@ custom_interfaces::msg::Path LocalTransceiver::receive()
 
         if (rsp.MO_status_ == 0) {
             if (rsp.MT_status_ == 0) {
-                if (log_debug_) {
-                    log_debug_("DEBUG: MT status 0 continue");
-                }
-                continue;
+                return parseInMsg("-1");
```
lross03: See other comment.

lross03 (changes requested):

src/network_systems/projects/local_transceiver/src/local_transceiver.cpp
lross03: empty waypoint array instead of -1 value*

raghumanimehta: agreed.

src/network_systems/projects/local_transceiver/test/test_local_transceiver.cpp
lross03: Please comment this test out again

lross03 (changes requested):
Please comment out the test I mentioned and ensure your PR passes the CI pipeline tests

lross03 (approved):

PR #987, opened 2026-07-01
Fix #967: Removed repeated global path republishes + Button to change global waypoints during run

### LINKED ISSUE #967: PATH: Update `node_mock_global_path.py`
### Problem
`node_mock_global_path.py` publishes the same global path on every GPS callback even though it does not use GPS. With the new global-path state handling, repeated publishes can reset navigation progress in node_navigate.

### Proposed fix
Replace the GPS-triggered republish loop with a startup publish using transient-local QoS, or publish once after startup/first GPS if needed for launch ordering.
### Acceptance criteria
- Mock GP no longer republishes continuously on GPS messages.
- We can control when we want to send messages by pointing it to read the a csv/yaml (i believe yaml would be more convenieint as it's ROS native) in the mock_nodes folder and sending it via the `global_path` topic. 
- Once you are done with this task, please contact me. I will ask you to test global path features in production nodes. I will provide more details then. 
Please refer to this: https://github.com/UBCSailbot/sailbot_workspace/pull/962#issuecomment-4803859770 and ask questions if you have any doubts.

MilcToast:
### Description

- Resolves #967 

### Changes Made
- Removed the GPS subscription and `global_path_callback` from `node_mock_global_path.py`, the node no longer republishes the global path on every GPS message
- Added a one-time publish at startup via `publish_global_path`, called before `rclpy.spin`
- Added a CSV file (global_paths/mock_global_path.csv) as a path source, with previous working path kept as a fallback if the CSV fails to load
- Added a `reload_token` ROS 2 parameter and `_on_set_parameters` callback to support runtime reloading, setting `reload_token` to any new value re-reads the CSV and republishes the global path
- Added a "Mock Global Path" control panel to the visualizer with a "Reload CSV" button that triggers the reload via ros2 param set /mock_global_path reload_token <n_clicks>

### Verification

-  Confirmed that node_mock_global_path no longer continuously republishes on GPS messages (only logs once)
- Confirmed that the global path is published once at startup and received by subscribers
- Confirmed that editing mock_global_path.csv and clicking "Reload CSV" in the visualizer triggers a republish with the updated waypoints
- Confirmed that the visualizer displays a success/error message after clicking "Reload CSV"


raghumanimehta (changes requested):
my bad, the requirements were not clear ig... we want to use the global path from the test plans.

I think a good design could be editting the mock_global_path.csv with new values then clicking the button on gui to change the global path to the one from csv i.e. stop using the one from test plan. 

any questions?

PR #992, opened 2026-07-04
Add a kill_ais_node ros param

### LINKED ISSUE #907: Allow AIS handling to be disabled
### Purpose

Since we are not able to test with the AIS before the next on water test, we need to be able to disable AIS handling, as it may cause segfaults.

### Description

Add a parameter to enable/disable AIS from CAN on the fly, or just disable AIS handling on main for now.

FireBoyAJ24:
### Description

- Resolves #907

I have added a ros2 param called `kill_ais_can`, which is set to false by default, as we want to read the AIS ship messages from CAN

### Verification

- [x] Check that we can run `ros2 param set /can_transceiver_node kill_ais_can true` and check by `ros2 param get /can_transceiver_node kill_ais_can` that kill_ais_can changes from it's orignal false to true


lross03 (approved):
lgtm

PR #1004, opened 2026-07-12
956 - Move from Generic Sensor to specific pH, Salinity, and Hydrophone sensor data

### LINKED ISSUE #956: WEB: Move from Generic Sensor to Specific Sensors We are Using
### Purpose

Currently, we have a section called "Generic Sensors". We should move away from that and specify the sensors we are currently storing from the data I provided in Slack (e.g, pH, hydrophone (It is on the team drive), etc.).

### Description

This issue can be broken down into two tasks.
1. Add the specific data to the https://[sailbotpolaris.com/download](https://sailbotpolaris.com/download) section
2. Add the specific data to the graphs in the https://sailbotpolaris.com/

These two tasks can be split into two simple pull requests, but should be completed soon so we can add more data to them by July 4th.

matthewcflam:
### Description
This PR resolves #956. It splits the Generic Sensor Data (which has been split on the backend) into pH, Salinity, and Hydrophone data on the website. Graphs were added for pH and Salinity, and downloads were added for all 3.

Graph layout files were changed to allow users who already configured their graph layouts (which persist) to see the new graphs.

### Verification
Local simulation files and cucumber tests were all added and passed.

image

image

Made with simulation data

Note: The diff looks MASSIVE, but a lot of it is just copy and paste.


raghumanimehta:
The graph pictures look really good! Thanks for the work.

src/website/lib/redux/rootSaga.ts
```diff
@@ -20,6 +23,9 @@ export function* rootSaga() {
     batteries: new BatteriesSagas().forkSagas(),
     windSensors: new WindSensorsSagas().forkSagas(),
     genericSensors: new GenericSensorsSagas().forkSagas(),
+    tempSensors: new TempSensorsSagas().forkSagas(),
```
minjunminji, line 24: The PR is titled *move from* Generic Sensors, but GenericSensors is only removed from the datasets route and the download UI — the store, saga, reducer, model, and `/api/generic-sensors` route all remain. Since `forkSagas()` starts every saga, `genericSensors` still polls `/api/generic-sensors` on every interval and dispatches into a reducer nothing reads anymore (dead polling + dead state).

Either fully remove GenericSensors (store dir, model, route, the `endpoints.ts` / `interfaces.ts` / `given.ts` entries, the `rootReducer` + `rootSaga` registrations, and `genericsensors.feature`), or if it's intentionally kept during the transition, please note that in the PR description.

minjunminji (commented):
Reviewed the full diff and traced it against the backend and simulation. Overall this is a clean, well-structured slice — the three sensors mirror the existing WindSensors pattern, and I confirmed the backend (`sailbot_db.cpp`) already writes identical document shapes to the same `temp_sensors` / `ph_sensors` / `salinity_sensors` collections, so the models, routes, and seed data line up end-to-end.

`mergeMissingGraphIds` is a nice touch: it stops cached sessionStorage layouts from silently hiding the newly added graphs, and it's unit-tested.

Two things I'd like addressed before merge — left inline below.

yuekaifred (dismissed):
yeah ok

yuekaifred:
Good job matthew! This is really huge. You cleanly owned an implementation covering 50% of the website

PR #1005, opened 2026-07-12
PATH: Dynamic Test Plan

### LINKED ISSUE #995: PATH: dynamic test plan
### Purpose

@FireBoyAJ24 worked on test plans. @jb3982 further improved them to let us run the test plans sequentially. Now, we want to go one step further by making the test plans dynamic.

The idea is to define mock inputs in a detailed YAML file using timestamped `events`. Each event describes when a mock input should change relative to the start of the test plan. For example, `mock_wind_sensor` could contain a list of wind updates, where each update includes the wind direction in degrees, wind speed in km/h, and a relative timestamp for when that update should be applied.

The timestamps should be strictly monotonic, meaning each new event must occur after the previous one. This makes the test plan deterministic and easier to reason about.

Example format:

```yaml
events:
  mock_wind_sensor:
    - timestamp: 0.0
      direction_deg: 90
      speed_kmph: 12

    - timestamp: 5.0
      direction_deg: 110
      speed_kmph: 18

    - timestamp: 10.0
      direction_deg: 135
      speed_kmph: 20
```

A similar structure can be used for other mock inputs such as `mock_gps`, `mock_global_path`, `mock_ais`, and other mocked sensors or planning inputs. Each mock input would define its own list of timestamped values, allowing the test plan to simulate changing world conditions over time.

You might be wondering, "why do this?".  Glad you asked. We will use AI to generate this file to mimic ocean data or we can even take the OWT data pass it to AI and convert it to this file for ease of use. *But* we will *REQUIRE* strong  checks on plausibility for example angles being within the right range or the wind speed not going to let's say 5000kmph,

ethanhu912:
### Description
- Resolves #995 
- Implemented events to make test plans dynamic:
  - Events (wind changes, drift, etc.) live in the same YAML file where the original test plans were written. Each event gets its own timestamp for when it will be applied.
  - `test_plan.py` has added functionality to parse these dynamic events as well as the initial snapshot. Also rejects unrealistic events and sends an error. 
  - Added a event_dispatcher.py to keep track of which events have been applied and discards them. This ensures that the previous event does not get reapplied over and over again while we are waiting for the next event.
  - Each mock node keeps track of time and polls the dispatcher to see if any new events are to be applied.
- Added unit tests for test_plan.py parsing data (allows realistic events to pass, rejects unrealistic events). Ensures that existing test plans still continue to function. Also added unit tests for event_dispatcher.py ensuring that events are applied correctly and events that have been already used are discarded.

### Verification
- Unit tests all pass
- Created a sample dynamic test plan. Launched the system and used rosbag recordings to record the data received by the mock nodes and verified that all changes were being passed.


src/local_pathfinding/local_pathfinding/global_path_storage/main_global_path.csv
```diff
@@ -1,6 +1,3 @@
 latitude,longitude
-49.276302337646484,-123.19969177246094
-49.28141403198242,-123.19366455078125
-49.28163528442383,-123.18553161621094
-49.281734466552734,-123.1733169555664
-49.2897834777832,-123.17166900634766
+48.5,-125.0
```
raghumanimehta: Revert this edit please.

src/local_pathfinding/local_pathfinding/mock_nodes/dispatch_event.py
```diff
@@ -0,0 +1,30 @@
...
+from typing import Any
+
+
+class EventDispatcher:
+
+    def __init__(self, events: Iterable[Any]) -> None:
+        self._events: tuple[Any, ...] = tuple(events)
+        self._cursor: int = 0
+
+    def pop_fired(self, elapsed_sec: float) -> Iterator[Any]:
```
raghumanimehta, line 19: nit:would `dispatch_event` be a better name?

src/local_pathfinding/local_pathfinding/mock_nodes/node_mock_ais.py
```diff
@@ -51,42 +53,39 @@ def __init__(self):
...
 
-            for ship in ais_ships:
-                csv_ship_ids.append(ship.id)
-                current_ship_ids = [ship.id for ship in self.ships]
-                if ship.id > 0 and ship.id not in current_ship_ids:
-                    self.ships.append(ship)
-                    msg.ships.append(ship)
+        self._start_monotonic_sec = time.monotonic()
+        self._ais_dispatcher = EventDispatcher(test_plan.ais_events)
+        # Skip physics on the tick right after a snapshot so the published state matches the
```
raghumanimehta: not sure what this means. Can you elaborate it in src code?

src/local_pathfinding/test_plans/scenario_events_demo.yaml
```diff
@@ -0,0 +1,42 @@
...
+events:
+  mock_wind_sensor:
+    - {timestamp: 3.0, direction_deg: 90, speed_kmph: 25.0}
+    - {timestamp: 12.0, direction_deg: -45, speed_kmph: 15.0}
+
+  mock_ais:
+    - timestamp: 6.0
+      ships:
+        - {id: 1, lat: 49.31, lon: -123.21, heading: 180.0, speed: 10.0, rot: 0, width: 1.0, height: 1.0}
+    - timestamp: 15.0
```
raghumanimehta, line 31: Quick question: does this append to the list or does it overwrrite? 
I haven't reviewed the whole code yet. 

I am wondering if which design would be better for AIS. What are your thoughts?

raghumanimehta (changes requested):

src/local_pathfinding/test_plans/scenario_events_demo.yaml
ethanhu912: Currently it overwrites. I believe overwriting is slightly better, since it is much clearer to see exactly what AIS looks like at each timestamp and verify using rosbag recordings (you can directly compare the message to the events).

src/local_pathfinding/local_pathfinding/global_path_storage/main_global_path.csv
```diff
@@ -1,6 +1,4 @@
 latitude,longitude
-49.276302337646484,-123.19969177246094
-49.28141403198242,-123.19366455078125
```
AMaharaj16: why'd these two get deleted?

src/local_pathfinding/test_plans/test_plan.py
```diff
@@ -109,3 +186,216 @@ def tw_speed_kmph(self):
...
+    timestamps: list[float] = []
+    for i, item in enumerate(raw):
+        context = f"mock_global_path event[{i}]"
+        _check_fields(
+            item, _GLOBAL_PATH_EVENT_FIELDS, _GLOBAL_PATH_EVENT_FIELDS, context
+        )
+        ts = float(item["timestamp"])
+        waypoints_raw = item["waypoints"]
+        if not isinstance(waypoints_raw, list) or not waypoints_raw:
+            raise ValueError(f"{context} 'waypoints' must be a non-empty list")
```
AMaharaj16: a route with one waypoint passes. maybe add a check for if waypoint list length is <2.

ethanhu912: what's the issue with only having one waypoint?

raghumanimehta: Is this global or local? @ethanhu912

ethanhu912: Global waypoints

raghumanimehta: because newly received global waypoints are only accepted if len >= 2. See `node_navigate.py` for more details.

raghumanimehta:
CI/CD is still broken here

src/local_pathfinding/test_plans/test_plan.py
ethanhu912: this should be fixed

ethanhu912:
> CI/CD is still broken here

not too sure why this is happening. colcon-test shows it is an error with network_systems and more specifically with TestLocalTransceiver.DebugSendAT. rerunning without changing anything yields CI/CD passing. i dont think any of my changes would have introduced this. @raghumanimehta

a description from github copilot - Root Cause: The test expects the echo server to receive and return a payload, but it's receiving an empty body. This suggests the webhook POST request isn't being sent with the proper payload data.

raghumanimehta:
> > CI/CD is still broken here
> 
> not too sure why this is happening. colcon-test shows it is an error with network_systems and more specifically with TestLocalTransceiver.DebugSendAT. rerunning without changing anything yields CI/CD passing. i dont think any of my changes would have introduced this. @raghumanimehta
> 
> a description from github copilot - Root Cause: The test expects the echo server to receive and return a payload, but it's receiving an empty body. This suggests the webhook POST request isn't being sent with the proper payload data.

Whenever this happens, rerun the CI... not your fault. I didn't open the logs mb

PR #1007, opened 2026-07-15
Fix #1001: Update Obstacles with new AIS data

### LINKED ISSUE #1001: PATH: AIS updates don't update the obstacle geometries
### Description

When we get new AIS messages we update `LocalPathState`'s `ais_ships` but that doesn't update the obstacle geometries that is used in the `in_collision_zone()` function. That is, for 10 minutes (at max), we don't have updated obstacle information. This can be problemeatic. 

#### Expected Behavior

Whenever we update the `ais_ships`, we should also update the obstacle geometries eg. if we get a new AIS signal in the time period, we should have the updated obstacle information to include this new AIS. 

#### Actual Behavior

We do have the udpated `ais_ships` but we don't have obstacles in the `LocalPath` so `in_collision_zone` is essentially missing that. 

### Relevant Code

- `local_path.py` : `in_collision_zone` and `LocalPathState
- `obstacles.py`
- `ompl_path.py` : `init_simple_setup`

### Potential Solutions

A `LocalPathState.update_obstacles()` method is the cleanest lifecycle fix.
It should:
- Preserve the existing land obstacle.
- Rebuild boat obstacles from the latest self.ais_ships.
- Use the state’s current reference_latlon, position, speed, and current AIS dimensions.
- Replace the boat-obstacle list atomically, naturally adding/removing ships.
Then `LocalPath.update_if_needed()` becomes:
``` bash 
self.state.update_state(...)
self.state.update_obstacles()
must_change_reason = self.must_change_path(...)
```

MilcToast:
### Description

- Resolves #1001

### Changes Made
- Added `LocalPathState.update_obstacles()`, which rebuilds boat obstacles from the latest `self.ais_ships` while preserving the existing `Land` obstacle instance rather than rebuilding it
- Called `update_obstacles()` from `LocalPath.update_if_needed()` immediately after `state.update_state`, so obstacle geometry is refreshed on every AIS update
- Left land geometry untouched between replans (only `update_sailbot_data` is called on the existing `Land` instance)
- Added a guard for the case where `update_obstacles()` is called with no existing `Land` obstacle in `self.obstacles`, since this should only happen before the first successful replan.

### Verification
- Added parametrized test `test_update_if_needed_syncs_boat_obstacles_with_ais_ships` with:
   - test where a new ship in `ais_ships` results in a new `Boat` obstacle being added
   - test where a ship no longer present in `ais_ships` has its `Boat` obstacle removed
   - test where all boat obstacles are removed when `ais_ships` is empty, while the `Land` obstacle is preserved
   - test where simultaneous addition and removal of ships is handled correctly
   - test where an unchanged ship list updates the existing `Boat` obstacle in place rather than replacing it


raghumanimehta (changes requested):
Two comments in order: 
1) There is a larger Obstacle level issue that is confusing. If you look at the function `is_state_valid` in `ompl_path.py` then you will realize that it is using `OMPLPath.obstacles`. This isn't updated and seems old design. Not to mention using `OMPLPath.obstacles` seems redundant since `OMPLPath` already has ownership of `LocalPathState` so we can simplify to just using `LocalPathState`'s obstacles (i.e. `self.state.obstacles` there). 
2) I noticed that you are updating land all the time by collecting for it. Again this seems not required as land is not updated at all if the reference lat lon is the same so you can implement a boat only api in `obstacles.py` which is called by `update_obstacles` in `local_path.py`. 

Lmk if these don't make sense.

raghumanimehta (approved):
Good job!

PR #1009, opened 2026-07-17
Raghumanimehta/1008 rudder heading

### LINKED ISSUE #1008: PATH: Use heading in `\rudder` topic instead of `GPS` topic
PATH was using `GPS` heading field as the heading for calculations. This is not right. As mentioned by @lross03, we should be use `rudder` topic for the heading.

raghumanimehta:
### Description

- Resolves #1008 
- Updated the src and mocks to use rudder (subsribe and publish respectively). 
- Other things were untouched. 
- Codex added some tests that i couldn't give a damn about.
- Unit tests pass 
- running test plans work

### DIFF
--- a/src/local_pathfinding/README.md
+++ b/src/local_pathfinding/README.md
@@ -11,7 +11,7 @@ Using main launch file: `ros2 launch local_pathfinding main_launch.py`
 Local pathfinding uses the following ROS topics for boat state:
 
 - `gps` (`custom_interfaces/GPS`) supplies geographic position and speed.
-- `rudder` (`custom_interfaces/HelperHeading`) supplies the e-compass boat heading.
+- `rudder` (`custom_interfaces/HelperHeading`) supplies the e-compass boat heading. The topic name `\rudder` is confusing; it's misleading. When you see rudder, think heading.
 
 Despite its topic name, `rudder` contains the direction the boat's bow is pointing, not the
 physical rudder angle. It uses the `HelperHeading` navigation convention: `0°` is north, values
--- a/src/local_pathfinding/README.md
+++ b/src/local_pathfinding/README.md
@@ -6,6 +6,18 @@ UBC Sailbot's local pathfinding ROS package
 
 Using main launch file: `ros2 launch local_pathfinding main_launch.py`
 
+## Navigation Inputs
+
+Local pathfinding uses the following ROS topics for boat state:
+
+- `gps` (`custom_interfaces/GPS`) supplies geographic position and speed.
+- `rudder` (`custom_interfaces/HelperHeading`) supplies the e-compass boat heading.
+
+Despite its topic name, `rudder` contains the direction the boat's bow is pointing, not the
+physical rudder angle. It uses the `HelperHeading` navigation convention: `0°` is north, values
+increase clockwise, and the valid range is `(-180°, 180°]`. Missing, invalid, or stale heading
+data prevents local pathfinding from enabling sail.
+
 ## Test Plans
 
 Test plans live in `test_plans/` and are documented in the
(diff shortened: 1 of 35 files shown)


src/local_pathfinding/test/test_local_path.py
```diff
@@ -83,7 +96,8 @@ def create_update_if_needed_inputs():
     gps = mock.Mock()
     gps.lat_lon = HelperLatLon(latitude=0.0, longitude=-0.1)
     gps.speed = mock.Mock(speed=0.0)
-    gps.heading = mock.Mock(heading=0.0)
+    gps.heading = mock.Mock(heading=123.0)
+    heading = HelperHeading(heading=0.0)
```
FireBoyAJ24, line 100: Why did you change this? Was there a reason in going from 0.0 to 123 degrees?

src/local_pathfinding/README.md
```diff
@@ -6,6 +6,18 @@ UBC Sailbot's local pathfinding ROS package
 
 Using main launch file: `ros2 launch local_pathfinding main_launch.py`
 
+## Navigation Inputs
+
+Local pathfinding uses the following ROS topics for boat state:
+
+- `gps` (`custom_interfaces/GPS`) supplies geographic position and speed.
+- `rudder` (`custom_interfaces/HelperHeading`) supplies the e-compass boat heading.
```
FireBoyAJ24: `rudder` is a very enigmatic and elusive name, as you'll see in this readme. I believe it is best to rename it to something reasonable. A reasoning over why gps heading is not used in this place may help future members know our reasoning.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -120,7 +120,8 @@ class Sailbot(Node):
 
     Subscribers:
         ais_ships_sub (Subscription): Subscribe to a `AISShips` msg.
-        gps_sub (Subscription): Subscribe to a `GPS` msg.
+        gps_sub (Subscription): Subscribe to GPS position and speed in a `GPS` msg.
+        heading_sub (Subscription): Subscribe to e-compass boat heading from `rudder`.
```
FireBoyAJ24, line 124: Again why rudder 😭. It is the most bizarre name for the boat heading.

FireBoyAJ24 (changes requested):
Could we change the topic message from rudder to something more clearer. Like `heading`?

src/local_pathfinding/test/test_local_path.py
raghumanimehta: No reason at all. The point was that the `gps.heading` will be ignored.

src/local_pathfinding/README.md
raghumanimehta: I used the name `rudder` because topic name is `rudder`. I will add the fact that the name is misleading

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: That's the topic name. I didn't choose the name.

FireBoyAJ24 (approved):

lross03:
> Could we change the topic message from rudder to something more clearer. Like `heading`?

The heading comes from a can frame titled "RUDDER_DATA_FRAME". I also think the name is a poor choice, but I think if we change it, we should consider also making the can frame documentation more clear.

raghumanimehta:
> > Could we change the topic message from rudder to something more clearer. Like `heading`?
> 
> The heading comes from a can frame titled "RUDDER_DATA_FRAME". I also think the name is a poor choice, but I think if we change it, we should consider also making the can frame documentation more clear.

Let's just skip this change for now. I have documented this change in PATH.

PR #1014, opened 2026-07-18
Changed COG and Speed bounds check to exempt raw values

### LINKED ISSUE #1013: Fix AIS unavailable value constants
### Changes

Update the AIS "unavailable value" constants in can_frame_parser to match the values inidacted by https://ubcsailbot.atlassian.net/wiki/spaces/prjt22/pages/1827176527/CAN+Frames and https://www.navcen.uscg.gov/ais-class-a-reports

i.e. ensure we are comparing raw CAN values to these constants with the exception of ROT which will have been converted to a positive value. (ELEC will never send negative values)

Apps247:
Resolves #1013

Added raw heading fields to AISShips class to check against 'unavailable' constants.

This seems better than just converting the constant values in the condition statements, because conversions may lose decimal places etc.


lross03 (dismissed):
lgtm

lross03:
Test case for AISShipsValid needs to be updated I think

lross03 (dismissed):
we don't use AIS heading. PR was good before just needed to check that test case that was failing

lross03 (approved):

PR #1015, opened 2026-07-18
Added wind and gps kill switch

### LINKED ISSUE #997: NET: Ability to shut down GPS or WindSensor to simulate failure in PATH
### Purpose

We want the ability to test if the PATH nodes react positively in failure scenearios. Some failure scenarios involve bad / no data from the on board sensors, this can be simulated by not sending the messages from NET to ROS.

Apps247:
Resolves #997


lross03 (approved):
lgtm

PR #1021, opened 2026-07-22
#1020: Handle unavailable COG and ROT values from NET

### LINKED ISSUE #1020: PATH: Handle unavailable COG and ROT values from NET
For AIS ships, COG represents course over ground, the direction in which the boat is moving on the ground (this is not necessarily heading, but that part is inconsequential to this issue) and ROT represents rate of turning. The values -128.0 for ROT and 360.0 for COG mean that these values are unavailable. PATH doesn't handle these cases in `obstacles.py`. Here is what we will do: 

```bash
if COG == 360.0:
    project a circle instead of trapezoid that extends equally in all directions
    to the maximum distance as if it were extending in one direction i.e. 
    case where the COG exists but instead of projecting along that direction only 
    we do this in all directions. 

if ROT == -128.0:
    use the branch specifically for this in `obstacles.py`. The comments in the codebase should be able to guide
    you. Let me know if you have any questions. 

```

ethanhu912:
### Description
- Resolves #1020 
- Added a guard to boundTo180 in `can_frame_parser.cpp` to prevent a COG value of 360.0 being converted to 0.0 before 
publishing to PATH
    - Also added this guard in mock AIS node
- Added functionality to handle unavailable COG and ROT values:
    - COG = 360.0: Project a circle with radius of speed * turn_projection_time_seconds, since direction is unknown
    - ROT = -128.0: Substitutes a pessimistic fallback ROT (10°/sec)
- Added unit tests

### Verification
- All unit tests pass
- Created new test plan with unavailable COG and ROT values on AIS ships, verified that the obstacle size makes sense.


src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -23,6 +23,9 @@
 COLLISION_ZONE_STRETCH_FACTOR = 1.25  # This factor changes the width of the boat collision zone
 RADIUS_MULTIPLIER = 5
 STRAIGHT_LINE_ROT_THRESHOLD_RAD_PER_SEC = 1e-4
+# Pessimistic ROT substituted when AIS reports ROT as unavailable (-128). ~10°/sec is
+# an aggressive turn rate for typical AIS-tracked vessels; sign is arbitrary (right turn).
+UNAVAILABLE_ROT_FALLBACK_RAD_PER_SEC = math.radians(10.0)
```
raghumanimehta: I think a better design would be to extend the collision in both left and right turn directions instead of just assuming that the boat turns only right.

raghumanimehta (changes requested):
Looks good except the comment I mentioned.

src/network_systems/projects/can_transceiver/src/can_frame_parser.cpp
```diff
@@ -510,7 +510,12 @@ msg::HelperAISShip AISShips::toRosMsg() const
     lat_lon.set__longitude(lon_);
 
     msg::HelperHeading cog;
-    cog.set__heading(utils::boundTo180(course_));
+    // 360.0 is the AIS sentinel for "COG unavailable"; preserve it instead of collapsing to 0.0
```
lross03, line 513: I only reviewed this section

lross03 (dismissed):

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -23,9 +23,13 @@
...
-# Pessimistic ROT substituted when AIS reports ROT as unavailable (-128). ~10°/sec is
-# an aggressive turn rate for typical AIS-tracked vessels; sign is arbitrary (right turn).
-UNAVAILABLE_ROT_FALLBACK_RAD_PER_SEC = math.radians(10.0)
+# Pessimistic ROT substituted when AIS reports ROT as unavailable (-128). ~2°/sec is on
+# the upper end for typical AIS-tracked vessels (container ships ~0.5°/s, small craft
+# up to ~5°/s); it balances "aggressive turn" against tight-spin degenerate projections
+# where a very high rate makes the boat effectively stationary. The full ROT is unknown,
+# so we union three projections at this rate: straight-line (rate 0), right turn, and
+# left turn (see Boat._set_unknown_rot_zone).
+UNAVAILABLE_ROT_FALLBACK_RAD_PER_SEC = math.radians(2.0)
```
raghumanimehta: Why was this updated?

raghumanimehta: I would argue keeping this at 5 degrees. We want a pessimistic bound as a guarantee. Optimizing this is not worth crashing the boat

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -413,6 +424,40 @@ def _set_turning_zone(
         self.collision_zone = self._raw_collision_zone
         prepared.prep(self.collision_zone)
 
+    def _set_unknown_rot_zone(
+        self,
+        rot_rps: float,
```
raghumanimehta: We don't need to pass this as a parameter as this is a global constant. This is a small detail but dramatically improves the readibility of the code. You can therefore immediately refere to this constant within the function.

src/local_pathfinding/local_pathfinding/obstacles.py
```diff
@@ -413,6 +424,40 @@ def _set_turning_zone(
...
+        Used when AIS reports ROT as unavailable (-128): both the rate and direction of
+        turn are unknown, so we cover the full maneuvering envelope by unioning three
+        projections: straight-line motion (ROT ~= 0) plus a pessimistic-magnitude turn
+        in each direction.
+        """
+        rot_magnitude = abs(rot_rps)
+        self._set_turning_zone(
+            rot_magnitude, speed_kmps, cog_rad, x, y, projected_distance
+        )
+        right_raw = self._raw_collision_zone
```
raghumanimehta: We don't need right_raw or left_raw at all. If you look at set_turning_zone it uses the same value for the collision zones, right? 
You can thus remove all the raw value from computations making this a simpler fucntion that what it is.

raghumanimehta (changes requested):
Some minor changes

raghumanimehta (approved):

PR #1032, opened 2026-08-01
Mock CAN Transceiver

FireBoyAJ24:
### Description

- The purpose of this pull request is to add a _new mode_ (yipeeeeeeee). In this mode, we can use the stored raw can messages from previous owt and run them to test our pathfinding without the need for the on-water test (YIPPEEEEEEEEEEEEEEEE).
- The mock can transceiver works by creating two threads: bus fd and can transceiver fd. From my understanding, the bus fd sends commands to the boat (it is stored), and the CAN transceiver fd should process the stored CAN frames and publish them on ROS2.
- The CAN frames are stored in a library folder (I didn't know where else to put it) `src/network_systems/lib/can_log/combined_candump.csv` and is a pretty large file, ~33 mb.
- The CAN frames are initially parsed (Read the file, try to fix any broken CAN messages and skip those that can't be fixed). All the can frames are stored in a vector, and we then use a for loop to publish each can frame's information.
- 

### Verification

For each command, check whether the ros2 sensor topics are getting the information we are providing to them.
- [x] `ros2 launch network_systems main_launch.py mode:=can`
- [x] `ros2 launch global_launch main_launch.py mode:=can config:=on_water_globals.yaml`


FireBoyAJ24:
I mistakenly branched from the wrong branch. I will open a new PR that contains only mock transceiver changes.

PR #1039, opened 2026-08-02
update rudder controller for simulator

### LINKED ISSUE #1038: SIM: Rudder Controller Refactoring
it doesnt work (sad emoji) so make it work (happy emoji)

- implement a proper PID 
- make it so that the controller recalculates the setpoint every tick
- make sure the integral term does not blow up when the rudder is at the maximum angle

bubbles5228:
### Description
- updated the rudder controller to use pid, as well as compute the pid values and error every tick
- resolves #1038

### DIFF
--- a/src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
+++ b/src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
@@ -1,4 +1,6 @@
 """Low level control logic for actuating the rudder and the sail."""
+
+from collections import deque
 from math import atan2, copysign, cos, sin
 from typing import Tuple
 
@@ -73,9 +75,21 @@ class ActuatorController:
 
 
 class RudderController(ActuatorController):
-    """General Class for the Actuator Controller.
+    """PID controller for the rudder.
 
-    RudderController Extends: ActuatorController
+    Attributes:
+        `current_heading` (float): Current boat heading direction in degrees,
+            0 degrees (North) at the positive y axis and increasing clockwise.
+        `desired_heading` (float): Target boating heading direction in degrees,
+            0 degrees (North) at the positive y axis and increasing clockwise.
+        `current_control_ang` (float): Last commanded rudder angle in degrees.
+        `time_step` (float): Time per iteration given in seconds.
+        `kp`, `ki`, `kd` (float): PID constants for the rudder controller.
+        `cp` (float): Tuning parameter for control action.
+        `buffer_size` (int): Number of samples used to filter the derivative error.
+        `max_angle_range` (Tuple): Max rudder control angle range in degrees, minimum[0] and
+            maximum[1]
+
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -73,9 +75,21 @@ def is_target_reached(self) -> bool:
 
 
 class RudderController(ActuatorController):
-    """General Class for the Actuator Controller.
+    """PID controller for the rudder.
+
+    Attributes:
+        `current_heading` (float): Current boat heading direction in degrees,
+            0 degrees (North) at the positive y axis and increasing clockwise.
```
FireBoyAJ24, line 82: What are their limits? Like [-180, 180) or [0, 360)

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -150,30 +190,39 @@ def _compute_error(self) -> float:
...
+        self._error_buffer.append(error_rad)
 
-        heading_error_rad = self._compute_error()
+        p_term = (self.kp * error_rad) / (1 + self.cp * abs(error_rad))
 
-        rudder_setpoint_rad = (self.kp * heading_error_rad) / (
-            1 + (self.cp * abs(heading_error_rad))
-        )
+        self._integral_error += error_rad * self.time_step
+        i_term = self.ki * self._integral_error
```
FireBoyAJ24, line 204: An integral windup (reset integral error to 0.0) is a good idea here if the actual rudder angle is consistently above or below the desired rudder angle. https://www.youtube.com/watch?v=z-IenBvGoYY

FireBoyAJ24 (changes requested):

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
```diff
@@ -97,26 +113,33 @@ def __init__(
...
-            control_speed,
-            max_angle_range,
-        )
+        self.current_control_ang = current_control_ang
+        self.time_step = time_step
+        self.max_angle_range = max_angle_range
 
         self.current_heading = bound_to_180(current_heading)
         self.desired_heading = bound_to_180(desired_heading)
+        self.setpoint = 0.0  # last commanded rudder angle in degrees
```
raghumanimehta, line 132: from path exp, adding units to the field/variable name makes life way easier for humans as well as AI.

raghumanimehta (changes requested):

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
bubbles5228: ill address this eventually... i think i have to change it - the angles in the sim are a mess

src/boat_simulator/boat_simulator/nodes/low_level_control/controller.py
FireBoyAJ24: @bubbles5228 Use the typing in `types.py` and such for the controller! They can be used here as well

PR #1044, opened 2026-08-07
Add support for debug rudder data frame and fix AIS ROT bounds

### LINKED ISSUE #1043: Enable getting heading from debug rudder frame
### Purpose

main rudder data frame does not work, but the debug frame does. we need to extract heading from this frame instead.

### Description

- add the manual option to use the debug frame instead of the main data frame
- enable the debug frame to be a fallback if we do not receive data from the main frame for 5s

lross03:
### Description

- Resolves #1043

- Adds support for debug rudder data frame to CAN transceiver
- Defaults to the regular rudder data frame for heading, unless it does not receive data for 5s. Then, it will switch to debug until regular rudder data frame returns
- Can manually use the debug frame by setting launch argument `rudder_debug:=true` where it only use the debug frame for as long as the software is online
- Changes ROT bounds from [-126,126] to [-127,127]

### Verification

- [x] Tested using OWT data on mock CAN transceiver

### DIFF
--- a/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
+++ b/src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
@@ -21,8 +21,9 @@
 #include "mock_can_bus.h"
 #include "net_node.h"
 
-constexpr int  QUEUE_SIZE     = 10;  // Arbitrary number
-constexpr auto TIMER_INTERVAL = std::chrono::milliseconds(500);
+constexpr int  QUEUE_SIZE              = 10;  // Arbitrary number
+constexpr auto TIMER_INTERVAL          = std::chrono::milliseconds(500);
+constexpr auto RUDDER_FALLBACK_TIMEOUT = std::chrono::seconds(5);
 
 namespace msg = custom_interfaces::msg;
 using CAN_FP::CanFrame;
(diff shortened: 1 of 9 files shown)


src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -21,8 +21,9 @@
 #include "mock_can_bus.h"
 #include "net_node.h"
 
-constexpr int  QUEUE_SIZE     = 10;  // Arbitrary number
-constexpr auto TIMER_INTERVAL = std::chrono::milliseconds(500);
+constexpr int  QUEUE_SIZE              = 10;  // Arbitrary number
+constexpr auto TIMER_INTERVAL          = std::chrono::milliseconds(500);
+constexpr auto RUDDER_FALLBACK_TIMEOUT = std::chrono::seconds(5);
```
FireBoyAJ24, line 26: Side note, the rudder debug frame roughly gets received every 500 ms as shown by below timestamps:

> 2026-06-06T17:47:01.396641
> 2026-06-06T17:47:01.400565
> 2026-06-06T17:47:01.463563
> 2026-06-06T17:47:01.575643

FireBoyAJ24 (approved):

PR #1046, opened 2026-08-07
#1031: AIS list overwrite bug

### LINKED ISSUE #1031: PATH: AIS list overwrite bug
## What happened
  On-water, AIS was inconsistent and sometimes came through as an empty list. Two halves to it. The
  transceiver is meant to only ever publish a complete fleet snapshot, but the July log has 314
  `Mismatched numShips` warnings, so that assembly was degrading and partial sets were going out. Then
  every consumer stage treats the latest message as the complete picture of the world. If `{a,b,c}` are
  tracked and the next message carries only `{d}`, the list becomes `{d}` and a, b and c stop existing
  for collision checking. They come back intact once a later message includes them, so nothing is
  permanently lost, but the planner is blind to them until then and a replan in that window will route
  straight through them. An empty message drops everything. Start with why a partial set is being
  published, then fix the consumer so a partial update cannot wipe known ships.

  ## Repro
  `test_plans/scenario_events_demo.yaml` does both in one 20 second run: it starts with ships 1 and 2,
  the t=6s event carries only ship 1, and the t=15s event carries `ships: []`. Watch tracked boats leave
  `state.obstacles` while `sail` stays true.

## Suggested design

Overhaul how AIS ships are tracked. As per https://chatgpt.com/share/6a6bd19e-d5e0-83e8-bd2a-c997641af07e , you need to track the ship for at least 10 minutes . In that timeline, if you receive a message for the ship then you can refresh the data and the timeline back to the baseline (10 minutes). You can use a map from `ais_id : AISShip Object` as id from ais is uuid (universally unique identifier). 

The workflow could be: 
```
receive message from NET on AIS topic (ais) -> 
if AIS list exists:
    refresh AIS list by looping through received list from NET
    checking the map if it exists, updating if it does else adding it to the map
else: 
    create list and add values from recevied list from NET
```

  ## Acceptance criteria
  - [ ] A ship stays tracked until it times out
  - [ ] Empty or shrunken message logs a warning and keeps known ships, node stays up
  - [ ] End-to-end test: publish `{a,b,c}`, then `{d}`, then `AISShips()`, assert a, b and c survive
  - [ ] more automated testing 
  - [ ] Also try asking AI to make a complex test plan to trigger such timing anomalies and verify the output.

ethanhu912:
### Description
- Resolves #1031 
- `Sailbot` stores ships using a map that have an associated last-refresh time
    - Ships are discarded if they have not been updated in an `ais_ships` message for 10 minutes, shown by `AIS_SHIP_TIMEOUT_SEC` = 600s
    - Previously, a message carrying `{d}` after tracking `{a,b,c}` left the planner with `{d}` alone since each `ais_ships` message was treated as the whole picture
    - Now, messages update the map stored in `Sailbot` with information that they carry instead of originally fully rewriting and erasing what is omitted
    - `_rebuild_ais_snapshot` then updates self.ais_ships from the map stored in `Sailbot`
    - Omissions are logged as warnings during each transition
- Wrote unit tests that covers warning logs, ship expiry after 10 minutes, merging new `ais_ships` updates into the existing map and ensuring omitted ships remain in obstacles
- Created new test plan `ais_partial_updates.yaml` that starts with two ships tracked that then publishes two messages:
    - t=6s: Message only carrying one ship
    - t=15s: Fully empty message

### Verification
- All unit tests pass
- Launched before any changes to reproduce the issue
- Launched with `ais_partial_updates.yaml` test plan
    - Ensured that obstacles and ships remain instead of disappear when passed an empty message
    - Omission warning logs are sent once per transition

## Notes
- Ships that are omitted retain their last known position. This means that for 10 minutes, the ship and obstacle information passed to the planner could be outdated
- Ships that are out of range would be avoided up to 10 minutes


raghumanimehta:
@codex review this PR.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -223,6 +239,12 @@ def __init__(self) -> None:
         )
 
         # attributes from subscribers
+        # ais_ships is derived from _tracked_ais_ships rather than assigned straight from the
+        # topic, so downstream consumers keep receiving a complete snapshot. It stays None until
+        # the first message arrives so _all_subs_active can tell "no AIS yet" apart from "AIS is
+        # working and nothing is in range".
+        self._tracked_ais_ships: dict[int, TrackedAISShip] = {}
```
raghumanimehta: I don't really understand the comment above. Can you simplify the language of what it is trying to say? 

Also why do you need `self.ais_ships` if `self._tracked_ais_ships` exists?

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -76,6 +78,20 @@
 """
 
 
+@dataclass(frozen=True)
+class TrackedAISShip:
+    """An AIS ship Sailbot is tracking, with the time its data was last refreshed.
+
+    An `ais_ships` message is a partial snapshot, not a statement about every vessel in range:
```
raghumanimehta: this could be two lines long. Let's tone down the AI comments that seem egregiously long.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -463,8 +485,43 @@ def _load_persisted_global_path(self) -> bool:
...
-        self.ais_ships = msg
+
+        now_sec = self._now_sec()
+        tracked_before = set(self._tracked_ais_ships)
+
+        received_ids: set[int] = set()
+        for ship in msg.ships:
+            # 0 is not a valid MMSI; it is the AIS_SHIPS_UNAVAILABLE sentinel, so it must never
+            # be tracked as a real vessel.
+            if ship.id == 0:
```
raghumanimehta: magic number violation.

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -463,8 +485,43 @@ def _load_persisted_global_path(self) -> bool:
...
+                continue
+            self._tracked_ais_ships[ship.id] = TrackedAISShip(ship=ship, last_seen_sec=now_sec)
+            received_ids.add(ship.id)
+
+        # A degraded transceiver republishes the same partial snapshot at the full AIS rate, so
```
raghumanimehta: The comment is bad. 

Also this logic is not required at all. Too defensive and not required. Even if you log what ship existed in the previous run and then re-log it when it come back, it's not useful in my eyes. 

This adds noise to the logs and I would strongly suggest just removing this and it's corresponding tests.

raghumanimehta (changes requested):
Good work. 

General feedback: Please review AI generated comments, they are sometimes too bloated. If you wrote those comments yourself, try to convey your thoughts in simpler language. 
The logging logic is too complicated, and totally unnecessary. We have a generous grace period to not require the complex logic. Not to mention, that those logs add more nosie than substance.

src/local_pathfinding/local_pathfinding/node_navigate.py
raghumanimehta: This was not resovled.

raghumanimehta (changes requested):
You forgot one thing!

raghumanimehta (approved):

PR #1050, opened 2026-08-08
Raghumanimehta/1047 ais 127 handling

### LINKED ISSUE #1047: PATH: handle the screenshot for AIS
Image

raghumanimehta:
### Description

- Resolves #1047 
- simplified unknown/uncertain ROT to make a circle around the boat case. 
- instead of complex calculations from before which handled -128 rot, now we just draw a cricle about the boat
- i also added +/- 127 to that because it gives us a lower bound so assuming worst case it should be a circle. 
- circle radius is whatever distance that would have been calculated as straight line projection had there been no turning. 

### Verification

- [ ] unit tests
- [ ] manual verification by running a test plan


raghumanimehta:
@codex review

AMaharaj16 (approved):
looks good

PR #1056, opened 2026-08-11
fix(web): reverted dynamic about page ui to static

AndyG6:
### Description
addresses the issue of the about page dynamic text loading breaking on mobile by making it static instead.

We originally wanted to keep it faithful to the desktop version, but that was causing a lot of issues, so we made it static instead. Should be fixed now!

### Verification
image


raghumanimehta (approved):

raghumanimehta:
@codex review

yuekaifred (approved):
GOODJOB

PR #1059, opened 2026-08-12
add `can_transceiver_node/manual_mode` param as global launch argument

jeffrey-xue:
### Description

- Resolves race condition issue described by Pawan. 
> So basically anytime the software system is running, the trim tab always zeroes itself after we manually send an actuation command via the GUI. This occurs even when CAN send from software is turned off but when the software system is fully turned off it doesn't happen anymore (no idea how that makes any sense)
- This change allows for someone operating the GUI to toggle `manual_mode` true or false immediately, as opposed to launching the software and then sending a `ros2 param set` command afterwards.
- Associate PR for the POLARIS GUI --> https://github.com/UBCSailbot/polaris-gui/pull/39 

### Verification

- Mirrors the implementation for the `rudder_debug` launch arg. But I'm not super familiar with network systems, so confirmation from NET would be very good.


raghumanimehta:
so the workflow would be to stop software, then restart it in manual mode, and then when you want to run software, run software without manual mode?

raghumanimehta:
> so the workflow would be to stop software, then restart it in manual mode, and then when you want to run software, run software without manual mode?

I would need a bit more reasoning as how this would work.

raghumanimehta:
@codex review

jeffrey-xue:
> so the workflow would be to stop software, then restart it in manual mode, and then when you want to run software, run software without manual mode?

To my understanding, there are situations during the owt where we want to run the software with CAN communications off (i.e. manual_mode = true) from the get go as opposed to launching --> then clicking the disable CAN comms button. Previously there was no way to achieve this. Now, if someone starts the software with manual_mode=true via the GUI they can. And afterwards they still have the option to renable CAN communications through one of the other GUI buttons (or they can restart software).

src/global_launch/main_launch.py
```diff
@@ -109,6 +109,15 @@ def resolve_config(filename: str) -> str:
...
         + " empty to use the value in the config file.",
     ),
+    DeclareLaunchArgument(
+        name="manual_mode",
+        default_value="",
+        choices=["", "true", "false"],
+        description="Override whether can_transceiver_node starts in manual mode."
+        + "This functionally disables CAN communication from soft to elec."
+        + "Leave empty to use the value in the config file. If undefined in the config file,"
+        + "the declared value is used (see can_transceiver_ros_intf.cpp)."
```
Apps247, line 119: Could add leading spaces to these strings, similar to how it's done in the `rudder_debug` description above. So `" This functionally..."` etc.

But it's just a minor cosmetic change, optional

Apps247 (approved):
Lgtm on the NET side. It just exposes the already existing `manual_mode_` value to a global launch param `manual_mode`, similar to how it's done with `rudder_debug`.

Unrelated: Could add leading spaces as described in above comment, but that's just minor and cosmetic

raghumanimehta (approved):

raghumanimehta:
@lross03 you going to review this?

lross03 (approved):

src/global_launch/main_launch.py
lross03: true but its aight

Issue #1065, opened 2026-08-15
PATH: Complex Scenario Verification


AMaharaj16:
### Background

Since pathfinding will not be validated in a real OWT, we need to rigorously verify the PATH algorithm through simulation. As of @ethanhu912's PR #1005, dynamic test plans are now supported, allowing us to define timestamped environment changes (wind shifts, AIS targets, GPS drift, etc.) in a single YAML file. Combined with the sequential test plan runner, we have everything we need to stress-test the algorithm under realistic conditions.

---

### Task

Every PATH member must design and run **one complex dynamic test plan** lasting **at least 6 hours** (longer the better). Submit a **PR by Sunday, August 16th at noon** with a short document summarizing your voyage and key findings.

Please make your scenario **genuinely complex**. We want to catch real bugs, so throw everything at it. The more realistic and challenging, the better. As we are officially launching the Sailbot on Tuesday, this is a VERY urgent task for all pathfinding members.

AI can be helpful for keeping an eye on long voyages. Use a powerful (but cheap) model and prompt it to check in periodically rather than continuously to keep token usage from maxing out. This can allow you to run the test plan and step away from the laptop while the AI monitors it.

---

### Steps

1. **Design a scenario.** Author a dynamic test plan YAML with many timestamped events. Make it hard.
2. **Run it.** Use the sequential test plan runner. Target **5+ hours** of runtime and monitor it.
4. **Document and PR.** Summarize your scenario, observations, and any bugs surfaced. PR due **Sunday, August 16th at noon**. (tomorrow)

PR #1072, opened 2026-08-20
Respawn subscription self-heal + launch logging cleanup

AMaharaj16:
Launch prep for local pathfinding. Three separate commits on this branch:

- **Respawn re-subscribe watchdog (#1065):** after navigate auto-respawns from an
  OMPL segfault it can silently fail to re-match a live publisher and run wind-blind
  for the rest of the voyage. Adds a recovery-only watchdog that recreates an
  orphaned subscription (best-effort, capped) and never exits the node, so the worst
  case is navigate's existing sail-disabled fallback.
- **Remove invalid-state file logging:** the OMPL validity check appended to
  invalid_states.log on every invalid state (unbounded disk growth) and re-raised on
  write failure from inside the solver callback. Removed; the collision check is
  unchanged.
- **Delete INFO-level logging:** strips all INFO logs from local_pathfinding,
  controller, and network_systems (CAN and satellite transceivers) to cut disk usage
  on the boat. Warn, error, and debug are untouched.

Verification: 510/510 local_pathfinding tests pass, flake8 clean, network_systems
builds clean, and a forced-respawn sim smoke recovered navigate with no crash-loop.

### DIFF
--- a/src/controller/controller/wingsail/wingsail_ctrl_node.py
+++ b/src/controller/controller/wingsail/wingsail_ctrl_node.py
@@ -195,6 +195,10 @@ class WingsailControllerNode(Node):
         msg.trim_tab_angle_degrees = self.__trim_tab_angle
 
         self.__trim_tab_angle_pub.publish(msg)
+        self.get_logger().debug(
+            f"Published to {self.__trim_tab_angle_pub.topic} \
+                the following angle: {msg.trim_tab_angle_degrees}"
+        )
 
     @property
     def pub_period(self) -> float:
--- a/src/controller/controller/wingsail/wingsail_ctrl_node.py
+++ b/src/controller/controller/wingsail/wingsail_ctrl_node.py
@@ -196,11 +196,6 @@ class WingsailControllerNode(Node):
 
         self.__trim_tab_angle_pub.publish(msg)
 
-        self.get_logger().info(
-            f"Published to {self.__trim_tab_angle_pub.topic} \
-                the following angle: {msg.trim_tab_angle_degrees}"
-        )
-
     @property
     def pub_period(self) -> float:
         return self.get_parameter("pub_period_sec").get_parameter_value().double_value
@@ -220,7 +215,6 @@ class WingsailControllerNode(Node):
             msg (WindSensor): Filtered wind sensor data from CanTrxRosIntf.
         """
         self.__filtered_wind_sensor = msg
-        self.get_logger().info(f"Received data from {self.__filtered_wind_sensor_sub.topic}")
 
     def __sail_sub_callback(self, msg: DesiredHeading) -> None:
         """Stores the latest desired heading data. We use the sail message to decide when the boat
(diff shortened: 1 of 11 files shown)


lross03 (changes requested):
The NET changes are too aggressive. We only needed to remove the CAN logging in can trasnceiver ros intf because it is constantly producing logs every second. local transceiver does not log often, so it should stay.

src/controller/controller/wingsail/wingsail_ctrl_node.py
```diff
@@ -196,11 +196,6 @@ def __publish(self):
 
         self.__trim_tab_angle_pub.publish(msg)
 
-        self.get_logger().info(
-            f"Published to {self.__trim_tab_angle_pub.topic} \
-                the following angle: {msg.trim_tab_angle_degrees}"
-        )
-
```
FireBoyAJ24, line 203: I recommend changing them to debug statements rather than deleting the log statements outright

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -264,23 +299,8 @@ def __init__(self) -> None:
...
-                f"On water globals config ({global_config}) is successfully loaded "
-            )
-        elif global_config == "launch_globals.yaml":
-            self.get_logger().info(
-                f"Launch globals config ({global_config}) is successfully loaded "
-            )
-        else:
-            self.get_logger().info(
-                f"Default globals config ({global_config}) is successfully loaded "
-            )
```
FireBoyAJ24, line 283: This is useful for debugging 😭

src/local_pathfinding/local_pathfinding/node_navigate.py
```diff
@@ -591,18 +609,8 @@ def desired_heading_callback(self) -> None:
...
-                or desired_heading != self.desired_heading.heading.heading
-            ):
-                self.get_logger().info(f"Updating desired heading to: {msg.heading.heading:.2f}")
-
             self.desired_heading = msg
 
-            self.get_logger().info(
-                f"Publishing to {self.desired_heading_pub.topic}: {msg.heading.heading}, "
-                f"sail == {msg.sail}"  # noqa
-            )
```
FireBoyAJ24, line 605: Make this into a debug statement

FireBoyAJ24 (changes requested):
A lot of the deleted info logs can be converted to debug statements

src/network_systems/projects/can_transceiver/src/can_transceiver_ros_intf.cpp
```diff
@@ -65,15 +64,13 @@ class CanTransceiverIntf : public NetNode
...
                 try {
                     can_trns_ = std::make_unique<CanTransceiver>();
                 } catch (std::runtime_error err) {
                     RCLCPP_ERROR(this->get_logger(), "%s", err.what());
                     throw err;
                 }
             } else if (mode == SYSTEM_MODE::DEV || mode == SYSTEM_MODE::SIM) {
-                RCLCPP_INFO(this->get_logger(), "Running CAN Transceiver in %s mode with CAN Sim Intf", mode.c_str());
                 try {
                     sim_intf_fd_ = mockCanFd("/tmp/CanSimIntfXXXXXX");
```
lross03, line 78: I would keep these but I'm not going to gate the PR on this

lross03 (dismissed):

